Issue #7, opened 2018-06-18
Ensure the semantics of context propagation are clear


mmusgrov:
If you have a microservice A that wants to participate in an LRA and calls microservice B (which does not have specific LRA behaviour) and this calls microservice C (which does have specific LRA behaviour), then B should not need to be annotated with @LRA. The spec should be clear that is the defined behaviour.

mmusgrov:
PR was merged

rdebusscher:
Reopened.

The `always propagate Context Header` is counter intuitive and does not allow to call an endpoint which doesn't have the Context header (when one is present)

rdebusscher:
The proposal is to define a configuration parameter to activate such propagation under all circumstances.

Note, the propagation of the LRA context is ALWAYS perform within an LRA context (JAX-RS method is annotated with @LRA except for @LRA.Type.NOT_SUPPORTED as this type means what you do not have an LRA context active)

rdebusscher:
No Consensus in trying to make this better.

PR #10, opened 2018-06-23
issue7 Context propagation across non-LRA aware services

mmusgrov:
https://github.com/eclipse/microprofile-lra/issues/7

This PR adds javadoc text to the LRA annotation that specifies that an LRA context is propagated across non-LRA aware services (ie one that does not specify any LRA behaviour).

The PR also adds a TCK test to verify this behaviour.


api/pom.xml
```diff
@@ -21,7 +21,7 @@
         <!-- This is just for now and will not work if the API has a separate release cycle than the rest. -->
         <groupId>org.eclipse.microprofile.lra</groupId>
         <artifactId>microprofile-lra-parent</artifactId>
-        <version>1.0-SNAPSHOT</version>
+        <version>0.0.2.Final-SNAPSHOT</version>
```
Emily-Jiang: Why do you change the version? The version should stay as 1.0-SNAPSHOT till 1.0 is released.

mmusgrov: Good point, the change crept in because that is what we are using privately, I will revert this part of the fix, thanks.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -121,11 +124,14 @@ private static void initTck(LRAClient lraClient) {
...
             int rcPort = Integer.getInteger(LRA_COORDINATOR_PORT_KEY, COORDINATOR_SWARM_PORT);
             String coordinatorPath = System.getProperty(LRA_COORDINATOR_PATH_KEY, "lra-coordinator");
+            String rcPath = System.getProperty(LRA_RECOVERY_PATH_KEY, "lra-recovery-coordinator");
 
             micrserviceBaseUrl = new URL(String.format("http://localhost:%d", servicePort));
-            rcBaseUrl = new URL(String.format("http://%s:%d", rcHost, rcPort));
+            rcBaseUrl = new URL(String.format("http://%s:%d/%s", rcHost, rcPort, rcPath));
 
             lraClient.setCoordinatorURI(new URI(String.format("http://%s:%d/%s", rcHost, rcPort, coordinatorPath)));
+//            lraClient.setRecoveryCoordinatorURI(new URI(String.format("http://%s:%d/%s", rcHost, rcPort, rcPath)));
```
tomjenkinson: I imagine this could be removed instead of commented out

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -589,10 +597,9 @@ private void joinAndEnd(boolean waitForRecovery, boolean close, String path, Str
             lraClient.cancelLRA(lra);
 
         if (waitForRecovery) {
-            String recoveryPath = System.getProperty(LRA_RECOVERY_PATH_KEY, "lra-recovery-coordinator");
             // trigger a recovery scan which trigger a replay attempt on any participants
             // that have responded to complete/compensate requests with Response.Status.ACCEPTED
-            resourcePath = recoveryTarget.path(recoveryPath).path("recovery");
+            resourcePath = recoveryTarget.path("recovery");
```
tomjenkinson: A lot of the other "paths" in the test use "static final String"

mmusgrov:
@tomjenkinson @Emily-Jiang What did people think of my fix for this issue?

api/src/main/java/org/eclipse/microprofile/lra/annotation/LRA.java
```diff
@@ -40,6 +40,12 @@
  * performs will carry the same header so that the receiving resource knows that it
  * is inside an LRA context (typically achieved using JAX-RS client filters).
  *
+ * Note that if an LRA is propagated to a resource that is not annotated with any
+ * particular LRA behaviour then the LRA will be suspended. However, if the method
+ * performs any outgoing JAX_RS requests then the LRA will be propagated to
```
Emily-Jiang: JAX-RS

api/src/main/java/org/eclipse/microprofile/lra/annotation/LRA.java
```diff
@@ -40,6 +40,12 @@
...
+ * the target of the request. This behaviour enables non-LRA aware resources
```
Emily-Jiang: Explain what is non-LRA aware resource

Emily-Jiang:
add more tck to check compensation occurs

api/src/main/java/org/eclipse/microprofile/lra/annotation/LRA.java
tomjenkinson: As per the call today you should make the wording in the Javadoc more explicit:
service A can start an LRA by using LRAClient or an annotation, 
then that it can call a (potentially) non-LRA aware service B (i.e. one that is not marked up with an @LRA annotation) and 
that service B can call an LRA aware service C (i.e. one that uses LRAClient to join the LRA or has the @LRA annotation) and that in service C the LRA should be available

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
```diff
@@ -335,6 +339,14 @@ private String restPutInvocation(URL lraURL, String path, String bodyText) {
         return id;
     }
 
+    @PUT
+    @Path(MANDATORY_LRA_RESOURCE_PATH)
+    @LRA(LRA.Type.MANDATORY)
+    public Response activityWithMandatoryLRA(@HeaderParam(LRA_HTTP_RECOVERY_HEADER) String rcvId,
+                                             @HeaderParam(LRA_HTTP_HEADER) String lraId) {
+        return activityWithLRA(rcvId, lraId);
```
tomjenkinson: Change the TCK test such that it subsequently verifies the compensation handler for activityWithMandatoryLRA is called during compensation

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -601,7 +609,38 @@ private void joinAndEnd(boolean waitForRecovery, boolean close, String path, Str
...
+
+        lraClient.cancelLRA(lra);
+
+        // check that second service (the LRA aware one), namely
+        // {@link org.eclipse.microprofile.lra.tck.participant.api.ActivityController#activityWithMandatoryLRA(String, String)}
+        // was told to compensate
+        int[] cnt2 = {completedCount(true), completedCount(false)};
+
+        assertEquals(cnt1[0], cnt2[0], "complete should not have been called", resourcePath);
+        assertEquals(cnt1[1] + 1, cnt2[1], "compensate should have been called", resourcePath);
```
mmusgrov: @Emily-Jiang This line verifies that the second service was asked to compensate

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
mmusgrov: see the comment below where I added the check to verify that compensate was called. Apologies for squashing but the only change was to ask the LRA aware service how many times compensate had been called before and after the service invocation:
> int[] cnt = {completedCount(true), completedCount(false)};
followed by an assertion that the compensation count had incremented.

tomjenkinson: OK

mmusgrov:
The test is the method on line 616 of the PR changeset (TckTests#noLRATest). This is the client (or service A). It starts an LRA on line 622:

    URL lra = lraClient.startLRA(null, "SpecTest#noLRATest", ...)

It performs a JAX-RS request on line 625 to service B (the resource path is on line 617 and it maps to StandardController#work2 which is on line 49 in the PR changeset).

StandardController#work2 should not be LRA aware (ie it does not have any LRA annotations to define whether a context should be present). If it finds such a context it returns a PRECONDITION_FAILED error on line 52 which causes the test to fail (line 628 checkStatusAndClose() in the changeset). *This step validates that service B does NOT run with a context*.

StandardController#work2 then sends a request to service C on line 59. The path is set up on line 55:

    resourcePath = ClientBuilder.newClient().target(context.getBaseUri())
                .path(ACTIVITIES_PATH)
                .path(ActivityController.MANDATORY_LRA_RESOURCE_PATH);

This path maps to ActivityController#activityWithMandatoryLRA (line 345 of the PR changeset) which just calls a pre-existing method in the same class (activityWithLRA). The class keeps a count of the number completion and compensation notifications.

Finally the client (service A / TckTests#noLRATest) sends a request to sevice C to ask it whether or not it was asked to compensate (line 635)

     int[] cnt2 = {completedCount(true), completedCount(false)};
     assertEquals(cnt1[0], cnt2[0], "complete should not have been called", resourcePath);
     assertEquals(cnt1[1] + 1, cnt2[1], "compensate should have been called", resourcePath);

Emily-Jiang:
Thanks for the explanation! LGTM

mmusgrov:
I added a revert commit for this PR since it did not merge cleanly (an earlier commit from another PR conflicted with this one). I will resubmit the PR with a commit that takes account of the changes made by other commits that happened after this PR was raised).

PR #23, opened 2018-08-27
Fix typos in the specification

xstefank:


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -455,23 +455,23 @@ of the spec as laid out in this document.
 === LRA Client API
 
 For completeness the proposal supports clients that wish to directly
-control LRAs and participants. To support this class of user an instance of
+control LRAs and participants. To support this, the class of user can obtain an instance of
```
mmusgrov: The original English was correct and the replacement you propose is not correct English.

If the original was not comprehensible then please say which bit doesn't make sense an I can change it.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -455,23 +455,23 @@ of the spec as laid out in this document.
...
 [[java-based-lra-participant-registration-api]]
 === Java based LRA participant registration API
 
 If an application does not directly expose JAX-RS endpoints for
 compensation activities then participants can join an LRA directly using an instance of
-<<source-LRAManagement,LRAManagement>>
-This interface requires that a participant be a serializable java class conforming to the
-<<source-LRAParticipant,LRAParticipant interface>>
+<<source-LRAManagement,LRAManagement>>.
+This interface requires that a participant is a serializable java class conforming to the
```
mmusgrov: Not sure why you changed the original text but okay.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
tomjenkinson: I agree with @mmusgrov I think it reads better as it was before actually

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: @mmusgrov sorry, I am not a native speaker, but `...requires that a participant be...` doesn't sound right to me, would you prefer to remove this change?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: I will revert this after feedback on the the next comment

spec/src/main/asciidoc/microprofile-lra-spec.adoc
tomjenkinson: Personally I think either would be fine. I think the periods after the <<source.. make sense to add

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: or if it is the word *class" that is the problem how about replacing with the equivalent "type" or "kind"  or "To support such users and instance of ... can be ...",  etc

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -457,21 +457,21 @@ of the spec as laid out in this document.
 For completeness the proposal supports clients that wish to directly
 control LRAs and participants. To support this class of user an instance of
 <<source-LRAClient,LRA Client API>>
-can be instantiated directly or injected if the client is using CDI. This
+which can be instantiated directly or injected if the client is using CDI. This
```
mmusgrov: it still doesn't read correctly. If it is the word *class" that is the problem how about replacing with the equivalent "type" or "kind" or "To support such users and instance of ... can be ...", etc

xstefank: my bad, reverted only the second line

Issue #16, opened 2018-07-02
Suggestions for improving the specification document


mmusgrov:
1. Add more examples that show how the developer uses the APIs such as how to inject an LRAClient.
1. Reference the Javadoc from the specification
1. Clearly articulate the use cases for the specification, for example,
    1. identify JAX-RS versus non JAX-RS use cases
    1. add more text on why the developer would use the programmatic as opposed to the CDI/annotation based APIs
1. Improve the javadoc for the Forget, Status and TimeLimit annotations
1. The state model for an LRA (section 3.1 of the spec) should have its own diagram and needs more descriptive text about the states and the transitions between them.

mmusgrov:
#6 requests a test for timelimits on LRAs but that issue also requested a javadoc change. The javadoc change would be better in this issue. Therefor this issue should improve the javadoc for the Forget, Status and TimeLimit annotations

mmusgrov:
@tomjenkinson I raised a separate issue #33 to cover your comment

mmusgrov:
PR merged

PR #37, opened 2018-11-04
Issue16: Suggestions for improving the specification document

mmusgrov:
Resolution of issue #16 
Suggestions for improving the specification document


mmusgrov:
I have done an mvn package on my PR and attached the pdf to this PR to make it easier to read the updated spec:

[microprofile-lra-spec.pdf](https://github.com/eclipse/microprofile-lra/files/2546285/microprofile-lra-spec.pdf)

mmusgrov:
This PR is ready for review - I will be fixing the two PR checks in the meantime

mmusgrov:
@tomjenkinson The only reference to coordinator/endpoint in the javadocs is a minor one in [CompensatorStatus](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/CompensatorStatus.java#L24) for which I have just pushed another commit.

The main reference to coordinators and endpoints is in [LRAClient](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java) and that should be the subject of separate issue, namely #32 which should be reopened to address it.

mmusgrov:
And here is a regenerated PDF of the spec that contains the changes I have just pushed:

[microprofile-lra-spec.pdf](https://github.com/eclipse/microprofile-lra/files/2576920/microprofile-lra-spec.pdf)

PR #46, opened 2018-11-22
Issue40: @LRA: Remove DelayClose, terminal() defaults now to true

rdebusscher:
The PR contains the changes for API, TCK, and Spec.

Changed the @LRA annotation on a few methods of ActivityController (added terminal = false) so that TCK tests should not be impacted by the change in the default value.

Should be reviewed if it is correct though.


tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
```diff
@@ -249,7 +249,8 @@ public Response forgetWork(@HeaderParam(LRA_HTTP_HEADER) String lraId) {
 
     @PUT
     @Path(ActivityController.ACCEPT_WORK)
-    @LRA(LRA.Type.REQUIRED)
+    @LRA(value = LRA.Type.REQUIRED, terminal = false)
+    // TODO Verify, this seems not used by any of the TCK tests
```
xstefank: valid, only used in one test that explicitly closes/compensates started LRA

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
```diff
@@ -267,7 +268,8 @@ public Response acceptWork(
 
     @PUT
     @Path("/supports")
-    @LRA(LRA.Type.SUPPORTS)
+    @LRA(value = LRA.Type.SUPPORTS, terminal = false)
+    // TODO Verify, this seems not used by any of the TCK tests
```
xstefank: my understanding is that `LRA.Type.SUPPORTS` should not end LRA by any means

mmusgrov: [The method is invoked here](https://github.com/eclipse/microprofile-lra/blob/5fdeb4c35a44e441395bde8feaf69b3fd459226e/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java#L298)

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
mmusgrov: The method is invoked by the [test that validates](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java#L565) that if the compensate callback method returns the `202 Accepted` HTTP status code then the implementation behaves correctly.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
mmusgrov: @xstefank the spec says the following about `LRA.Type.SUPPORTS`

>   *  If called inside a LRA context the managed bean method execution
>   *  must then continue inside this LRA context.

ie it does not say that the implementation is not allowed to end the LRA if one was present. In other words I do not see any problem with setting `terminal = true`.

rdebusscher:
thx @mmusgrov for pointing out where they are used. I didn't found those usages the first time. 

Removed the TODOs in the new commit.

PR #47, opened 2018-11-23
Add missing leaveUrl parameter to LRAClient#updateCompensator

xstefank:


mmusgrov:
@xstefank My point was that there are two join methods so you need to have two updateCompensator methods as well.

xstefank:
@mmusgrov should I then update this PR to add that method?

xstefank:
I will add it to this PR. If the outcome of the discussion will be to remove URLs methods I will update this PR to just add a new updateCompensator method

xstefank:
superseded by https://github.com/eclipse/microprofile-lra/pull/55

Issue #34, opened 2018-10-28
Add a time limit attribute to the @LRA and @Compensate annotations


mmusgrov:
The spec defines an annotation called @TimeLimit which can be applied to methods annotated with @LRA and @Compensate:

> The business logic may wish to control how long an LRA should remain active for before it becomes eligible for automatic cancellation by annotating methods that start LRAs with the @TimeLimit annotation.

and

> the ability to compensate may be a transient capability of a service so participants can be timed out. When the time limit is reached participants will be notified via their compensation (ie the method annotated with @Compensate)

It would be preferable for this behaviour to be specified as attributes of the @LRA and @Compensate annotations.

xstefank:
+1

tomjenkinson:
+1

mmusgrov:
@tomjenkinson it is preferable since:
- it then becomes clear that time limits only apply to compensate, complete and methods that start LRAs
  (ie with @TimeLimit the user may wonder what effect it would have on other methods);
- to a user who is looking at these annotations in an IDE it will be obvious that there is an option to apply time limits.

PR #50, opened 2018-11-27
issue34 Move @TimeLimit annotation to attributes of the other annotat…

mmusgrov:
…ions

Signed-off-by: Michael Musgrove <mmusgrov@redhat.com>

PR for #34


api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
```diff
@@ -36,4 +37,23 @@
...
+     * @return the period for which the participant can guarantee it
+     * will be able to complete when asked to do so
+     */
+    long timeLimit() default 0;
+
+    /**
+     * @return the unit of time that the {@link Complete#timeLimit()} attribute is
+     * measured in.
+     */
+    TimeUnit timeUnit() default TimeUnit.SECONDS;
```
xstefank: there is nothing wrong with adding this capability to Complete also but my concern is that LRAClient join method will now need two timelimit parameters to stay aligned with functionality available by the annotations - https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java#L278

Emily-Jiang: Use ChronoUnit instead of TimUnit to be consistent with other MP specs.

xstefank: We've decided on the mtg that LRAClient#joinLRA taking individual URLs and timelimit will be removed, so my previous comment is invalid

PR #51, opened 2018-12-03
Fix StandardController default response media type

xstefank:
StandardController returns an `Entity.text(...)` on error [1, 2]. Some frameworks, e.g. RestEasy, default response media type to `application/octet-stream` which in turn when the execution hits any of the above mentioned statements returns 

`Failed executing PUT /activities3/work: org.jboss.resteasy.core.NoMessageBodyWriterFoundFailure: Could not find MessageBodyWriter for response object of type: javax.ws.rs.client.Entity of media type: application/octet-stream`

[1] https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/StandardController.java#L52
[2] https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/StandardController.java#L64

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/StandardController.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/StandardController.java
@@ -49,7 +49,7 @@ public class StandardController {
     public Response work2(@HeaderParam(LRA_HTTP_HEADER) String lraId) throws NotFoundException {
 
         if (lraId != null) {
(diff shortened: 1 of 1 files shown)


xstefank:
I forgot to mention another possible fix is to add `@Produces(MediaType.TEXT_PLAIN) to StandardController definition. Neither of these fixes change the functionality of the tests, only aligning it to be compatible with jax-rs

Issue #69, opened 2018-12-12
TCK test changes for the new terminal attribute


mmusgrov:
Issue #40 removed the `delayClose` attribute of the LRA annotation since it effectively duplicated the `terminal` attribute. The fix for that issue also changed the default behaviour for closing an LRA at the end of the method. But some of the TCK test resources methods were changed to have a different semantic (in particular the @LRA annotation applied at class level provides did not specify the new attribute value).

Issue #40, opened 2018-11-12
remove delayClose() member on @LRA


rdebusscher:
The idea of the LRA spec is that multiple participants do some work related to an action initiated by the client.

So it doesn't make much sense to have a single method in an LRA. There will be always multiple services involved in the action.

so _delayClose()_ will need to be set to `true`, unless on the last participant where _terminal()_ is to `true`.

So by removing _delayClose_, it will result in a correcter and simplified usage of the API.

tomjenkinson:
This seems well with me it could well fit well with https://github.com/eclipse/microprofile-lra/issues/8

xstefank:
+1

mmusgrov:
+1

rdebusscher:
@mmusgrov I'll also update the default to `terminal = true`. 

The developer will need to set the value then explicitly to false (which he will forget a few times with all the consequences) when multiple participants. But then he is reminded that he should do the necessary steps to close it.

Quite a few entries in the [activityController](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java) which needs to be changed I think.

I'll include this in the PR.

PS: Not the renaming as this is handled by another issue.

PR #70, opened 2018-12-12
Issue69  TCK test updates (for the LRA terminal attribute)

mmusgrov:
#69 

Issue #40 removed the delayClose attribute of the LRA annotation since it effectively duplicated the terminal attribute. The fix for that issue also changed the default behaviour for closing an LRA at the end of the method. But some of the TCK test resources methods were changed to have a different semantic (in particular the @lra annotation applied at class level provides did not specify the new attribute value).

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
@@ -71,7 +71,7 @@ import static org.eclipse.microprofile.lra.client.LRAClient.LRA_HTTP_RECOVERY_HE
 
 @ApplicationScoped
 @Path(ActivityController.ACTIVITIES_PATH)
-@LRA(LRA.Type.SUPPORTS)
+@LRA(value = LRA.Type.SUPPORTS, terminal = false)
 public class ActivityController {
     public static final String ACTIVITIES_PATH = "activities";
     public static final String ACCEPT_WORK = "acceptWork";
@@ -106,7 +106,7 @@ public class ActivityController {
     @Path("/status")
     @Produces(MediaType.APPLICATION_JSON)
     @Status
-    @LRA(LRA.Type.NOT_SUPPORTED)
+    @LRA(value = LRA.Type.NOT_SUPPORTED)
     public Response status(@HeaderParam(LRA_HTTP_HEADER) String lraId) throws NotFoundException {
         Activity activity = activityService.getActivity(lraId);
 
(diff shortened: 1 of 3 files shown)


mmusgrov:
@rdebusscher Would it be possible for you to take a look at this small update?

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
```diff
@@ -106,7 +106,7 @@
     @Path("/status")
     @Produces(MediaType.APPLICATION_JSON)
     @Status
-    @LRA(LRA.Type.NOT_SUPPORTED)
+    @LRA(value = LRA.Type.NOT_SUPPORTED, terminal = false)
```
rdebusscher: Does it matter what the value for `terminal` is when NOT_SUPPORTED?

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
```diff
@@ -443,7 +443,7 @@ public Response getCompensatedCount() {
     @GET
     @Path("/cancelOn")
     @Produces(MediaType.APPLICATION_JSON)
-    @LRA(value = LRA.Type.REQUIRED, cancelOn = {Response.Status.NOT_FOUND, Response.Status.BAD_REQUEST})
+    @LRA(value = LRA.Type.REQUIRED, cancelOn = {Response.Status.NOT_FOUND, Response.Status.BAD_REQUEST}, terminal = false)
```
rdebusscher: `terminal` doesn't matter here I think since we are always returning a Response which leads to Cancel

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
```diff
@@ -455,7 +455,7 @@ public Response cancelOn(@HeaderParam(LRA_HTTP_HEADER) String lraId) {
     @GET
     @Path("/cancelOnFamily")
     @Produces(MediaType.APPLICATION_JSON)
-    @LRA(value = LRA.Type.REQUIRED, cancelOnFamily = {Response.Status.Family.CLIENT_ERROR})
+    @LRA(value = LRA.Type.REQUIRED, cancelOnFamily = {Response.Status.Family.CLIENT_ERROR}, terminal = false)
```
rdebusscher: `terminal` doesn't matter here I think since we are always returning a Response which leads to Cancel

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
mmusgrov: +1

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
mmusgrov: +1

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java
mmusgrov: +1

PR #72, opened 2018-12-17
Update startLRA(String, Long, ChronoUnit, Class<?>, URI, String) to r…

xstefank:
…eturn both lraId and recoveryUrl

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java
@@ -150,13 +150,14 @@ public interface LRAClient {
      * @throws GenericLRAException a new LRA could not be started. The specific
      * reason is available in {@link GenericLRAException#getStatusCode()}
      *
-     * @return the identifier of the new LRA
+     * @return the identifier of the new LRA and the recovery URL for 
+     * this enlistment
      *
      * @see LRAClient#startLRA(String, Long, ChronoUnit)
      * @see LRAClient#joinLRA(URL, Class, URI, String)
      */
-    URL startLRA(String clientID, Long timeout, ChronoUnit unit,
-                 Class<?> resourceClass, URI baseUri, String compensatorData)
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java
```diff
@@ -150,13 +150,14 @@ URL startLRA(String clientID, Long timeout, ChronoUnit unit)
...
-     * @return the identifier of the new LRA
+     * @return the identifier of the new LRA and the recovery URL for 
+     * this enlistment
      *
      * @see LRAClient#startLRA(String, Long, ChronoUnit)
      * @see LRAClient#joinLRA(URL, Class, URI, String)
      */
-    URL startLRA(String clientID, Long timeout, ChronoUnit unit,
-                 Class<?> resourceClass, URI baseUri, String compensatorData)
+    StartLRAResultPair startLRA(String clientID, Long timeout, ChronoUnit unit,
```
mmusgrov: I am not convinced adding a pair class here is the right approach. How about returning LRAInfo instead and update LRAInfo to include a method Collection<ParticipantInfo> getParticipants() (which we need anyway) where

```java
public interface ParticipantInfo {
    URL getRecoveryUrl();
}
```

Less desirable (but preferable to a Pair) would be to return a javax.ws.rs.core.Link with two relations, one for the LRA and one for the participant.

xstefank: @mmusgrov I like the first approach but maybe if ParticipantInfo would provide only a recoveryUrl I don't see a value by adding this as a wrapper so would this

```java
Collection<URL> getParticipants()
```

 be sufficient?

mmusgrov: I am okay with this but let's discuss it at the next hangout.

xstefank:
closing per discussion on the hangout

PR #73, opened 2018-12-17
issue77 Update joinLRA to return URL

xstefank:
the javadoc for these methods already states that the return value is a recovery URL

fix for issue #77 


mmusgrov:
Looks good but we should really have an accompanying issue.

xstefank:
@mmusgrov Thanks Mike, created in #77. Can you associate appropriate milestone please?

PR #75, opened 2018-12-27
Remove LRAParticipantDeserializer. Fixes #60.

### LINKED ISSUE #60: TCK tests time limits should be configurable
The TCK tests use the timeout which should be configurable. Even the timeout for the TCK is set to 50000 ms (https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java) it could be run on arbitrary fast machine. There is another hardcoded time limit values like at https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ActivityController.java#L502 where `renewTimeLimit` requires processing being done in 100ms. That could be difficult on a slower machine.

The time limits and timeouts should be configurable. Possibly with some time out factor as j2ee tck does.

rdebusscher:


rdebusscher:
Forgot Signed-off and amending commit doesn't fix it apparently.

Issue #80, opened 2019-01-07
Revert issue45 Add a client method that combines startLRA and joinLRA in a single call


xstefank:
As discussed on the hangout meeting, the return value of this method would impose significant changes to the spec which is why we will postpone this until it becomes a real performance burden for users.

mmusgrov:
PR was merged

Issue #45, opened 2018-11-19
Add a client method that combines startLRA and joinLRA in a single call


xstefank:
This is a proposal to add a method to the `LRAClient` and possibly `LRAManagement` that starts a new LRA and enlists calling service as a first participant in a single call. This can be a common scenario and as both `startLRA` and `joinLRA` can be remote invocations this method can save one unnecessary round trip.

ochaloup:
I like this idea. I think it's a good fit for specification. It provides way to reduce number of calls to the coordinator.

PR #81, opened 2019-01-07
[issue #80] Revert issue45 Add a client method that combines startLRA…

xstefank:
… and joinLRA in a single call

#45 and #80


mmusgrov:
Rudy this PR is reverting https://github.com/eclipse/microprofile-lra/issues/45 which added an startLRA API call that started an LRA and enlisted a participant in a single call. As we discussed in Mondays hangout we decided to leave this functionality for a later revision of the spec.

Issue #19, opened 2018-07-18
Include an API to obtain LRA stats


tomjenkinson:
If you include things like finishTime then make the spec clear that that should only be used after checking the state of the LRA as it may be initialized to a non-null value which is unreliable until after the LRA completes.

from mmusgrov: align with microprofile metrics and perhaps tracing

mmusgrov:
For example:

```java
public ManagementSPI {​
LRAInfo getStatus(LRA lra); // get status of one LRA
Collection<LRAInfo> getLRAs(CompensatorStatus status); // list LRAs in the given state
//anything else
}
```

mmusgrov:
@tomjenkinson The equivalent methods in the LRAClient API are used to implement the TCK and the proposal in issue #59 is to remove them in milestone 1.0. In order to have a viable TCK for milestone 1.0 I would like to move this issue (#19) from milestone 1.x to 1.0

tomjenkinson:
Did we agree this was going into the TCK rather than the api? If so then I would concur with you on that milestone, if you want it in the API we need to discuss on hangout because I think management API should not be part of the spec API

mmusgrov:
If we do not include it in the API how does a client know when the LRA has closed or cancelled?

tomjenkinson:
I think they return a status ideally or some proprietary method is used as per JTA?

mmusgrov:
OK, I added a comment to PR #82 to match your expectation and ask if Rudy will take a look, thanks.

mmusgrov:
@tomjenkinson with respect to the description of this issue:

> If you include things like finishTime then make the spec clear ...

I decided not to make any changes to LRAInfo and keep fixes for this issue to a minimum. If any stakeholders have an opinion about the information provided by LRAInfo then we can discuss that in a future hangout.

Issue #59, opened 2018-12-05
Remove query methods from LRAClient


tomjenkinson:
These are:
```
* getActiveLRAs
* getAllLRAs
* getRecoveringLRAs
* isActiveLRA
* isCompensatedLRA
* isCompletedLRA
* getStatus
```

They could be used by the TCK if it was in an SPI method as per https://docs.google.com/document/d/1HBPiXpsXcQp9T8JeoT-IDs-EblZX1SxuTV5qjUPzwrM/edit#bookmark=id.59ji03x3hyfp

mmusgrov:
The equivalent functionality has been moved to https://github.com/eclipse/microprofile-lra/issues/19

PR #82, opened 2019-01-10
issue19 and issue59: Move query methods from the API into the TCK

mmusgrov:
Fixes issues #19 and #59


api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java
```diff
@@ -188,120 +185,20 @@ URL startLRA(String clientID, Long timeout, ChronoUnit unit,
...
-     *
      * @throws NotFoundException if the LRA no longer exists
      *
      * @throws GenericLRAException Communication error (the reason is availalbe via
      * the {@link GenericLRAException#getStatusCode()} method
      *
      * @return the response MAY contain the final status of the LRA as reported by
+     *         {@link LRAStatus#name()}
+     *         If the final status is not returned the
+     *         client can only discover the final state using proprietary APIs
```
mmusgrov: @tomjenkinson This is where the problem with moving the query API into the TCK lies

tomjenkinson: Can you expand on that a little please? I imagine proprietary is OK because that would be the same for JTA JTS etc right?

mmusgrov: So the caller uses the client API to close an LRA and gets back something like status `Closing` (since some participants may still be completing). But now the caller can never discover the final outcome of his request to close the LRA since there is no interface for that.

mmusgrov: Rudy, there is also some discussion about this in the issue #19 where I accepted Tom's point of view.

api/src/main/java/org/eclipse/microprofile/lra/annotation/LRAStatus.java
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * The LRA is currently informing participants that they should
+     * compensate for any work they performed when the LRA was active
+     */
+    Cancelling,
+    /**
+     * All participants associated with the LRA have successfully
+     * compensated for any work they performed when the LRA was active
+     */
+    Compensated,
```
xstefank: can this be `Cancelled` to stay uniform with rest compensation states?

api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java
tomjenkinson: I imagine vendors could provide some method to determine the outcome of a call that returned "Closing" - so that situation is fine fom spec point of view IMO

xstefank: I think we should not be encouraging usage of proprietary APIs. At least as I understand it, MicroProfile aims to have an easily replaceable implementations without the need of code changes in the user code. So users should be able to get all required information through LRAClient API. For this use-case maybe we can instead return Future<LRAStatus> or make the call blocking until definite result can be provided?

tomjenkinson: I like the idea of the future for normal use case.

There would still be some need for a proprietary or otherwise way to deal with a client application that crashed after issuing the complete but before being able to obtain the result.

api/src/main/java/org/eclipse/microprofile/lra/annotation/LRAStatus.java
mmusgrov: Good spot, thanks

api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java
mmusgrov: Returning a Future is going to be difficult to implement in all cases as it needs to remain valid potentially forever. The `Closing` state is provided to indicate that the request is in progress which therefor obviates the need for Futures.

tomjenkinson: Couldn;t the future return an exception (NoLongerAvailable) if it is polled at some point in violation of the protocol?

mmusgrov: I mean that it is valid for a participant to take an arbitrarily long period to complete.

There is also the issue of the client failing which would mean he does not have access to the future when he restarts.

And I still stand by my point that if we block or return a future from the close call then the Completing states become redundant.

xstefank: both valid points. I would also be more inclined to keep single `getStatus(lraId)` method. Can someone reiterate why we should remove it please?

tomjenkinson: I was thinking some implementations might not be able to provide it immediately but perhaps "Try again later" can be a return code so long as it is certain at some point it could provide it.

I am considering "routing slip" or "event sourcing" potential implementations

xstefank: I think both of these concepts are capable of providing LRA status information. In my view LRAClient#getStatus provides a client view of the status of the LRA (note that saga is defined as eventually consistent) so I don't see any problem if for instance routing slip will be returning Cancelling/Completing until the slip is again received by this particular service.

tomjenkinson: what status could it return if it was disconnected from the slip and recovered with a reference to the LRA ID though, it would not know if it was cancelling or completing. I think something like "Try later" might be needed for that scenario

tomjenkinson: maybe 503

xstefank: Understood, but I would still stand by the fact that client.getStatus() provides a client view of the LRA. If it doesn't know, it can say that (e.g. again by exception). In orchestration, if both client and coordinator fail and client recovers first with the lraId but the coordinator is still down we would have similar situation right?

tomjenkinson: yes, we would

mmusgrov: How are we going to run the TCK against an implementation that does not contain an API for reporting the status of an LRA?

xstefank: we will definitely need such method. I believe the question discussed is whether it will be in the spec API or in the TCK SPI.

tomjenkinson: It is the question

mmusgrov: If the implementation can provide one for the TCK why can't it also provide one for the Client API too?

mmusgrov:
@tomjenkinson The commit (40f18f6) I just pushed reinstates the LRAClient#getStatus() method. Will you check it so that I can merge this PR.

api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java
```diff
@@ -379,4 +277,22 @@ URL updateCompensator(URL recoveryUrl, Class<?> resourceClass, URI baseUri, Stri
...
+     * @throws NotFoundException if the LRA no longer exists
+     *
+     * @throws UnknowableException if the status is unknowable
+     *
+     * @throws GenericLRAException if the request to the coordinator failed.
+     * {@link GenericLRAException#getCause()} and/or
+     * {@link GenericLRAException#getStatusCode()}
+     * may provide a more specific reason.
+     */
+    ParticipantStatus getStatus(URL lraId) throws GenericLRAException;
```
tomjenkinson: It seems to me inconsistent to return a specific participantstatus from a getStatus method that describes itself capable to lookup an LRA and return the status of said item.

mmusgrov:
I pushed a version that returns an instance of LRAStatus instead

mmusgrov:
I squashed the last 2 commits

api/src/main/java/org/eclipse/microprofile/lra/client/LRAClient.java
```diff
@@ -379,4 +277,22 @@ URL updateCompensator(URL recoveryUrl, Class<?> resourceClass, URI baseUri, Stri
...
+    /**
+     * Lookup the status of an LRA
+     *
+     * @param lraId the LRA whose status is being requested
+     *
+     * @return the status of the the LRA.
+     *
+     * @throws NotFoundException if the LRA no longer exists
+     *
+     * @throws UnknowableException if the status is unknowable
```
xstefank: I miss here a case where LRA client doesn't know current state right now but can potentially provide this information in the future. Some use-cases -- coordination is down and then comes up again, routing slip is received by this service by the next call (deducing from javadoc of UnknowableException)

Issue #83, opened 2019-01-11
Move LRAInfo interface into the TCK package


mmusgrov:
#19 is removing the "query based" methods from the LRAClient interface (and adding them to the TCK instead). Since these LRAClient methods are the only ones using the LRAInfo class it is makes sense to move it to the TCK package too.

During the move it would be helpful to clean up the LRAInfo interface (for example add a single method to obtain the status of the LRA rather than multiple methods).

Also the meaning of the clientId should be clarified in the javadoc

PR #89, opened 2019-02-05
issue83 Move LRAInfo into TCK

mmusgrov:
Fixes issue #83 
The class LRAInfo is only ever used by the TCK so it should be moved into that module

The fix also fixes a checkstyle issue for an unused import in LRAClient that was introduced in PR #82


ochaloup:
+1

mmusgrov:
@rdebusscher and @xstefank Thank you for the speedy review.

PR #90, opened 2019-02-05
[issue59] after getAll method moved to TCK

ochaloup:
there is no need for this import which causes checkstyle violation

https://github.com/eclipse/microprofile-lra/pull/82

/cc @mmusgrov


ochaloup:
sorry, I missed the follow-up PR

mmusgrov:
@ochaloup thanks

PR #91, opened 2019-02-05
Removing getStatus method from ManagementSPI as it's duplicate of one in LRAClient

ochaloup:
Creating PR as follow-up to https://github.com/eclipse/microprofile-lra/pull/82 as there is duplicate method at spi and `LRAClient`.


mmusgrov:
I think I would like to keep the SPI getStatus method since it will eventually be used to return more than just the LRA status (it is for the stats requirements)

mmusgrov:
@xstefank renaming is good. @ochaloup the class contains more than the status such as the clientId and isTopLevel and we need to use those in TCK tests. The stats to count the LRAs is going to be useful too.

ochaloup:
Closing as `ManagementSPI` will be in different discussion. Replaced by PR#93

Issue #84, opened 2019-02-01
Define behaviour when @LRA annotated method throws an exception


ochaloup:
The specification defines that a method annotated with `@LRA` finishes the LRA (either with `complete` or `cancel`). There is not defined what happens when such method is ended by a thrown exception.
It would be good to define this behaviour.

JTA java developers are used to how the JTA works in way that `RuntimeException` rolls-back the transaction while the checked exceptions do not. The CDI [@Transactional](https://docs.oracle.com/javaee/7/api/javax/transaction/Transactional.html#rollbackOn--) defines what are exceptions that rolls-back the transaction then. The similar approach could be taken for @LRA as well.

ochaloup:
on discussion on meeting 2019-02-04: any exception thrown from the `@LRA` annotated method cancel the LRA. The `cancelOn`/`cancelOnFamily` controls to `cancel` the LRA based on the error code returns from the `@LRA` annotated method which is the REST endpoint.

ochaloup:
I was thinking to create a PR on this but I found one case I'm not sure how that should be handled. What if the exception is thrown from the method which does not started the LRA context.

```java
// -> incoming LRA context with already started LRA
@LRA(Type.SUPPORTS)
public void method() {
  throw new RuntimeException();
}
```

Will this cancel the LRA context which came from outer space? Or throwing the exception will cause to cancel the LRA only when started at the same method?

xstefank:
@ochaloup I believe that depends on the value of the `end` attribute of the `@LRA` annotation

ochaloup:
If you have non-exception case like microservice A calls a microservice B then

```
@LRA(Type.REQUIRES_NEW)
public void microserviceAMethod() {
   ClientBuilder.newClient().target(<<microserviceBMethod>>).post();
}

@LRA(Type.SUPPORTS, end = true)
public void microserviceBMethod() {
  // do some business logic
}
```

The `end` is by default to `true`, this fact is emphasized at the `microserviceBMethod`. When the `microserviceBMethod` and returns back to microservice B the LRA is not ended. The LRA will be ended at the end of the `microserviceAMethod`. It means the responsibility for ending the LRA is on he method that started the original context. How is it with exception that would be thrown in `microserviceBMethod`. Will the microservice B cancel the LRA even it only received the context and the B does not start it? Or it should be responsibility of microservice A (`microserviceAMethod` to handle the error codes and then decides to throw an exception (and cancel the LRA) or ignore the errors?

[...]

PR #99, opened 2019-02-18
issue84: canceling LRA when exception is thrown from @LRA annotated method

ochaloup:
fixes https://github.com/eclipse/microprofile-lra/issues/84

clarification of the rules for ending the `@LRA` annotation when using `cancelOn`/`cancelOnFamily`. adding tck tests to verify the behaviour

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/LRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/LRA.java
@@ -20,7 +20,11 @@
 
 package org.eclipse.microprofile.lra.annotation;
 
+import javax.ws.rs.HeaderParam;
 import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.lra.client.LRAClient;
+
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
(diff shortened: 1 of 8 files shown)


mmusgrov:
How does this test differ from the existing cancelOn and cancelOnFamily TCK tests?

ochaloup:
@mmusgrov in functionality it does not differ. Those two tests which were part of the TckTests are rip off and moved to the separate test class that gathers all of the cancelOn cases. The new tck class is then enriched with new scenarios.
It would be possible to put all the tests (aka. adding the new ones) to the `TckTests` class but I think it's already quite huge and transforming it for having multiple test classes each with separate responsibility is a good move. Especially I think we need to add more tck tests in general and having all of them in one test class does not seem to me a good practice.

ochaloup:
@xstefank @mmusgrov @rdebusscher I updated the javadoc with clarification about `end` attribute as discussed on the meeting. Please, let me know if something is wrong or could be done better. Thanks.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
```diff
@@ -0,0 +1,198 @@
...
+
+    @Before
+    public void before() {
+        LOGGER.info("Running test: " + testName.getMethodName());
+        this.beforeCompletedCount = getCompletedCount();
+        this.beforeCompensatedCount = getCompensateCount();
+    }
+
+    private WebTarget getSuiteTarget() {
+        String tckSuiteBaseUrl = System.getProperty("lra.tck.base.url");
```
xstefank: why are you getting this directly when you are injecting this config property in `LraTckConfigBean`?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    /**
+     * See {@link LraCancelOnController#cancelOnFamily3xx()}
+     */
+    @Test
+    public void cancelOnFamily3xx() {
+        Response response = getSuiteTarget().path(LRA_CANCEL_ON_CONTROLLER_PATH)
+                .path(LraCancelOnController.CANCEL_ON_FAMILY_3XX).request().get();
+        assertEquals("The 303 status response is expected", Status.INTERNAL_SERVER_ERROR.getStatusCode(), response.getStatus());
```
xstefank: INTERNAL_SERVER_ERROR is 500

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
```diff
@@ -0,0 +1,198 @@
...
+        assertEquals("Even the 200 status was received the remotly called participant should cause the LRA being cancelled",
+                beforeCompletedCount, getCompletedCount());
+    }
+
+
+
+    private int getCompletedCount() {
+        Response response = getSuiteTarget().path(LRA_CANCEL_ON_CONTROLLER_PATH)
+                .path(LraCancelOnController.COMPLETED_COUNT_PATH).request().get();
+        return Integer.valueOf(response.readEntity(String.class));
```
xstefank: you can also use directly `response.readEntity(Integer.class)`

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
ochaloup: this is a think I'm not convinced I did in the best way. The `TckCancelOnTests` are written to be run on client (`@RunAsClient`). It means it runs in the surefire/failsafe JVM and there is no container which would manage the property injection. I think the MicroProfile config projects can be used as standalone libraries but the configuration could be a bit difficult. Then I took this approach for the tests on the client.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
ochaloup: ouch, thank you for catching this

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
ochaloup: thanks!

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
tomjenkinson: Thanks for the clarification - it seems reasonable to me

xstefank: Thanks. This is ok with me. We can take look after merge of this PR. But maybe we could add a note to the setup document?

ochaloup: @xstefank good point. I provided the default value as defined in the document and added a sentence mentioning this.

api/src/main/java/org/eclipse/microprofile/lra/annotation/LRA.java
```diff
@@ -184,20 +226,37 @@
...
-     * HTTP response codes will cause the current LRA to cancel. By default client
-     * errors (4xx codes) and server errors (5xx codes) will result in
-     * cancellation of the LRA.
+     * HTTP response codes will cause the current LRA to cancel.
+     * This expects the annotated method is the JAX-RS endpoint where the {@link Response}
+     * with the status code could be returned.
+     * </p>
+     *
+     * <p>
+     * By default client errors (<code>4xx codes</code>) and server errors (<code>5xx codes</code>)
```
mmusgrov: Will you also set this default on the element please.

ochaloup: @mmusgrov I'm sorry but I'm not sure what `element` you mean. Would you be so kind to elaborate where I can add the info?

mmusgrov: The cancelOnFamily element in the annotation class LRA.java needs a default, ie you should add:

```java
Response.Status.Family[] cancelOnFamily() default {Response.Status.Family.CLIENT_ERROR, Response.Status.Family.SERVER_ERROR};
```

ochaloup: good point. I added that there.

ochaloup:
@rdebusscher I updated headers of the new files to be on 2019. I absolutely agree with the #101. I didn't realize it. Thank you.

Issue #11, opened 2018-06-25
Include jaxrs in the annotations package name


mmusgrov:
The LRA annotations only apply to JAX-RS resources so the package name should include jaxrs

PR #104, opened 2019-03-03
issue11 include jaxrs in the package hierarchy

mmusgrov:
#11 

This is just a move of three files and I have followed the package naming used by JAX-RS, namely ws.rs


api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/Leave.java
```diff
@@ -18,7 +18,7 @@
  * limitations under the License.
  *******************************************************************************/
 
-package org.eclipse.microprofile.lra.annotation;
+package org.eclipse.microprofile.lra.annotation.ws.rs;
```
xstefank: @mmusgrov According to [this comment](https://github.com/eclipse/microprofile-lra/issues/11#issuecomment-442098892) only `@LRA` and `@NestedLRA` should be covered. I believe we should stay consistent with handling of participant annotations (`@Compensate`, `@Complete`, `@Status`, `@Forget`, `@Leave`).

Issue #109, opened 2019-03-06
Clarify multiple registrations by the same participant


xstefank:
This issue covers `TckCancelOnTests#cancelFromRemoteCall()` test which makes multiple registrations of the same participant (`LraCancelOnController`) and tests that compensate method has been called several times. This differs from already defined behavior in some tests of `TckTests` class so I would like to propose to clarify this behavior in the spec text or javadoc so that multiple registration of the same participant are ignored and refactor `TckCancelOnTests#cancelFromRemoteCall()` to register two separate participant resources.

xstefank:
@mmusgrov 

> This spec currently requires that participants can only be invoked via JAX-RS endpoints. Since the endpoints uniquely identify the resource and since calling the compensate/complete multiple times is allowed (if there is no status method) there is no ambiguity about multiple registrations.

In here [1] TckTests explicitly test that multiple registration by the same participant are ignored. Additionally `LraController`. However, there is `@Status` annotated method in `LraController`.

If this is the behavior you would prefer it's still requires clarification in the specification. I will create a new PR then with specifying explicitly that multiple registrations of same participant within the same LRA are allowed if there no `@Status` method and there is no need to change TckCancelOnTests then.

> When we change the spec to allow non-JAX-RS based participants can you re-submit the PR and update the issue description to clarify that this issue does not apply to JAX-RS based participant resources.

I don't see why the situation would be any different?

[1] https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java#L435

mmusgrov:
The spec is unambiguous. If there is no @Status annotation the compensate endpoint will be repeatedly called until it reaches one of the end states.

xstefank:
@mmusgrov I still think this may be confusing for users (and implementors) if it is not going to be mentioned explicitly. We are assuming that from the fact you've described it's easy to derive that multiple registrations work when there is no Status method and don't otherwise. But this can be just my view so if you prefer I will close the issue and the PR.

mmusgrov:
The situation is that there is a unique JAX-RS endpoint for the registration in the context of a particular LRA. Therefore a second registration would just be registering the same endpoint ie it would effectively be a null op and therefore there is no need to add any explanatory text to the spec. Unless of course what I have just written is not clear from a reading of the spec.

xstefank:
[...]

Issue #92, opened 2019-02-06
Add TCK tests for MicroProfile LRA 1.0


xstefank:
I've got this idea from [MP concurrency](https://github.com/eclipse/microprofile-concurrency/issues/60). I would like to propose this issue as tracker for individual scenarios that should be included in the LRA TCK. Please when you create a PR for any scenario prefix this scenario with github pull request number and once it's merged please check its checkbox.

To add new scenario:

1. Propose scenario on this issue as comment or better directly in the agenda for the Monday hangout
2. The team will review new scenarios in Monday hangout 
3. If there are no objections it will be added to this list

Current scenarios:

- [x] #2 timelimit
- [x] #2 start LRA
- [x] #2 cancel LRA
- [x] #2 close LRA
- [x] #2 one level nested activity
- [x] #2 multi level nested activity complete
- [x] #2 multi level nested activity compensate
- [x] #2 multi level nested activity mixed complete and compensate outcomes
- [x] #2 join LRA via HTTP header
- [x] #2 join LRA when new one is started (REQUIRED, REQUIRES_NEW)
- [x] #2 leave LRA via annotation
- [x] #2 management of LRA in not annotated resource
- [x] #2 LRA recovery
- [x] #10 non LRA aware resource LRA propagation
- [X] #119 Multiple registrations by same participant within the same LRA should be ignored and related tests
- [x] #2 cancel on response code
- [x] #2 cancel on response code family
- [x] #99 cancel on family default -- 4xx really cancels LRA
- [x] #99 cancel on family default -- 5xx really cancels LRA
- [x] #99 cancel on family set to 3xx really cancels LRA
- [x] #99 cancel on family changed default is respected (i.e. change default to just 4xx and verify that 5xx code doesn't cancel LRA)
- [x] #99 cancel on set to precise status code really cancels LRA
- [x] #99 cancel on remote call -- call participant which starts the LRA and then makes a REST call to second `@LRA` annotated method which cancels the LRA through cancel on attributes
- [x] #122 add tests to validate that the LRA context is propagated correctly
- [ ] #117 Check that completed nested participants are told to forget when the top level LRA finishes
- [ ] #6 Include tests for time limits
- [ ] please add more scenarios here 

----
Deferred to 1.x (will open new issue after 1.0 release)
- [x] #2 leave LRA via client API
- [x] #2 get Active LRAs
- [x] #2 get all LRAs
- [x] #2 is particular LRA active
[...]

PR #110, opened 2019-03-06
issue109 Clarify multiple registrations by the same participant

xstefank:
issue #109 

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
@@ -20,6 +20,7 @@
 package org.eclipse.microprofile.lra.tck;
 
 import static org.eclipse.microprofile.lra.tck.participant.api.LraCancelOnController.LRA_CANCEL_ON_CONTROLLER_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.LraCancelOnRemoteCallController.LRA_CANCEL_ON_REMOTE_CALL_CONTROLLER_PATH;
 import static org.junit.Assert.assertEquals;
 
 import java.net.MalformedURLException;
@@ -34,6 +35,7 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.Response.Status;
 
 import org.eclipse.microprofile.lra.tck.participant.api.LraCancelOnController;
+import org.eclipse.microprofile.lra.tck.participant.api.LraCancelOnRemoteCallController;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.RunAsClient;
 import org.jboss.arquillian.junit.Arquillian;
@@ -169,12 +171,12 @@ public class TckCancelOnTests {
     }
 
     /**
-     * See {@link LraCancelOnController#cancelFromRemoteCall()}
+     * See {@link LraCancelOnRemoteCallController#cancelFromRemoteCall()}
      */
     @Test
     public void cancelFromRemoteCall() {
-        Response response = getSuiteTarget().path(LRA_CANCEL_ON_CONTROLLER_PATH)
-                .path(LraCancelOnController.CANCEL_FROM_REMOTE_CALL).request().get();
+        Response response = getSuiteTarget().path(LRA_CANCEL_ON_REMOTE_CALL_CONTROLLER_PATH)
+                .path(LraCancelOnRemoteCallController.CANCEL_FROM_REMOTE_CALL).request().get();
         assertEquals("The 200 status response is expected", Status.OK.getStatusCode(), response.getStatus());
         assertEquals("Status was 200 but compensate should be called twice as LRA should be cancelled for remotely called participant as well",
                 beforeCompensatedCount + 2, getCompensateCount());
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/activity/ParticipantService.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/activity/ParticipantService.java
@@ -1,0 +1,63 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


ochaloup:
+1 from me

mmusgrov:
-1

This spec currently requires that participants can only be invoked via JAX-RS endpoints. Since the endpoints uniquely identify the resource and since calling the compensate/complete multiple times is allowed (if there is no status method) there is no ambiguity about multiple registrations.

When we change the spec to allow non-JAX-RS based participants can you re-submit the PR and update the issue description to clarify that this issue does not apply to JAX-RS based participant resources.

xstefank:
@mmusgrov I will reply in the issue, PR should be left for reviews

xstefank:
@mmusgrov @rdebusscher @ochaloup I've changes in the spec text. The PR now only changes cancelOnRemoteCall test to align with the spec expectations on multiple registrations. I will extract the multiple registration test from leaveLRA tests to separate test and propose this in #92

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -210,6 +210,8 @@ enter the state model until it receives a complete or compensate message:
 .Participant State Model
 image::participant-state-model.png[align="center" link="participant-state-model"]
 
+A single participant may be concurrently enlisted within multiple
+LRAs. However, it CAN NOT be enlisted multiple times within the same LRA.
```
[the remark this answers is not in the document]
xstefank: @mmusgrov I thought I removed spec text changes. It also says 'outdated' for me. Maybe error on your side?

xstefank:
@mmusgrov I've added all cancel on scenarios to issue #92

ochaloup:
still +1 for this PR from me

Even I understand the spec text should be concise I think still think that emphasizing the behaviour in this area is beneficial and not much redundant.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
```diff
@@ -169,12 +171,12 @@ public void notCancelOnFamily5xx() {
...
      */
     @Test
     public void cancelFromRemoteCall() {
-        Response response = getSuiteTarget().path(LRA_CANCEL_ON_CONTROLLER_PATH)
-                .path(LraCancelOnController.CANCEL_FROM_REMOTE_CALL).request().get();
+        Response response = getSuiteTarget().path(LRA_CANCEL_ON_REMOTE_CALL_CONTROLLER_PATH)
+                .path(LraCancelOnRemoteCallController.CANCEL_FROM_REMOTE_CALL).request().get();
         assertEquals("The 200 status response is expected", Status.OK.getStatusCode(), response.getStatus());
         assertEquals("Status was 200 but compensate should be called twice as LRA should be cancelled for remotely called participant as well",
                 beforeCompensatedCount + 2, getCompensateCount());
```
rdebusscher: Why do we assume here the compensate is called twice? A single LRA should only be compensated once (in the case the first call returned status 200 as we do here)

xstefank: there are 2 participants subscribed within this test:

test calls A, A calls B, B returns cancel on status (5xx) so the LRA should be cancelled. So both A and B should call compensate

mmusgrov: What is the status of this PR. To me the test looks wrong, the same resource is invoked twice in the context of the same LRA and therefore there should be only a single participant registered with the LRA, hence a single compensate callback. I recommend either deleting the test or replace it with something that we aren't already testing.

mmusgrov: An analogy with XA would be if a client started a transaction and made two updates to a database and to expect that the XA resource is asked to commit twice for the same transaction which would be incorrect behaviour.

mmusgrov: My advice is to change the test to assert that the compensate call is only issued once:

```java
assertEquals("Compensate was called more than once on the same registered participant",
                beforeCompensatedCount + 1, getCompensateCount());
```
Then the test is a good test because it validates that the implementation does not issue duplicate compensate notifications (I don't believe we currently have a test for that).

xstefank: @mmusgrov your assumption is not right, there are two resources enlisted within the LRA -- LraCancelOnRemoteCallController and LraCancelOnController. They just both share an application scoped bean ParticipantService which serves as a shared counter to verify that compensate has been called on both participants. 

I am not saying the test case is right. It is something that was already accepted as part of cancel on tests. This is really just refactoring because as we've discussed on the meeting the same participant cannot be enlisted within the same LRA twice.

mmusgrov: I am looking at the code in the cancelFromRemoteCall test that is currently in the repo which only calls a single participant. If you have changed the test to call two different participants then I would have thought we already have a test for that. We should just change the assert along the lines I suggested above.

rdebusscher: @xstefank I missed that  ParticipantService is shared between the two resources.

But the test is not deterministic then. Was there a compensate for A and B? or was it just Compensate B which was called twice, or Compensate A twice?

xstefank: @rdebusscher yes you are right. I didn't think that would be a problem. But I can update ParticipantService to take into account where the compensate has been called.

mmusgrov: @ochaloup Can you raise a bug fix for the existing test to change the assert to 1. That way our implementations will start passing again, That way we can continue this discussion at our leisure.

ochaloup: @mmusgrov changing the assert to 1 is not the correct solution. It could be that the Narayana implementation will start to pass but the test will be wrong. We can't drive the specification based on what the implementation expects.

Summary of the issue
* I wrote the test `cancelFromRemoteCall` which tested two separate things "cancel on remote call" and "multiple registration of the same participant"
* this PR changes the test `cancelFromRemoteCall` in the correct way that the test will be testing only the behaviour of the "cancel on remote call"
* the issue #109 was created based on our discussion to clear the specification on this and to fix the failing tests. 
* this  PR was gradually changed to contain only  fix for the test `cancelFromRemoteCall
* this PR fixes the `cancelFromRemoteCall` and I think it fixes it in the good manner
* we possibly should consider adding a specific test for the "clarify multiple registrations by the same participant"
* I would be glad if we can shortly talk on the hangout if the spec is really clear in the manner of the issue #109

xstefank:
@rdebusscher @mmusgrov @ochaloup 

I will close this PR because as most of you already said it diverged from its original intent a lot.

I will create a new issue for the cancelOnRemoteCall test fix and create a new PR for it incorporating changes @rdebusscher requested (I have it already done based on this PR)

I will keep issue #109 open for now and we can discuss on the meeting what if any action should be taken.

PR #118, opened 2019-03-18
[issue#109] clarify multiple enlistment of the same participant

### LINKED ISSUE #109: Clarify multiple registrations by the same participant

ochaloup:
to the LRA and fixing test expecting the wrong behaviour

fixes #109

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -210,6 +210,9 @@ enter the state model until it receives a complete or compensate message:
 .Participant State Model
 image::participant-state-model.png[align="center" link="participant-state-model"]
 
+The participant can be enlisted only once per LRA instance. When participant
+is about to be enlisted multiple times (e.g. client calls the same `@LRA` method
+several times) the follow-up enlistments of such participant MUST be ignored.
 
 Different usage patterns for LRAs are possible, for example LRAs may be
 used sequentially and/or concurrently, where the termination of one LRA
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckCancelOnTests.java
@@ -176,8 +176,9 @@ public class TckCancelOnTests {
         Response response = getSuiteTarget().path(LRA_CANCEL_ON_CONTROLLER_PATH)
                 .path(LraCancelOnController.CANCEL_FROM_REMOTE_CALL).request().get();
         assertEquals("The 200 status response is expected", Status.OK.getStatusCode(), response.getStatus());
-        assertEquals("Status was 200 but compensate should be called twice as LRA should be cancelled for remotely called participant as well",
-                beforeCompensatedCount + 2, getCompensateCount());
+        // LraCancelOnController enlists twice the same participant, compensate is expected to be called only once
+        assertEquals("Status was 200 but compensate should be called as LRA should be cancelled for remotely called participant as well",
+                beforeCompensatedCount + 1, getCompensateCount());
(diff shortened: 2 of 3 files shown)


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -210,6 +210,9 @@ enter the state model until it receives a complete or compensate message:
 .Participant State Model
 image::participant-state-model.png[align="center" link="participant-state-model"]
 
+The participant can be enlisted only once per LRA instance. When participant
+is about to be enlisted multiple times (e.g. client calls the same `@LRA` method
+several times) the follow-up enlistments of such participant are to be ignored.
```
```suggestion
several times) the follow-up enlistments of such participant MUST be ignored.
```
just to put an emphasis on what we expect?

ochaloup: +1

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -210,6 +210,9 @@ enter the state model until it receives a complete or compensate message:
 .Participant State Model
 image::participant-state-model.png[align="center" link="participant-state-model"]
 
+The participant can be enlisted only once per LRA instance. When participant
+is about to be enlisted multiple times (e.g. client calls the same `@LRA` method
+several times) the follow-up enlistments of such participant MUST be ignored.
```
tomjenkinson: Request for clarification, same business logic is deployed to a cluster (like kubernetes) but different instances (e.g. pods) respond to same rest endpoint (e.g. /book) which was annotated with `@Lra`. I don't think it is feasible to prevent the second call from enlisting a new resource.

xstefank: I believe it is because they will have the same identification which needs to be globally unique (e.g., in case of narayana the participant URIs/URLs will be the same for all pods as they are behind same load balancer or service in kubernetes)

tomjenkinson: Good point, I guess compensation URL of www.foo.com/<service_name>/compensation will attempt to be registered with a specifically identified LRA instance. Something for implementations to watch out for and not try to optimize for using verification on the client side alone if their implementation uses a client/server architecture.

ochaloup: it's good point, I wasn't thinking this way when we were discussing this on the meeting. But still I agree with what Martin wrote here. The identification should be globally unique.

mmusgrov: I agree with Martin, if the application cannot be coded to deal with multiple compensate/complete notifications then it is his responsibility to ensure the participant is registered with a unique id.

tomjenkinson: If the client side of the participant identification is somehow unique it is not cluster aware.

Also, it should not be getting multiple compensate notifications (for this particular requirements) as "follow-up enlistments of such participant MUST be ignored". Anyway this is really now just a note for implementations that they cannot expect to be able to process this requirement entirely on the client side (in client-server architectures).

mmusgrov: Tom, I merged this PR but then realised it wasn't clear as to whether or not you were asking for changes?

tomjenkinson: No, it's fine thanks

Issue #122, opened 2019-03-20
Requirement of @LRA on @Leave annotated method: Clarify


rdebusscher:
Javadoc for `@Leave` for example 

`invoking such a method in the context of the same LRA that it has joined will result in the resource leaving the LRA (ie it will not receive any callbacks when the LRA is later closed or cancelled)
`

So, in order to know which is the LRA, the request needs to contain the LRA_HTTP_HEADER ,identifying the LRA id.
So there is no need to join the same LRA before the functionality of @Leave can be performed.

Using LRA can create even more confusion and errors, for example, what should happen in this case?
```
@LRA(type=REQUIRES_NEW)
@Leave
```

xstefank:
@rdebusscher exactly and I think we already don't require `@LRA` on `@Leave` methods [1] - the TCK test is already passing.

IMO `@Leave` should work only with LRA_HTTP_HEADER so that LRA_HTTP_HEADER should be present on invocation to indicate which LRA is concerned. If the header is not provided by user invoking the leave method than current context should be used and injected by the implementation to LRA_HTTP_HEADER.

[1] https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraController.java#L157

rdebusscher:
@xstefank I saw your comment on a PR and based on that I created this issue. (because your comment was not linked to the PR content but valid)

mmusgrov:
@rdebusscher The annotations are for the end user to define his transaction boundaries. If there is no such annotation then the implementation is allowed to ignore or remove the header ie it cannot be relied upon to be present without the LRA annotation. This rule ought to make using spec less error prone / easier to use.

But I will update the spec to make this clear. (Remark: The header is still useful to clients since they are allowed to manually add the header in order to propagate the context to target resource methods that are annotated with @LRA, I will make sure that is clear too).

With regards to @xstefank comment about the test resource - that is a bug and we need a test to ensure that the implementation ignores the context if there is no @LRA annotation.

mmusgrov:
@rdebusscher But to answer your question: typically the application would use SUPPORTS, REQUIRED or MANDATORY. I did want to have the annotations driving everything but if we cannot do that by simply documenting the expected behaviour then, yes, let's nail down the semantics of the header along the lines that Martin assumed.

mmusgrov:
Re-opened for TCK tests

mmusgrov:
[...]

PR #124, opened 2019-03-24
issue122 Clarify LRA context on @Leave methods

mmusgrov:
#122 

The PR adds some text to clarify the LRA context and in particular clarifies the interaction between `@Leave` and `@LRA`

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
@@ -27,7 +27,7 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * LRA annoations support distributed communications amongst software
+ * LRA annotations support distributed communications amongst software
  * components and due to the unreliable nature of networks,
  * messages/requests can be lost, delayed or duplicated etc and the
  * implementation component responsible for invoking {@link Compensate}
@@ -53,6 +53,13 @@ import java.lang.annotation.Target;
  * status code (such as
  * {@link org.eclipse.microprofile.lra.client.IllegalLRAStateException}
  * but any exception that maps to 412 will do).
+ *
+ * The id of the currently running LRA can be obtained by inspecting the
+ * incoming JAX-RS headers.
+ *
+ * Since the participant generally needs to know the id of the LRA in order
+ * to report its status there is generally no benefit to combining this
+ * annotation with the `@LRA` annotation (though it is not prohibited).
  */
 @Inherited
 @Retention(RetentionPolicy.RUNTIME)
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
@@ -64,7 +64,8 @@ import java.time.temporal.ChronoUnit;
  *
  * <p>
  * If an LRA is propagated to a resource that is not annotated with any
- * particular LRA behaviour then the LRA will be suspended. But if this resource
+ * particular LRA behaviour then the LRA will be suspended (ie the context
+ * will not be available to the resource). But if this resource
  * then performs an outgoing JAX-RS request then the suspended LRA must be propagated
  * on this second outgoing request. For example, suppose resource <code>A</code> starts an LRA
  * and then performs a JAX-RS request to resource <code>B</code> which does not contain any LRA
@@ -75,8 +76,9 @@ import java.time.temporal.ChronoUnit;
  * </p>
  *
  * <p>
- * Resource methods can access the LRA context id, if required, by injecting it via
- * the JAX-RS {@link HeaderParam} annotation. This may be useful, for example, for
+ * Resource methods can access the LRA context id, if required, by inspecting the request headers
+ * using standard JAX-RS mechanisms, ie `@Context` or by injecting it via the JAX-RS {@link HeaderParam}
+ * annotation with value {@value #LRA_HTTP_HEADER}. This may be useful, for example, for
  * associating business work with an LRA.
  * </p>
  */
@@ -100,7 +102,7 @@ public @interface LRA {
     /**
      * <p>
      * The Type element of the LRA annotation indicates whether a bean method
-     * is to be executed within the context of a LRA.
+     * is to be executed within the context of an LRA.
      *
      * <p>
      * If the method is to run in the context of an LRA and the annotated class
@@ -136,13 +138,13 @@ public @interface LRA {
     enum Type {
         /**
          * <p>
-         * If called outside a LRA context a new LRA will be created for the
+         * If called outside an LRA context a new LRA will be created for the
          * the duration of the method call and when the call completes it will
          * be closed.
          * </p>
          *
          * <p>
-         * If called inside a LRA context the invoked method will run with the
+         * If called inside an LRA context the invoked method will run with the
          * same context and the context will remain active after the method
          * completes.
          * </p>
@@ -151,13 +153,13 @@ public @interface LRA {
 
         /**
          * <p>
-         * If called outside a LRA context a new LRA will be created for the
+         * If called outside an LRA context a new LRA will be created for the
          * the duration of the method call and when the call completes it will
          * be terminated (closed or cancelled).
          * </p>
          *
          * <p>
-         * If called inside a LRA context it will be suspended and a new LRA
+         * If called inside an LRA context it will be suspended and a new LRA
          * context will be created for the duration of the call. When the method
          * finishes this new LRA will be terminated (closed or cancelled) and
          * the original context will be resumed.
@@ -187,19 +189,19 @@ public @interface LRA {
 
         /**
          *  <p>
-         *  If called outside a LRA context the bean method execution
-         *  must then continue outside a LRA context.
+         *  If called outside an LRA context the bean method execution
+         *  must then continue outside an LRA context.
          *  </p>
          *
          *  <p>
-         *  If called inside a LRA context the managed bean method execution
+         *  If called inside an LRA context the managed bean method execution
          *  must then continue inside this LRA context.
          *  </p>
          */
         SUPPORTS,
 
         /**
-         * The bean method is executed without a LRA context. If a context is
+         * The bean method is executed without an LRA context. If a context is
          * present on entry then it is suspended and then resumed after the
          * execution has completed.
          */
@@ -207,12 +209,12 @@ public @interface LRA {
 
         /**
          * <p>
-         * If called outside a LRA context the managed bean method execution
-         * must then continue outside a LRA context.
+         * If called outside an LRA context the managed bean method execution
+         * must then continue outside an LRA context.
          * </p>
          *
          * <p>
-         * If called inside a LRA context the method is not executed and a
+         * If called inside an LRA context the method is not executed and a
          * <code>412 Precondition Failed</code> HTTP status code is returned
          * to the caller.
          * </p>
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/Leave.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/Leave.java
@@ -26,22 +26,42 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
- * A resource may be enlisted with an LRA if one of its resource methods
- * annotated with @LRA is invoked. If the resource also contains a method
- * annotated with @Leave, then invoking such a method in the context of the
- * same LRA that it has joined will result in the resource leaving the LRA
- * (ie it will not receive any callbacks when the LRA is later closed or
- * cancelled). However, leaving an LRA does not prohibit the resource from
- * rejoining the LRA if one of the @LRA methods is invoked again in the
- * context of the same LRA.
+ * <p>
+ * If a resource method is annotated with `@Leave` and is invoked in the context of
+ * an LRA and if the bean class has registered a participant with that LRA then
+ * it will be removed from the LRA just before the bean method is entered.
+ * The participant can forget about this LRA, in particular it will not
+ * be asked to complete or compensate when the LRA is subsequently ended.
+ * Even though the method runs without an LRA context, the implementation
+ * MUST still make the context available via a JAX-RS header and any outgoing
+ * JAX-RS invocations performed by the method will still carry the context that
+ * the participant has just left. Therefore the business logic must be
+ * careful about any JAX-RS invocations it makes in the body of the annotated
+ * method which may result in other resources being enlisted with the LRA.
+ * </p>
  *
- * Note that it is not possible to join or leave an LRA that has already
- * been asked to cancel or close.
+ * <p>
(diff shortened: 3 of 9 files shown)


api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/Leave.java
```diff
@@ -26,22 +26,42 @@
...
- * annotated with @LRA is invoked. If the resource also contains a method
- * annotated with @Leave, then invoking such a method in the context of the
- * same LRA that it has joined will result in the resource leaving the LRA
- * (ie it will not receive any callbacks when the LRA is later closed or
- * cancelled). However, leaving an LRA does not prohibit the resource from
- * rejoining the LRA if one of the @LRA methods is invoked again in the
- * context of the same LRA.
+ * <p>
+ * If a resource method is annotated with `@Leave` and is invoked in the context of
+ * a LRA and if the bean class has registered a participant with that LRA then
```
xstefank:
```suggestion
 * an LRA and if the bean class has registered a participant with that LRA then
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/Leave.java
```diff
@@ -26,22 +26,42 @@
...
+ * it will be removed from the LRA just before the bean method is entered.
+ * The participant can forget about this LRA, in particular it will not
+ * be asked to complete or compensate when the LRA is subsequently ended.
+ * Even though the method runs without a LRA context, the implementation
```
xstefank:
```suggestion
 * Even though the method runs without an LRA context, the implementation
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -304,6 +304,24 @@ that it could not until explicitly told that it can clean up by providing a meth
 
 If multiple methods are annotated with the same annotation an arbitrary one will be chosen.
 
+[[lra-context]]
+==== The LRA Context
+
+When a method in invoked in the context of a LRA its identifier MUST be made
```
```suggestion
When a method in invoked in the context of an LRA its identifier MUST be made
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -304,6 +304,24 @@ that it could not until explicitly told that it can clean up by providing a meth
...
+is made available via request and response headers which the business method
+can obtain using standard JAX-RS mechanisms (ie @Context or by injecting a
+JAX-RS header param with the name specified by the Java constant `LRA_HTTP_HEADER`
+as defined <<source-LRA, in the LRA annotation class>>
+footnote:[issue #35 (Do not define JAX-RS annotations on @Compensate methods)
+removes the requirement for some methods to be JAX-RS resources. The resolution
+of that issue should ensure the context is still available to business logic.
+
+If an LRA is propagated to a resource that is not annotated with any of the
+annotations defined in this specification then the LRA will be not be made
```
```suggestion
annotations defined in this specification then the LRA will not be made
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -683,12 +696,34 @@ automatically cancel the LRA which will result in this method being invoked.
...
 ==== Leaving an LRA
 
-If a resource method annotated with `@Leave` is invoked in the context of
-a LRA and if the bean class has registered a participant with the active LRA
-it will be removed from the LRA just before the bean method is entered
-(and will not be asked to complete or compensate when the LRA is subsequently
-ended). Even though the method runs without an LRA context the implementation
-MUST still make the context available via a JAX-RS header.
+If a resource method is annotated with `@Leave` and is invoked in the context of
+a LRA and if the bean class has registered a participant with that LRA then
```
```suggestion
an LRA and if the bean class has registered a participant with that LRA then
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -683,12 +696,34 @@ automatically cancel the LRA which will result in this method being invoked.
...
+it will be removed from the LRA just before the bean method is entered.
+The participant can forget about this LRA, in particular it will not
+be asked to complete or compensate when the LRA is subsequently ended.
+Even though the method runs without a LRA context, the implementation
```
```suggestion
Even though the method runs without an LRA context, the implementation
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -683,12 +696,34 @@ automatically cancel the LRA which will result in this method being invoked.
...
+will execute with this new context). Note that in this case the context exposed in
+the `LRA_HTTP_HEADER` JAX-RS header will be set to the new LRA (and not the original
+one), ie the orignal context will not be available to the business logic.
+
+Also note that it is not possible to join or leave an LRA that has already
+been asked to cancel or close (since that would conflict with the
+the participant state model as defined in the LRA specification).
+
+Leaving a particular LRA has no effect on any other LRA - ie the same
+resource can be enlisted with many different LRA's and leaving one
```
```suggestion
resource can be enlisted with many different LRAs and leaving one
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -683,12 +696,34 @@ automatically cancel the LRA which will result in this method being invoked.
...
+particular LRA will not affect its participation in any of the other
+LRA's it has joined.
```
xstefank: there is nothing wrong with this but since the specifications says that it should be read together with javadoc, do we need to replicate this `@Leave` documentation both in javadoc and here?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -304,6 +304,24 @@ that it could not until explicitly told that it can clean up by providing a meth
 
 If multiple methods are annotated with the same annotation an arbitrary one will be chosen.
 
+[[lra-context]]
+==== The LRA Context
+
+When a method in invoked in the context of a LRA its identifier MUST be made
+available to the business logic. For JAX-RS resource methods the identifier
+is made available via request and response headers which the business method
+can obtain using standard JAX-RS mechanisms (ie @Context or by injecting a
```
ochaloup:
```suggestion
can obtain using standard JAX-RS mechanisms (ie `@Context` or by injecting a
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -304,6 +304,24 @@ that it could not until explicitly told that it can clean up by providing a meth
...
+JAX-RS header param with the name specified by the Java constant `LRA_HTTP_HEADER`
+as defined <<source-LRA, in the LRA annotation class>>
+footnote:[issue #35 (Do not define JAX-RS annotations on @Compensate methods)
```
ochaloup:
```suggestion
footnote:[issue #35 (Do not define JAX-RS annotations on `@Compensate` methods)
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -75,9 +76,10 @@
  * </p>
  *
  * <p>
- * Resource methods can access the LRA context id, if required, by injecting it via
- * the JAX-RS {@link HeaderParam} annotation. This may be useful, for example, for
- * associating business work with an LRA.
+ * Resource methods can access the LRA context id, if required, by inspecting the request headers
+ * using standard JAX-RS mechanisms, ie @Context or by injecting it via the JAX-RS {@link HeaderParam}
```
ochaloup:
```suggestion
 * using standard JAX-RS mechanisms, ie `@Context` or by injecting it via the JAX-RS {@link HeaderParam}
```

mmusgrov:
Check that this PR is clear about the expected behaviour when the user sets the context in request and response headers.

rdebusscher:
Spec and Javadoc are clear now around @Leave and @LRA interaction.
Thx @mmusgrov !

mmusgrov:
@rdebusscher and @xstefank I added more text about the expected behaviour when the user sets the context in request and response headers. Can you re-verify that I still have the approval to merge.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -322,6 +326,35 @@ annotations defined in this specification then the LRA will not be made
...
+===== Setting the context on outgoing JAX-RS requests
+
+If the context is set on an outgoing JAX-RS request then it becomes the `active
+context` for that invocation. When the invocation returns the `active context`
+will depend on the behaviour of the remote resource:
+
+- if the remote resouce ends the context (and does not start a new one) then
+  the `active context` will be whatever the client was using before it manually set
+  the context on the outgoing header;
+- if the remote resouce does not end the context (and does not start a new one)
```
rdebusscher: In the following scenario

@LRA
Response doSomething() {
  // here we have LRAid = 123
  client.  ... .header(LRA.LRA_HTTP_HEADER, "456").get() ;   // We set LRAid 456 to outgoing call (I know they are URI, just for simplicity)
 // Does this means the 'Active Context' is now LRAid = 456 and no longer 123?
  client.  ... .get();  .// Without specifying the header, does this mean we are using LRAid 456?

-> If so, this will be confusing. We specify the header for a certain REST call but it should not affect the 'Current Context' of method doSomething().
And also, how will the system know that the called endpoint ended the LRA (as mentioned in previous bullet) because it should not alter the 'current Context' of method doSomething()? 
Is the LRA_HTTP_HEADER only added to responses when the LRA is still active (end = false)?

mmusgrov: The code in the doSomething() method has explicitly set the context on the outgoing request so it is managing the context manually - ie the business developer understands the resulting behaviour since he has read this bit of the spec. Now if the service invoked by the doSomething() method does not end the LRA and if the code in the doSomething() method wants the next request to use the original context then he just needs to explicitly set it on any other outgoing invocations made inside the doSomething() method.

Note that I do say in the following paragraph that

> The user MAY manage the context manually by tracking in progress LRAs (via the context header) and ensure he manually sets the desired context on outgoing JAX-RS invocations.

To your second point, I think the spec is clear (but we can clarify if not) that if the LRA annotation does not automatically end an LRA it becomes the active one for the caller (otherwise the LRA id would be lost/unavailable to the caller).

To your final point, yes, the LRA_HTTP_HEADER is only added to responses when the LRA is still active (end = false). This is because the `context header` is reserved for the `active context`.

mmusgrov: I think the problem is that I am wanting to support "implicit context propagation" where the spec implementation manages the association. We currently have both explicit (where the user is allowed to manage the context) and implicit. I acknowledge that supporting both can be a challenge to specify in a clear and unambiguous way. Manage thread association can also be tricky in reactive environments.

So would people be more comfortable if we say that the user has to manage the context on outgoing requests (I think Martin asked for this a while back) and the container manages the response header according to what is dictated by the @LRA annotation.

If so then I will update this PR to remove the requirement for "implicit context propagation". Thoughts?

Leave:
Thanks, I think this will help. 
#include<iostream.h> #include<conio.h> #include<fstream.h> #include<stdio.h> #include<stdlib.h> void main() { clrscr(); ifstream fs; ofstream ft; char ch, fname1[20], fname2[20]; cout<<"Enter source file name with extension (like files.txt) : "; 	gets(fname1); fs.open(fname1); if(!fs) 	{ cout<<"Error in opening source file..!!"; getch(); exit(1); } cout<<"Enter target file name with extension (like filet.txt) : "; gets(fname2); ft.open(fname2); if(!ft) 	{ cout<<"Error in opening target file..!!"; fs.close(); getch(); exit(2); } while(fs.eof()==0) { fs>>ch; ft<<ch; } cout<<"File copied successfully..!!"; fs.close(); ft.close(); getch();

spec/src/main/asciidoc/microprofile-lra-spec.adoc
rdebusscher: Implicit context propagation is good in my opinion and probably 99% of the cases.
But when the developer sets the context, it should only be for that single request which doesn't affect the current context.

I think this can easily be achieved by not looking at the value on the Response. (Which also covers then the case where the called endpoint starts a new; completely separate LRA)

We can discus this on next meeting

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/Leave.java
```diff
@@ -26,22 +26,42 @@
...
+ * If a resource method is annotated with `@Leave` and is invoked in the context of
+ * an LRA and if the bean class has registered a participant with that LRA then
+ * it will be removed from the LRA just before the bean method is entered.
+ * The participant can forget about this LRA, in particular it will not
+ * be asked to complete or compensate when the LRA is subsequently ended.
+ * Even though the method runs without an LRA context, the implementation
+ * MUST still make the context available via a JAX-RS header and any outgoing
+ * JAX-RS invocations performed by the method will still carry the context that
+ * the participant has just left. Therefore the business logic must be
+ * careful about any JAX-RS invocations it makes in the body of the the annotated
```
xstefank:
```suggestion
 * careful about any JAX-RS invocations it makes in the body of the annotated
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -304,6 +304,57 @@ that it could not until explicitly told that it can clean up by providing a meth
 
 If multiple methods are annotated with the same annotation an arbitrary one will be chosen.
 
+[[lra-context]]
+==== The LRA Context
+
+When a method in invoked in the context of an LRA its identifier MUST be made
```
```suggestion
When a method is invoked in the context of an LRA its identifier MUST be made
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
```diff
@@ -53,6 +53,13 @@
  * status code (such as
  * {@link org.eclipse.microprofile.lra.client.IllegalLRAStateException}
  * but any exception that maps to 412 will do).
+ *
+ * The id of the currently running LRA can be obtained by inspecting the
+ * incoming JAX-RS headers.
+ *
+ * Since the participant generally needs to know the id of the LRA in order
+ * to report its' status there is generally no benefit to combining this
```
xstefank:
```suggestion
 * to report its status there is generally no benefit to combining this
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -683,12 +735,15 @@ automatically cancel the LRA which will result in this method being invoked.
 [[leaving-an-lra]]
 ==== Leaving an LRA
 
-If a resource method annotated with `@Leave` is invoked in the context of
-a LRA and if the bean class has registered a participant with the active LRA
+If a resource method annotated with @Leave is invoked in the context of a
```
```suggestion
If a resource method annotated with @Leave is invoked in the context of an
```
since it is already changed everywhere else :+1:

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: +1 (we also got +1's from everyone present at today's hangout).
@rdebusscher Will you check that my last commit (https://github.com/eclipse/microprofile-lra/pull/124/commits/4faab7d2f84f135bd19896a8117c7de730ceb02f) matches your expectation. Thanks

Issue #120, opened 2019-03-19
Add TCK tests verifying functionality of different LRA.Type 


ochaloup:
The `LRA.Type` (https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L123) defines the LRA scope the invocation belongs to. The TCK should contain tests which verify whether all the `Type`s are correctly handled.

- [ ] `Type.REQUIRED` : if LRA context exists it joins it
- [ ] `Type.REQUIRED` : if LRA context does not exist it creates new LRA
- [ ] `Type.REQUIRES_NEW` : if LRA context exists it creates new LRA
- [ ] `Type.REQUIRES_NEW` : if LRA context does not exist it creates new LRA
- [ ] `Type.MANDATORY` : if LRA context exists it joins LRA
- [ ] `Type.MANDATORY` : if LRA context does not exist it fails with `412, precondition failed`
- [ ] `Type.SUPPORTS` : if LRA context exists it joins LRA
- [ ] `Type.SUPPORTS` : if LRA context does not exist it does not join LRA (running outside scope of LRA)
- [ ] `Type.NOT_SUPPORTED` : if LRA context exists the method is executed outside of the scope
- [ ] `Type.NOT_SUPPORTED` : if LRA context does not exist the method is executed outside of the scope
- [ ] `Type.NOT_SUPPORTED` : if LRA context exists the context is suspended and resumed(!) at the end of the annotated method
- [ ] `Type.NEVER` : if LRA context exists it fails with `412, precondition failed`
- [ ] `Type.NEVER` : if LRA context does not exist it executes outside of the LRA context

xstefank:
@ochaloup +1 great idea. One small request: each bullet point here should really be mostly two test cases in my opinion

ochaloup:
@xstefank thanks, I changed the description

mmusgrov:
+1

PR #128, opened 2019-03-29
issue120 test the behaviour of the LRA#type annotation

mmusgrov:
#120

Adds tests that validate that the implementation of the {@link LRA.Type} attribute is correct:
For each of the type values there are 4 tests:
 * start an LRA before invoking the resource method annotated with {@link LRA#end()} set to true
 * do not start an LRA before invoking the resource method annotated with {@link LRA#end()} set to true
 * start an LRA before invoking the resource method annotated with {@link LRA#end()} set to false
 * do not start an LRA before invoking the resource method annotated with {@link LRA#end()} set to false
 
Each test then validates {@link TckLRATypeTests#resourceRequest} that:
 * the correct JAX-RS status was returned
 * the resource method executed with the correct LRA context
 * the LRA started by the resource method is closed if {@link LRA#end()} is true
 * the LRA started by the resource method is not closed if {@link LRA#end()} is false

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java
@@ -30,6 +30,14 @@ import javax.ws.rs.core.Response;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.time.temporal.ChronoUnit;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Logger;
 
 import static junit.framework.TestCase.fail;
 import static org.eclipse.microprofile.lra.tck.participant.api.NonParticipatingTckResource.END_PATH;
@@ -41,8 +49,11 @@ import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckR
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.TCK_PARTICIPANT_RESOURCE_PATH;
 
 public class LRAClientOps {
+    private static final Logger LOGGER = Logger.getLogger(TckLRATypeTests.class.getName());
+
     /**
-     * TODO recovery needs to be properly speced
+     * TODO recovery needs to be properly speced: see the proposal
+     * on issue https://github.com/eclipse/microprofile-lra/issues/116
      * It is used in the two tests for participants that return 202:
      * {@link TckTests#acceptCancelTest()}  and {@link TckTests#acceptCloseTest()}
      * to connect to an endpoint that will trigger recovery (the alternative would
@@ -68,9 +79,13 @@ public class LRAClientOps {
     static final String LRA_RECOVERY_PATH_KEY = "lra.http.recovery.path";
 
     private final WebTarget target;
+    private final ScheduledExecutorService executor;
+    private final Map<LRATask, ScheduledFuture<?>> lraTasks;
 
     public LRAClientOps(WebTarget target) {
         this.target = target;
+        this.executor = Executors.newSingleThreadScheduledExecutor();
+        this.lraTasks = new HashMap<>();
     }
 
     // see if it is possible to join with an LRA - if it is possible to do that then the LRA is still active
@@ -105,7 +120,8 @@ public class LRAClientOps {
         return invokeRestEndpoint(lra == null ? null : lra.toASCIIString(), basePath, path, coerceResponse);
     }
 
-    private Response invokeRestEndpoint(String lra, String basePath, String path, int coerceResponse) {
+    // synchronize access to the connection since it is shared with the LRA background cancellation code
+    private synchronized Response invokeRestEndpoint(String lra, String basePath, String path, int coerceResponse) {
         WebTarget resourcePath = target.path(basePath).path(path).queryParam(STATUS_CODE_QUERY_NAME, coerceResponse);
         Invocation.Builder builder = resourcePath.request();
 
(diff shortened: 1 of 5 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -0,0 +1,349 @@
...
+        tckSuiteClient = ClientBuilder.newClient();
+    }
+
+    // enum to indicate which checks to perform on the expected and actual active LRA after running a resource method
+    private enum MethodLRACheck {
+        NONE, NOT_PRESENT, EQUALS, NOT_EQUALS
+    }
+
+    @Test
+    public void requiredWithLRATest() {
```
ochaloup: just a not to consider. When all those are tests not be more readable to name them just `requiredWithLRA`, `requiredWithoutLRA`

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -0,0 +1,349 @@
...
+            } else if (methodLRA.length() != 0) {
+                // otherwise it should be finished
+                assertTrue(lraClient.isLRAFinished(methodLRA));
+            }
+
+            if (lra != null) {
+                lraClient.closeLRA(lra);
+            }
+        } finally {
+            response.close();
```
ochaloup: could not be the `response` `null` in some bad circumstances

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
mmusgrov: +1

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
mmusgrov: The response object is initialised on line 285 and it is outside of the try/finally block so if that failed then an exception would be thrown and the thread would never have entered the try/finally block, i.e. it cannot be null:

> Response response = lra == null ? target.get() : target.header(LRA.LRA_HTTP_HEADER, lra).get();
> try {
> ...

tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java
```diff
@@ -163,6 +163,11 @@ void closeLRA(URI lraId) throws GenericLRAException {
         invokeRestEndpointAndReturnLRA(lraId.toASCIIString(), TCK_NON_PARTICIPANT_RESOURCE_PATH, END_PATH, 200);
     }
 
+
+    public void closeLRA(String lraId) {
```
rdebusscher: All other methods are package scoped. Here public can also be removed.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -0,0 +1,347 @@
...
+    @Inject
+    private LraTckConfigBean config;
+
+    private LRAClientOps lraClient;
+
+    private static Client tckSuiteClient;
+
+    private WebTarget tckSuiteTarget;
+
+    @Deployment(name = "lra-type-tck-tests", managed = true, testable = true)
```
rdebusscher: minor: managed and testable are true by default.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -0,0 +1,347 @@
...
+     * @param methodLRAShouldBeActive if true the LRA started by the invoked method should still
+     *                               be active after the resource invocation completes
+     */
+    private void resourceRequest(String path, boolean startLRA, int expectedStatus, MethodLRACheck lraCheckType,
+                                 boolean methodLRAShouldBeActive) {
+        Invocation.Builder target = tckSuiteTarget.path(TCK_LRA_TYPE_RESOURCE_PATH)
+                .path(path).request();
+        URI lra = startLRA ? lraClient.startLRA(null, lraClientId(), lraTimeout(), ChronoUnit.MILLIS) : null;
+
+        Response response = lra == null ? target.get() : target.header(LRA.LRA_HTTP_HEADER, lra).get();
```
rdebusscher: This is better readable

        if (lra != null) {
            target = target.header(LRA.LRA_HTTP_HEADER, lra);
        }
        Response response = target.get();

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -0,0 +1,347 @@
...
+     * @param lraCheckType the type of check to perform on the request context and the context specified
+     *                     by the {@link LRA.Type} annotation
+     * @param methodLRAShouldBeActive if true the LRA started by the invoked method should still
+     *                               be active after the resource invocation completes
+     */
+    private void resourceRequest(String path, boolean startLRA, int expectedStatus, MethodLRACheck lraCheckType,
+                                 boolean methodLRAShouldBeActive) {
+        Invocation.Builder target = tckSuiteTarget.path(TCK_LRA_TYPE_RESOURCE_PATH)
+                .path(path).request();
+        URI lra = startLRA ? lraClient.startLRA(null, lraClientId(), lraTimeout(), ChronoUnit.MILLIS) : null;
```
rdebusscher: .startLRA() doesn't use the client Id and Timeout parameters. Probably needs cleanup.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -0,0 +1,347 @@
...
+    @Test
+    public void neverWithoutLRATest() {
+        resourceRequest(NEVER_PATH, false, 200, MethodLRACheck.NOT_PRESENT, false);
+    }
+
+    // same set of tests except that the invoked method end = true set on the LRA annotation
+
+    @Test
+    public void requiredEndWithLRATest() {
+        resourceRequest(REQUIRED_WITH_END_PATH, true, 200, MethodLRACheck.EQUALS, true);
```
rdebusscher: _WITH_END is confusing here. Because it is an endpoint were end=false is defined.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -0,0 +1,347 @@
...
+    public void neverWithRATest() {
+        resourceRequest(NEVER_PATH, true, 412, MethodLRACheck.NOT_PRESENT, false);
+    }
+
+    @Test
+    public void neverWithoutLRATest() {
+        resourceRequest(NEVER_PATH, false, 200, MethodLRACheck.NOT_PRESENT, false);
+    }
+
+    // same set of tests except that the invoked method end = true set on the LRA annotation
```
rdebusscher: the next set of tests have end=false!

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
mmusgrov: It seems overkill to schedule a timer to cancel the LRA and to cancel the timer when the LRA is closed. Personally I think we should run all LRAs with an indefinite timeout. If a test hangs because of that then the test is brittle and needs rewriting. Do you still want me to manage timers and their cancellation if the LRAs finish early?

mmusgrov: Well I implemented it anyway (took over 100 lines of code diffs, lol). I didn't squash the diff so that it is clear which bit of code is implementing the timeout behaviour.

mmusgrov:
@rdebusscher I addressed all of your comments and the other reviewer has approved so I went ahead and merged.

Issue #129, opened 2019-04-02
Specification clarification for @LRA(CancelOn, ...) behaviour


mmusgrov:
Add a note to say LRA#CancelOn is ignored if the transaction is already closed.

In addition, check other annotation text for completeness.

PR #136, opened 2019-04-10
Issue129 Clarify interaction between LRA#end and LRA#cancelOn

mmusgrov:
#129 
javadoc update only

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
@@ -24,6 +24,7 @@ import javax.ws.rs.HeaderParam;
 import javax.ws.rs.core.Response;
 
 import org.eclipse.microprofile.lra.annotation.Compensate;
+import org.eclipse.microprofile.lra.annotation.Complete;
 
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Inherited;
@@ -35,19 +36,23 @@ import java.time.temporal.ChronoUnit;
 /**
  * <p>
  * An annotation for controlling the lifecycle of Long Running Actions (LRAs).
+ * </p>
  *
  * <p>
  * The annotation <b>SHOULD</b> be applied to JAX-RS annotated methods otherwise
- * it <b>MAY</b> have no effect. The annotation determines whether or not the annotated
- * method will run in the context of an LRA and controls whether or not:
+ * it <b>MAY</b> have no effect. The annotation determines whether or not the
+ * annotated method will run in the context of an LRA and controls whether or not:
  * </p>
  *
  * <ul>
- *   <li>any incoming context should be suspended and if so if a new one should be started</li>
+ *   <li>any incoming context should be suspended and if so if a new one should be
+ *       started</li>
  *   <li>to start a new LRA</li>
  *   <li>to end any LRA context when the method ends</li>
- *   <li>to throw an exception if there should be an LRA context present on method entry</li>
- *   <li>to cancel LRA context when the method returns particular HTTP status codes</li>
+ *   <li>to return a error status code without running the annotated method if
+ *       there should have been an LRA context present on method entry</li>
+ *   <li>to cancel the LRA context when the method returns particular HTTP
+ *       status codes</li>
  * </ul>
  *
  * <p>
@@ -56,10 +61,11 @@ import java.time.temporal.ChronoUnit;
  * called {@value #LRA_HTTP_CONTEXT_HEADER}.
  * The implementation (of the LRA specification) is expected to manage this context
  * and the application developer is expected to declaratively control the creation,
- * propagation and destruction of LRAs using the {@link LRA} annotation. When a JAX-RS bean
- * method is invoked in the context of an LRA any JAX-RS client requests that it
- * performs will carry the same header so that the receiving resource knows that it
- * is inside an LRA context.
+ * propagation and destruction of LRAs using this {@link LRA} annotation. When a
+ * JAX-RS resource method is invoked in the context of an LRA, any JAX-RS client
+ * requests that it performs will carry the same header so that the receiving
+ * resource knows that it is inside an LRA context. The behaviour may be overridden
+ * by manually setting the context header.
  * </p>
  *
  * <p>
@@ -67,19 +73,20 @@ import java.time.temporal.ChronoUnit;
  * particular LRA behaviour then the LRA will be suspended (ie the context
  * will not be available to the resource). But if this resource
  * then performs an outgoing JAX-RS request then the suspended LRA must be propagated
- * on this second outgoing request. For example, suppose resource <code>A</code> starts an LRA
- * and then performs a JAX-RS request to resource <code>B</code> which does not contain any LRA
- * annotations. If resource <code>B</code> then performs a JAX-RS request to a third service, <code>C</code> say,
- * which does contain LRA annotations then the LRA context started at <code>A</code> must be propagated
+ * on this second outgoing request. For example, suppose resource <code>A</code>
+ * starts an LRA and then performs a JAX-RS request to resource <code>B</code> which
+ * does not contain any LRA annotations. If resource <code>B</code> then performs a
+ * JAX-RS request to a third service, <code>C</code> say, which does contain LRA
+ * annotations then the LRA context started at <code>A</code> must be propagated
  * to <code>C</code> (for example if <code>C</code> uses an annotation to join the LRA,
  * then <code>C</code> must be enlisted in the LRA that was started at <code>A</code>).
  * </p>
  *
  * <p>
- * Resource methods can access the LRA context id, if required, by inspecting the request headers
- * using standard JAX-RS mechanisms, ie `@Context` or by injecting it via the JAX-RS {@link HeaderParam}
- * annotation with value {@value #LRA_HTTP_CONTEXT_HEADER}. This may be useful, for example, for
- * associating business work with an LRA.
+ * Resource methods can access the LRA context id by inspecting the request headers
+ * using standard JAX-RS mechanisms, ie `@Context` or by injecting it via the JAX-RS
+ * {@link HeaderParam} annotation with value {@value #LRA_HTTP_CONTEXT_HEADER}.
+ * This may be useful, for example, for associating business work with an LRA.
  * </p>
  */
 @Inherited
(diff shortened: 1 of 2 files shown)


mmusgrov:
@ochaloup I changed some of the cancelOn and cancelOnFamily javadoc text in this PR

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -241,38 +241,48 @@
...
      *
      * <p>
-     * If the <code>end</code> value is set to <code>false</code>
-     * while the annotated method returns the {@link Response} status code
-     * with the meaning the LRA to be cancelled (see {@link #cancelOn()} and {@link #cancelOnFamily()})
-     * then this error state has precedence over the <code>end</code> attribute.
-     * It means the LRA will be cancelled despite the end was set to <code>false</code>.
+     * If the <code>end</code> value is set to {@literal false}
+     * but the annotated method finishes with a status code that matches
+     * one of the {@link #cancelOn()} and {@link #cancelOnFamily()} values then
```
xstefank:
```suggestion
     * one of the {@link #cancelOn()} or {@link #cancelOnFamily()} values then
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -241,38 +241,48 @@
...
+     * the LRA will be cancelled instead of closed. In other words the
```
xstefank:
```suggestion
     * the LRA will be cancelled. In other words the
```
this gives wrong perception that even when end = false then the LRA is by default expected to be closed.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: A pointless comment. But I'd prefer to stay with my text.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: Not sure what your point is but I'd prefer to use my text.

mmusgrov:
@ochaloup An extra test is a good idea, I will add one.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
xstefank: for me it is either one of cancelOn or cancelOnFamily, "and" just doesn't fit here, but ok

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
xstefank: reading from the beginning of the paragraph, we have `@LRA(end = false)` so it should not end when the method is finished. If status code matching cancel* attributes is returned then LRA will be cancelled. But that doesn't mean it will be cancelled instead of closed because it should not be closed when the method is finished because of the end=false.

But I understand that it should be by default ended in some other method later. This is just not clear from the sentence and it may be confusing for someone reading the spec top bottom. Keep it as you wish.

mmusgrov: You are correct, +1

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: Try that last commit (it rewords the text and also uses the word `or`)

mmusgrov:
@ochaloup I took a look at the exception types that we define in the API and I am of the opinion that they do not add value and should be removed from the spec. I will raise a separate issue for that. If we do not end up removing the exception types then the extra tests that you call for can be discussed/introduced there.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -101,37 +103,35 @@
...
-     * The Type element of the LRA annotation indicates whether a bean method
+     * The Type element of the LRA annotation indicates whether a resource method
      * is to be executed within the context of an LRA.
      *
      * <p>
      * If the method is to run in the context of an LRA and the annotated class
-     * also contains methods annotated with {@link Compensate}/{@link Compensate}
-     * then the bean will enlisted with the LRA. Enlisting with an LRA means that
-     * the bean will be notified when the current LRA is later cancelled/closed.
+     * also contains a method annotated with {@link Compensate}
```
rdebusscher: The resource is also enlisted when only a @Complete annotated method is present.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -234,45 +234,55 @@
...
 
     /**
      * <p>
-     * Normally if an LRA is present when a bean method is executed it will not
-     * be ended when the method returns. To override this behaviour and force LRA
-     * termination on exit use the end element.
+     * Normally if an LRA is present when a resource method is executed it will not
+     * be ended when the method returns. To override this behaviour and force the LRA
+     * to close on exit use this <code>end</code> element with the value set to
+     * {@literal true}.
```
rdebusscher: The default value is 'true'. So normally (without explicitly changing the value to false) LRA is ended when the method returns.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -234,45 +234,55 @@
...
     boolean end() default true;
 
     /**
      * <p>
      * The cancelOnFamily element can be set to indicate which families of
-     * HTTP response codes will cause the current LRA to cancel.
-     * This expects the annotated method is the JAX-RS endpoint where the {@link Response}
-     * with the status code could be returned.
+     * HTTP response codes will cause the current LRA to cancel. If the LRA
+     * has already been closed when the annotated method returns then this
```
rdebusscher: In which circumstances is it possible that the LRA is already closed?

Are there any other that the following one?

Method A calls Method B (with B having type=Required for example, not starting a new LRA)
Method B has end=true
So when Service A ends, LRA is already closed.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -234,45 +234,55 @@
...
-     * By default client errors (<code>4xx codes</code>) and server errors (<code>5xx codes</code>)
+     * If a JAX-RS method is annotated with this element and the method
+     * returns a response code which matches any of the specified families
+     * then the LRA will be cancelled. The method can return status codes
+     * in a {@link Response} or via an exception mappper.
+     * </p>
+     * 
+     * <p>
+     * By default client errors {@link Response.Status.Family#CLIENT_ERROR}
+     * and server errors {@link Response.Status.Family#SERVER_ERROR}
```
rdebusscher: The **default value** of _cancelOnFamily_ is CLIENT_ERROR and SERVER_ERROR.

I think we should not just  use "By default" as this can cause confusion.

So when the developer set the element value to an empty list, there is no HTTP status anymore which triggers the cancellation of the LRA. So there is no "default behind the scenes" which always kicks in.

mmusgrov: Okay thanks. I deleted the paragraph and will now merge the PR

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: No it isn't, @Compensate is mandatory.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: okay

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: Marking method B with any of MANDATORY, REQUIRED and SUPPORTS with end=true or if method B calls other methods similarly annotated then the LRA will be closed. The Java API would have been another mechanism for causing it to be closed.

The text I have written is a catch all and listing each possible scenario is risky since we may omit one. But if you want me to include examples then let me know.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: Personally I think that both `@Compensate` and `@Complete` should be mandatory - they used to be but one of the Camel team argued that Complete should be optional. When I accepted that change request I added the example text in the above quote.

tomjenkinson: I guess it was @nicolaferraro with whom this discussion took place? If the idea is to change this to make them both mandatory we should try to get some feedback from wider community

nicolaferraro: Yes, I remember I mentioned I've found some use cases where you don't want to be notified when the LRA is @Complete. For example it makes a lot of sense in **passive** services.

E.g. I've a system that just tracks hotel reservations and has operations for booking a room or cancelling the reservation (@Compensate). Since this system is passive, once a room is booked, it does not make any difference if the LRA is completed or not: the room will be unavailable for others. If it receives a call to "compensate", it will free the room. But it won't do anything on @Complete.

mmusgrov: @nicolaferraro Do you mind if I use that example?

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: @rdebusscher I have just pushed a commit that makes sure the documentation around `end` is accurate.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: @nicolaferraro I included the example but if that was not okay then I will back out the commit.

nicolaferraro: That's great @mmusgrov , thanks. It helps clarifying this point.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
rdebusscher: @mmusgrov Thanks for confirmation. I wasn't sure if there was any other possibility.  No need to add examples. Thx

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
rdebusscher: @mmusgrov Ok. Probably a rare case that you only want to perform some action on @Complete  but not on @Compensate.

Issue #137, opened 2019-04-11
Remove package org.eclipse.microprofile.lra.client from the API


mmusgrov:
The org.eclipse.microprofile.lra.client package should have been removed when the Java Client API was removed in issue #95 and this new issue will remove it. The only things left in the client API package are the exception types (GenericLRAException, IllegalLRAStateException, InvalidLRAIdException and UnknowableException) which are no longer required.

xstefank:
+1

mmusgrov:
PR merged

PR #138, opened 2019-04-12
issue137 Remove package org.eclipse.microprofile.lra.client from the API

mmusgrov:
#137


api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
```diff
@@ -46,14 +46,6 @@
...
  * participant to free up resources.
  *
- * If this status method is invoked before either of the <em>@Compensate</em>
- * or <em>@Complete</em> methods have been called then the participant
- * should report that it does not yet have a status using a JAX-RS
- * exception mapper that maps to a <em>412 Precondition Failed</em> HTTP
- * status code (such as
- * {@link org.eclipse.microprofile.lra.client.IllegalLRAStateException}
- * but any exception that maps to 412 will do).
- *
```
xstefank: I am approving PR but I just wanted to ask - is this covered somewhere else in the spec or do we leave this behavior unspecified?

mmusgrov: @xstefank Earlier in the project the state model for a participant was only entered via the `Compensating` or the `Completing` states. Later on we added a new state called `Active` which is entered when the participant joins the LRA. Though not strictly necessary, the new state had the advantage of simplying the spec and is more natural for the business code to return a status of Active instead of the exception together with a mapper that maps to 412 and this text should have been removed back then. One could argue that the extra state adds complexity but the benefits probably outweigh the drawbacks.

So to answer your question, the behaviour should be and is fully specified.

Issue #117, opened 2019-03-18
Clarify spec around @NestedLRA when no parent LRA present.


rdebusscher:
See also https://github.com/eclipse/microprofile-sandbox/issues/46.

Within the spec it isn't defined what happens in case a `@NestedLRA`annotated JAX-RS endpoint is called when no parent LRA is present.

mmusgrov:
The javadoc for @NestedLRA does already state this behaviour:

{quote}
 * Used on the interface or class. Defines that the container will create
 * a new LRA for each method invocation, regardless of whether or not there is
 * already an LRA associated with the caller. These LRAs will then
 * either be top-level LRAs or nested automatically depending upon the
 * context within which they are created.
{quote}

and the spec doc states:

{quote}
The definitive documentation for how the annotations affect the behaviour of a running program is
the javadoc and this specification should be read in conjunction with it. The annotations are
defined in the org.eclipse.microprofile.lra.annotation package:
...
{quote}

and there is a brief reference to this behaviour in section  3.2.4. Nesting LRAs of the spec doc:

{quote}
But if there is no LRA present then a new top level LRA is started ...
{quote}

mmusgrov:
I close issue #123 since it is a duplicate of this one. To the second point raised by @nmcl regarding "what happens if an LRA isn't active when @NestedLRA is called": the answer is that a new top level LRA will be started which is documented in the javadoc and in the spec doc:

https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/NestedLRA.java#L46

and in section 3.25 of the spec:

https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#324-nesting-lras

ochaloup:
As we talked at the meeting I read through the [current javadoc of @NestedLRA](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/NestedLRA.java#L46) and I think there is no functional issue with adding the nested lra as the type.

And I like it from the user perspective. I think it makes sense.

I would only add that maybe there will be need to add some more information to say how the other `@LRA` attributes interact with the `NESTED` type. Like adding info that the `cancelOnFamily` cancels the nested LRA but not the parent one or if `end = false` make sense.
```
@LRA(type = NESTED, end = false)
```

PR #140, opened 2019-04-23
issue117 Replace @NestedLRA with an LRA.Type.Nested element

mmusgrov:
#117 

This PR replaces the @NestedLRA annotation with a new value of the @LRA.Type element


mmusgrov:
The last commit addresses [comment](https://github.com/eclipse/microprofile-lra/issues/117#issuecomment-476142383)  made by  @ochaloup

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -215,12 +215,59 @@
...
          * </p>
          */
-        NEVER
+        NEVER,
+
+        /**
+         * <p>
+         *     An activity can be scoped within an existing LRA using the Nested
+         *     element value. A new LRA will be created even if there is already
+         *     once present when the method is invoked, i.e. these LRAs will then
```
rdebusscher: typo : once -> one

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -215,12 +215,59 @@
...
+         *     enclosing LRA is closed or cancelled:
+         * </p>
+         *
+         * <ul>
+         *     <li>If the nested LRA is closed but the outer LRA is cancelled then
+         *     the participants registered with the nested LRA will be told to compensate.
+         *     </li>
+         *     <li>If the nested LRA is cancelled the outer LRA can be still closed
+         *     (and in this case the nested participants will not be asked to complete).
+         *     </li>
```
rdebusscher: We should mention explicitly that the participant will gets called twice, once for the nested LRA, and once for the enclosing LRA. (only when called with an enclosing LRA of course, not when top level)

mmusgrov: But the parent and nested LRA are two different LRAs so the participant will be called with a different context.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: ok

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
rdebusscher: exactly. But JAX-RS resource is unaware of the value of the enclosing LRA id (or should we add it as another header to the method annotated with @NestedLRA)
Otherwise the @Complete and @Compensate methods are called with a LRAid which is unknown to the JAX-RS resource (and methods) and they don't know what to do with it.

mmusgrov:
@rdebusscher 
> What happens in the case the enclosing LRA does not have a 'enlisted participant' (because there is no @compensate)?
> 
> Can nestedLRAs be nested again? (LRA -> NestedLRA -> NestedLRA)
> 
> when LRA -> Nested LRA -> LRA(Required) I assume the last LRA joins the 'Nested LRA'. Since I assume, it is best we explicitly describe it.

A nested LRA is treated just like any other LRA with respect to participant enlistment. When an invocation results in the creation of a nested LRA that LRA becomes the "current context" and any further operations performed by the method will be executed with that context. The semantics of nested LRAs follows previous transactions models:                                                                                  

# a nested LRA can close or cancel independently of its parent
# a nested LRA which has closed must retain the ability to cancel the effects if the the parent cancels                                                                                                         
#. if a nested LRA aborts all of its children must abort (even if they closed - see 2)
#. if a nested LRA closes all of its children must close (but retain the ability to later compensate - see 2)                                                                                                   

Downstream LRAs will only be part of this nesting hierarchy if the downstream methods carry the NESTED element, otherwise they are independent of the current nested LRA.                                    

There is one problem I have overlooked: in order for a service to support point 2 above it needs to be told when it is safe to forget about the LRA. To support this requirement I propose that the spec should say that the implementation MUST call forget on any participants registered with the nested LRA when the parent closes. And an optimisation (though not strictly necessary) could be to tell the participant that it is being called in the context of a nested LRA.

mmusgrov:
@xstefank That would only be an optimisation (though not strictly necessary) that we could add by telling the participant that it is being called in the context of a nested LRA by introducing a JAX-RS header:

```java
String LRA_HTTP_PARENT_CONTEXT_HEADER = "Long-Running-Action-Parent";
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: If the enclosing LRA was created elsewhere and the current resource method is marked NESTED then the resource will only be enlisted with the NESTED LRA, hence there is no ambiguity.

But I agree that adding another header to identify the parent LRA is needed in order to support the nesting semantics I have outlined in a later comment. If I add that I can add text that will explain better how the callbacks work for nesting.

mmusgrov:
@rdebusscher and @xstefank I have pushed an updated PR that addresses each of your comments, can you both take another look.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: So I added the extra explanatory test you asked for and I also added the extra JAX-RS header.

mmusgrov:
@rdebusscher and @xstefank I added a paragraph explaining why cancelling a closed nested LRA makes sense and why closing a cancelled nested LRA does not make sense:
```
         * <p>
         *     The reason why the model does not allow a cancelled nested LRA to be closed
         *     is because the business activity has already been compensated for which means
         *     there is no longer any outstanding work in need of completion.
         * </p>
         * <p>
         *     On the other hand it does make sense to cancel a closed nested LRA since
         *     the work has been done so there is something that can be compensated for.
         * </p>
```

xstefank:
@mmusgrov +1 thanks for that

Issue #139, opened 2019-04-18
Spec: Remove any reference to coordinator


rdebusscher:
There is a general consensus that this term is not needed as it is an implementation detail.

See for example https://github.com/eclipse/microprofile-lra/issues/112#issuecomment-477218921

PR #141, opened 2019-04-24
issue139 Remove references to recovery coordinators

mmusgrov:
#139 

This PR removes references to recovery coordinators. In particular, the tests that triggered a recovery scan have been replaced by a blocking wait for a configurable period (see property lra.tck.recovery.timeout below). The default is 200 seconds which is appropriate to the implementation that I tested against and if the period elapses without recovery having taken place then the test is marked as failed. Let me know if this default is too short.

I did include a hint in LraTckConfigBean#useRecoveryHeader that implementations are free to use if they want to provide more timely recovery.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTestBase.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTestBase.java
@@ -19,6 +19,7 @@
  *******************************************************************************/
 package org.eclipse.microprofile.lra.tck;
 
+import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
 import org.eclipse.microprofile.lra.tck.participant.api.Util;
 import org.jboss.shrinkwrap.api.ShrinkWrap;
 import org.jboss.shrinkwrap.api.asset.EmptyAsset;
@@ -31,6 +32,8 @@ import org.junit.rules.TestName;
 import javax.inject.Inject;
 import javax.ws.rs.client.Client;
 import javax.ws.rs.client.ClientBuilder;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.Invocation;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.Response;
 import java.net.MalformedURLException;
@@ -38,10 +41,14 @@ import java.net.URI;
 import java.net.URL;
 import java.util.logging.Logger;
 
+import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.ACCEPT_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.RECOVERY_PARAM;
+import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.TCK_PARTICIPANT_RESOURCE_PATH;
 import static org.junit.Assert.assertEquals;
 
 public class TckTestBase {
     private static final Logger LOGGER = Logger.getLogger(TckTestBase.class.getName());
+    private static final Long RECOVERY_CHECK_INTERVAL = 10L; // check for recovery every 10 seconds
 
     @Rule public TestName testName = new TestName();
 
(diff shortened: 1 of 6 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTestBase.java
```diff
@@ -131,26 +117,62 @@ String checkStatusReadAndCloseResponse(Response.Status expectedStatus, Response
...
+     * @param resource the resource on which recovery should be replayed. If null then recovery
+     *                should be attempted on all eligible participants
+     */
+    void replayEndPhase(String resource) throws InterruptedException {
+        // trigger a replay attempt on any participants
         // that have responded to complete/compensate requests with Response.Status.ACCEPTED
-        WebTarget recoveryPath = recoveryTarget.path("recovery");
-        Response response2 = recoveryPath
-                .request().get();
+        int recoveryPasses = 2; // 1 for the normal end call and 2 for the recovery pass call
```
rdebusscher: default value not useful. `recoveryPasses` gets his value from `checkStatusReadAndCloseResponse()`(line 159) and loop is controlled by time or when `checkStatusReadAndCloseResponse()` returns 0.

So probably also use another name.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -35,112 +35,42 @@
...
 import org.junit.Before;
-import org.junit.Rule;
 import org.junit.Test;
-import org.junit.rules.TestName;
 import org.junit.runner.RunWith;
 
 @RunWith(Arquillian.class)
-public class TckTests {
+public class TckTests extends TckTestBase {
     private static final Logger LOGGER = Logger.getLogger(TckTests.class.getName());
```
rdebusscher: Unused

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ParticipatingTckResource.java
```diff
@@ -79,13 +82,44 @@ public Response completeWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) String lraId,
...
+            COMPLETED_COUNT.incrementAndGet();
+        } else {
+            COMPENSATED_COUNT.incrementAndGet();
+        }
 
-        LOGGER.info(String.format("LRA id '%s' was completed", lraId));
+        return Response.ok().build();
+    }
+
+    private int recoveryPasses = 0;
```
rdebusscher: Define at the top of the class before the methods?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTestBase.java
```diff
@@ -131,26 +117,62 @@ String checkStatusReadAndCloseResponse(Response.Status expectedStatus, Response
...
+    long lraTimeout() {
         return Util.adjust(LraTckConfigBean.LRA_TIMEOUT_MILLIS, config.timeoutFactor());
     }
 
-
-    void replayEndPhase(String resource) {
-        // trigger a recovery scan which trigger a replay attempt on any participants
+    /**
+     * Trigger or wait for a recovery scan to replay the protocol termination phase (complete or compenstate)
+     * on a participants that is in need of recovery.
```
xstefank:
```suggestion
     * on a participant that is in need of recovery.
```

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTestBase.java
mmusgrov: The variable `recoveryPasses` is used to tell the resource how many times it should return `202 ACCEPTED` from the completion call on line 144 and the initial value is used. This second usage of the variable that you refer to is the resource saying how many more times it is expecting to be asked to complete and when that value hits zero we know that recovery has happened.

In other words the variable has the same semantic meaning in both usages (it is keeping track of the number completion calls: the client sets the initial value and the resource decrements it) so using the same name makes sense.

mmusgrov: @rdebusscher So can I resolve this conversation?

mmusgrov: I added more documentation explaining how the method operates

Issue #101, opened 2019-02-22
Improve testability of 'eventually consistency'


rdebusscher:
Some of the TCK tests are very brittle, for example (I'm referring to the code in https://github.com/eclipse/microprofile-lra/pull/99)

     TckCancelOnTests.cancelOnFamilyDefault4xx

After executing the GET method, we test for the status and call the resource to see if the compensate method is invoked.

But we do not mandate in the spec that this call to compensate is called immediatly.

I also understand that testing a thing like 'eventual consistency'  is basically impossible.

A bit extreme but even a system where no compensate events are sent is also 'eventual consistency'   as long as someone in the next century builds something to send out those events for all actions gathered at that time. 

So we should introduce a parameter `lra.tck.consistency.delay` which is the number of milliseconds that the test waits before it checks if the compensate method is called.

We can give it probably a default value of say 100, but this allows an implementation which has a delay for reason or another (in the range of seconds I hope and not a century) to pass the TCK.

tomjenkinson:
+1

rdebusscher:
A fixed timeout and a factor to change the timeout (and not defining a fixed delay) (meeting 4 March)

mmusgrov:
@ochaloup Can you provide an update on the status of this issue please?

ochaloup:
@mmusgrov hi. I haven't started to work on this issue. Please, feel free to grasp it if you have capacity to work on it.

mmusgrov:
The `TckCancelOnTests.cancelOnFamilyDefault4xx` test invokes a resource annotated with `@LRA(value = Type.REQUIRED)'. This means that the implementation should begin an LRA (since not context is present) when the resource method is entered and should cancel the LRA after the method returns (ie the client/test can be sure that the end phase of the protocol has been entered when the request returns). Since we are not supporting asynchronous callbacks in version 1.0, the implementation must invoke the resource compensation endpoint before returning control to the client. But the compensation endpoint in this particular test returns `Response.ok().build();` therefore the client can be certain that participant compensated and that is what the test asserts.

Note also that according to the state model, participants that enter the state `Compensating` must return `402 Accepted` when asked to compensate. But in this particular test it has returned `200 OK` which means that the test can be certain that it is in state `Compensated` and it checks that that is the case.

[...]

PR #144, opened 2019-04-27
issue101 Eventual consistency requirements

mmusgrov:
#101


tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTestBase.java
```diff
@@ -130,7 +130,20 @@ long lraTimeout() {
     }
 
     /**
-     * Trigger or wait for a recovery scan to replay the protocol termination phase (complete or compenstate)
+     * @see {@link LraTckConfigBean#consistencyDelay}
```
rdebusscher: no {@link } required so this is also ok

````
@see LraTckConfigBean#consistencyDelay
````

Issue #149, opened 2019-05-01
Add a specification note about how applications can use LRAs in reactive environments


mmusgrov:
The specification is designed for use with JAX-RS which supports asynchonous operations via:

- JAX-RS `@Suspended` annotation
- Java 8 CompletionStages

We need a section that shows examples and we also need some TCK tests to verify that implementations operate correctly when services use these async features.

xstefank:
@mmusgrov can this be closed? Both PRs are merged.

PR #150, opened 2019-05-01
issue149 Examples of async usage

mmusgrov:
#149 

One of my "reactive" colleagues reported that the spec was not clear about whether or not reactive was supported in LRA. This PR adds a [new section to the spec](https://github.com/eclipse/microprofile-lra/pull/150/files#diff-06aca14982a5ee1e4547e9ec14b41bbbR735) and [3 TCK tests](https://github.com/eclipse/microprofile-lra/pull/150/files#diff-691a384040bd56b139d108d6d1a64cb9R213) that showcase reactive support.

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -560,6 +560,7 @@ The forget method MUST be annotated with the JAX-RS `@DELETE` annotation.
 If there is no `@Status` then the `@Compensate` or `@Complete` methods will continue
 to be invoked until the implementation knows it has the final status.
 
+[[async-compensators]]
 If the resource cannot perform a compensation activity
 immediately the `@Compensate` method SHOULD do one or more of the following:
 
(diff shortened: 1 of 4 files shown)


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -731,6 +732,110 @@ Only when the implementation is certain that participant has finished
...
+    public CompletionStage<Response> asyncInvocationWithException(
+        @HeaderParam(LRA_HTTP_CONTEXT_HEADER) String lraId) {
+
+        ExecutorService es = Executors.newSingleThreadExecutor();
+        final CompletableFuture<Response> response = new CompletableFuture<>();
+
+        try {
+            es.submit(() -> {
+                // excecute long running business activity finishing with a NOT_FOUND error
+                // is one of the default LRA#cancelOn codes that should cause the LRA to cancel
```
xstefank: minor, this is true but NOT_FOUND is explicitly defined for this LRA

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -731,6 +732,110 @@ Only when the implementation is certain that participant has finished
...
+In addition to the use of completion stages, a resource method may also produce
+asynchronous responses by injecting a JAX-RS `AsyncResponse` parameter using
+the JAX-RS `@Suspended` annotation:
+
+[source,java]
+----
+    @LRA(value = LRA.Type.REQUIRED, // the method must run with an LRA
+         end = true) // the LRA must end when the method completes
+    @Path("asyncresponse-lra")
+    public CompletionStage<Response> asyncResponseWithLRA(
```
xstefank: seems this is forgotten previous method definition

tomjenkinson: +1

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: I don't think these are valid comments. The example demonstrates that reactive and LRA can be combined,.

mmusgrov: okay

mmusgrov:
@rdebusscher I didn't understand your point about custom wrappers. The example code just runs some async logic. When that logic completes exceptionally with a code that matches one of the cancelOn codes  the spec implementation cancels the LRA which will eventually trigger the compensation callbacks on any registered participant.

mmusgrov:
But yes Rudy you make an excellent point. I will merge this PR and ask @xstefank to allow a signature which returns a CompletionStage<type>. When his PR is merged we can update the reactive examples.

rdebusscher:
@mmusgrov My comment about wrappers was not correct, sorry

I had the idea yesterday evening that we need custom implementations of CompletionStage and AsyncResponse to be able to know when the response was finished and the the LRA could be ended. (since the method returns before the processing is finished)

But I'm wrong, the JAX-RS filter will catch the end of the business processing correctly.

mmusgrov:
So Rudy, the examples relate to methods annotated with `@LRA` which are always JAX-RS endpoints so the implementation can use JAX-RS filters. When the async logic finishes the filter can be used to decide whether or not the LRA should be ended. Our Narayana implementation of the spec already does this.

A slightly different issue is that the `@Compensate` annotation does not need to be a CDI interceptor because the spec implementation invokes the compensate method. But if the annotated method is to run async then Martin needs to add the option for returning `CompletionStage<type>`.

mmusgrov:
Ah, Rudy I sent my comment at the same time as you did. But I see that we are in agreement so I will merge this PR. We still need the new non-JAX-RS signature to support returning completion stages which we can discuss on the other PR.

xstefank:
@mmusgrov you are right, the implementation is responsible for calling compensate method when it's not bound to JAX-RS so we can definitely implement CompletionStage as supported return type. Would that be all? I know there are several other "reactive" types that can be returned in rxJava, reactive streams etc. The question is what everything we want to cover in the spec.

BTW we have the implementation of everything in the second PR ready for narayana :)

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -731,6 +732,107 @@ Only when the implementation is certain that participant has finished
 will it tell it that it is okay to release any resources it associated
 with the LRA.
 
+[[reactive-support]]
+==== Reactive Support
+
+Implementations are expected to operate correctly when services use the
+asychronous and reactive features provided by JAX-RS.
```
cescoffier: missing `n`

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -731,6 +732,107 @@ Only when the implementation is certain that participant has finished
...
+
+[source,java]
+----
+    @LRA(value = LRA.Type.REQUIRED, // the method must run with an LRA
+         end = true) // the LRA must end when the method completes
+    @Path("completion-stage-lra")
+    public CompletionStage<Response> asyncInvocationWithLRA(
+        @HeaderParam(LRA_HTTP_CONTEXT_HEADER) String lraId) { // the id of LRA
+
+        ExecutorService es = Executors.newSingleThreadExecutor();
```
cescoffier: I'm not sure it's a good idea to show to the user the creation of a thread pool in a JAX-RS method. 

What about:

```
service.doSomethingAsynchrnously() // returns a CompletationStage<Void>
    .thenApply(x -> Response.ok().entity(lraId).build())
```

cescoffier: It would also remove the confusing thread pool shutdown.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -731,6 +732,107 @@ Only when the implementation is certain that participant has finished
...
+----
+    @LRA(value = LRA.Type.REQUIRED, // the method must run with an LRA
+         end = true, // the LRA must end when the method completes
+         cancelOn = {Response.Status.NOT_FOUND}) // cancel on a 404 code
+    @Path("completion-stage-exceptionally-lra")
+    public CompletionStage<Response> asyncInvocationWithException(
+        @HeaderParam(LRA_HTTP_CONTEXT_HEADER) String lraId) {
+
+        ExecutorService es = Executors.newSingleThreadExecutor();
+        final CompletableFuture<Response> response = new CompletableFuture<>();
```
cescoffier: Same comment here. The example I've proposed above works there too.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -731,6 +732,107 @@ Only when the implementation is certain that participant has finished
...
+the JAX-RS `@Suspended` annotation:
+
+[source,java]
+----
+    @LRA(value = LRA.Type.REQUIRED, // the method must run with an LRA
+         end = true) // the LRA must end when the method completes
+    @Path("asyncresponse-lra")
+    public void asyncResponseLRA(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) String lraId,
+                         final @Suspended AsyncResponse ar) {
+        ExecutorService es = Executors.newSingleThreadExecutor();
```
cescoffier: Same comment.

```
service.doSomethingAsync()
  .thenAccept(x -> ar.resume(Response.ok().entity(lraId).build());
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: @xstefank preferred that option too

PR #151, opened 2019-05-04
issue149 Do no show creation of an executor in the doc examples

mmusgrov:
#149 

This PR follows up on https://github.com/eclipse/microprofile-lra/pull/150 which was merged before Clement had the opportunity to contribute (sorry @cescoffier )

The change request was to not show the creation of a thread pool in the example (since thread pools are normally managed by the container).


mmusgrov:
@cescoffier I added your AWS S3 example, thanks

PR #152, opened 2019-05-19
Fix +1 problem in replayEndPhase

xstefank:
replayEndPhase method is calling this resource for querying whether recovery has happened. It initializes recoveryPasses to 2 (see [this](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTestBase.java#L167)) and expects that this value will be decremented two times. However, the current version of the spec first checks the value and then decrements it which results into this newly stared LRA being not finished by the end of the test. Typically it is finished by the next test that does replayEndPhase again but that leaves the newly started LRA not finished and so on. This means that at the end of the TCK there is at least one LRA hanging in Cancelling or Completing state.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ParticipatingTckResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ParticipatingTckResource.java
@@ -90,9 +90,7 @@ public class ParticipatingTckResource {
     }
 
     private Response getEndPhaseResponse(boolean complete) {
-        if (recoveryPasses > 0) {
-            recoveryPasses -= 1;
-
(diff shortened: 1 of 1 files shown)


mmusgrov:
What problem have you noticed that your fix is resolving?

mmusgrov:
Is it s performance issue?

xstefank:
@mmusgrov No, I described it in description of this PR - it takes 3 calls to `ParticipatingTckResource` to close/cancel LRA but replayEndPhase assumes only 2. This leaves LRA started in replayEndPhase open in closing/cancelling state after test ends.

mmusgrov:
Okay

xstefank:
@mmusgrov Thanks!

PR #153, opened 2019-05-20
issue146 Allow JAX-RS endpoints Complete/Compensate to be invoked usi…

### LINKED ISSUE #146: Allow JAX-RS endpoints Complete/Compensate to be invoked using any HTTP method
When the Compensate or Complete methods are JAX-RS endpoints they should be allowed to be annotated with any HTTP method, ie the following spec requirement is too restrictive:

> The application developer indicates which JAX-RS method to use for a compensating
activity by annotating it with both the `@Compensate` and the JAX-RS `@PUT` annotations.

and

> ... If the annotation is not accompanied by
> a JAX-RS `@PUT` annotation the error SHOULD be reported using a JAX-RS
> exception mapper that maps to a `412 Precondition Failed` HTTP status code.

xstefank:
…ng any HTTP method

resolves #146 

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java
@@ -33,10 +33,10 @@ import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
-import javax.ws.rs.DELETE;
 import javax.ws.rs.GET;
 import javax.ws.rs.HeaderParam;
 import javax.ws.rs.NotFoundException;
+import javax.ws.rs.POST;
 import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 import javax.ws.rs.WebApplicationException;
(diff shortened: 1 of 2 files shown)


mmusgrov:
We should have at least one test that uses an HTTP method that is different from the recommended HTTP method:

- `@PUT` `@Forget`
- `@PUT` `@Status`
- `@Compensate` `@GET` 
- `@Complete` `@POST`

xstefank:
+1, I was thinking whether we want to have it in TCK. I will update exisiting test.

xstefank:
@mmusgrov I've update ContextTckResource with your suggestions

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java
```diff
@@ -345,7 +345,7 @@ public Response leave(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) String lraId) {
         return Response.ok().entity(lraId).build();
     }
 
-    @PUT
+    @GET
```
mmusgrov: This is is quite a big change to the spec so I would advise that we at least include a comment, something like:

> @GET // @PUT SHOULD be used but verify that other HTTP methods are supported on compensations (and ditto for the completion, status and forget callbacks).

mmusgrov:
@rdebusscher Do you mind if I merge Martin's PR?

PR #154, opened 2019-05-20
issue131 Describe security requirements of the JAX-RS endpoint genera…

### LINKED ISSUE #131: Spec: Describe security requirements of the JAX-RS endpoint generated by the Proxy participant.
When the methods annotated with @Complete or @Compensate are not defined as JAX-RS resources, the implementation (this is implementation specific as not all implementations will use REST communication between participant and coordinator) can provide JAX-RS endpoints by the Proxy participant which is generated automatically.

These endpoints needs to be protected by default in such a way that they can't be exploited by hackers. They should ideally only be called by the coordinator and not by any other party.

xstefank:
…ted by the Proxy participant

resolves #131 

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -684,6 +684,10 @@ or may in the process of compensation (completion) the exception in these method
 result into failure condition (in JAX-RS this condition is represented by 
 500 return HTTP status code) which individual interpretation is left further unspecified.
 
+In case the implementation of this specification exposes non-JAX-RS participant methods 
+to be able to call them externally (e.g. the HTTP proxy) then it MUST protect every 
+exposed method from unauthorized access. The specific security details are not specified.
+
 [[eventual-compensations]]
 ==== Eventual compensations
 


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -684,6 +684,10 @@ or may in the process of compensation (completion) the exception in these method
 result into failure condition (in JAX-RS this condition is represented by 
 500 return HTTP status code) which individual interpretation is left further unspecified.
 
+The non-JAX-RS participant methods MUST be invoked by the implementation of this specification
```
rdebusscher: I would say

> ... invoked by the implementation of the specification **only** and the implementation MUST ...

otherwise the sentence seems a bit odd to me.

xstefank:
@rdebusscher PR updated, thanks for the review.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -684,6 +684,10 @@ or may in the process of compensation (completion) the exception in these method
 result into failure condition (in JAX-RS this condition is represented by 
 500 return HTTP status code) which individual interpretation is left further unspecified.
 
+The non-JAX-RS participant methods MUST be invoked by the implementation of the
```
mmusgrov: The business author can call whichever methods he wants and if that results in something undesirable then he should re-code his app - ie I don't think this is doing anything to enhance the spec.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -684,6 +684,10 @@ or may in the process of compensation (completion) the exception in these method
 result into failure condition (in JAX-RS this condition is represented by 
 500 return HTTP status code) which individual interpretation is left further unspecified.
 
+The non-JAX-RS participant methods MUST be invoked by the implementation of the 
+specification only. The implementation MUST protect them from unauthorized access.
```
mmusgrov: Aren't the non JAX-RS methods local in-VM calls, in which case this statement is vacuous

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -684,6 +684,10 @@ or may in the process of compensation (completion) the exception in these method
 result into failure condition (in JAX-RS this condition is represented by 
 500 return HTTP status code) which individual interpretation is left further unspecified.
 
+The non-JAX-RS participant methods MUST be invoked by the implementation of the 
+specification only. The implementation MUST protect them from unauthorized access. 
+The specific security details are not unspecified.
```
mmusgrov: Don't you mean one of "are unspecified" or "are not specified"

mmusgrov:
So this one looks complete now so I am merging

PR #155, opened 2019-05-20
issue130 Specify behaviour when @Complete and @Compensate methods are…

### LINKED ISSUE #130: Spec: Specify behaviour when @Complete and @Compensate methods are called by developer (and not by coordinator)
Especially when the @Complete and @Compensate methods are exposed as JAX-RS endpoints, the developer can call these methods from an application.

We need to describe in the spec document what happens when

> For an active LRA the JAX-RS endpoint/method is called outside the cancelling or closing events initiated by termination of the LRA by a _end=true_ call. So the method is called by the developer and not by the coordinator.

**Possible solutions**
1- The LRA is closed or cancelled just as if a call is made to an endpoint which has _end=true_. We need to make sure that when the LRA is closed/cancelled, the participant which initiated this termination does not receive another complete/compensate calls (as that would mean the method is called twice)
2- There is no effect on the LRA and the method body is executed and the return value indicate the status of the participant for that LRA (most likely being Active)

xstefank:
… called by developer

resolves #130 

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -615,6 +615,10 @@ If the participant annotation is not accompanied by an associated JAX-RS HTTP me
 annotation according to the <<jaxrs-response-table,table above>>, the error SHOULD be reported using a JAX-RS 
 exception mapper that maps to a `412 Precondition Failed` HTTP status code.
 
+Users are allowed to reuse existing JAX-RS endpoints for participant methods definitions. 
+In this case, LRA implementation MUST ensure that invoking these methods outside of the 
+implementation of the LRA specification will not influence any running LRA.
+
 [[non-jaxrs-participant-methods]]
 ===== Non-JAX-RS participant methods
 


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -615,6 +615,10 @@ If the participant annotation is not accompanied by an associated JAX-RS HTTP me
 annotation according to the <<jaxrs-response-table,table above>>, the error SHOULD be reported using a JAX-RS 
 exception mapper that maps to a `412 Precondition Failed` HTTP status code.
 
+Users are allowed to reuse existing JAX-RS endpoints for participant methods definitions. 
+In this case, LRA implementation MUST ensure that invoking these methods outside of the 
+LRA orchestration will not influence any running LRA.
```
mmusgrov: Again, I think this is just noise, but if you want to add it then I will still approve.

xstefank: ok

mmusgrov: So what was your opinion about this comment? Are you pushing back or accepting my input?

xstefank:
@mmusgrov sorry, I forgot to push. Should be ok now.

PR #156, opened 2019-05-21
Add a note about replayEndPhase expectations in TCK

xstefank:

### DIFF
--- a/tck/running_the_tck.asciidoc
+++ b/tck/running_the_tck.asciidoc
@@ -109,3 +109,9 @@ Debugging is dependent on the TCK implementor. The implementor configures Arquil
 to deploy the test deployments and run the LRA client side implementation there. In general we can say that
 the runtime needs to define java debug properties `-Xdebug -Xrunjdwp:transport=dt_socket,address=8787,server=y,suspend=y`.
 Then debugger may be connected to port 8787 and to track the test execution.
+
+== Additional notes
+
+* `TckTestBase#replayEndPhase(String)` method, used for instance in `@Status` and 
+`@Forget` call verifications, expects that recovery happens on all active LRAs in 
+a reasonable time frame defined by the configuration property `lra.tck.recovery.timeout`.


tck/running_the_tck.asciidoc
```diff
@@ -109,3 +109,10 @@ Debugging is dependent on the TCK implementor. The implementor configures Arquil
 to deploy the test deployments and run the LRA client side implementation there. In general we can say that
 the runtime needs to define java debug properties `-Xdebug -Xrunjdwp:transport=dt_socket,address=8787,server=y,suspend=y`.
 Then debugger may be connected to port 8787 and to track the test execution.
+
+=== Additional notes
+
+* `TckTestBase#replayEndPhase(String)` method, used for `@Status` and `@Forget` calls
```
rdebusscher: > used for `@Status` and `@Forget` call verifications, expects that recovery happens on all active LRAs in a reasonably time frame

-> adjusted a few words.

Aren't we using the (configurable) timeouts also for these calls? Then we should refer to the parameter to change the factor for the timeout here also.

tck/running_the_tck.asciidoc
```diff
@@ -109,3 +109,11 @@ Debugging is dependent on the TCK implementor. The implementor configures Arquil
...
 the runtime needs to define java debug properties `-Xdebug -Xrunjdwp:transport=dt_socket,address=8787,server=y,suspend=y`.
 Then debugger may be connected to port 8787 and to track the test execution.
+
+=== Additional notes
+
+* `TckTestBase#replayEndPhase(String)` method, used for `@Status` and `@Forget` call 
+verifications, expects that recovery happens on all active LRAs in a reasonable 
+time frame defined by the config property `lra.tck.recovery.timeout`. The 
+implementation must ensure that recovery is triggered for every active LRA at 
+approximately same time.
```
mmusgrov: The bit of text about "approximately same time" is just noise but I will approve anyway.

tck/running_the_tck.asciidoc
```diff
@@ -109,3 +109,11 @@ Debugging is dependent on the TCK implementor. The implementor configures Arquil
 to deploy the test deployments and run the LRA client side implementation there. In general we can say that
 the runtime needs to define java debug properties `-Xdebug -Xrunjdwp:transport=dt_socket,address=8787,server=y,suspend=y`.
 Then debugger may be connected to port 8787 and to track the test execution.
+
+=== Additional notes
+
+* `TckTestBase#replayEndPhase(String)` method, used for `@Status` and `@Forget` call
```
mmusgrov: It is also used for testing participants that compensate in the background.

tck/running_the_tck.asciidoc
xstefank: I am struggling how to state this. This is exactly the point of this PR. "lra.tck.recovery.timeout" sets the maximum time until which recovery should happen, let's say it is 2 minutes. But when recovery happens on ParticipatingTckResource then we have one second interval after which we check that actual TCK resource under test recovered also. So it's not possible to recover one LRA  after 10s and another after 110s. And this what I am trying to say with approximately same time. Any suggestions?

tck/running_the_tck.asciidoc
xstefank: Thanks! I will add "for instance" in this sentence.

tck/running_the_tck.asciidoc
mmusgrov: The previous paragraph is sufficient so just delete the second one.

tck/running_the_tck.asciidoc
mmusgrov: Is this discussion resolved?

mmusgrov: @rdebusscher the updated text says:

> ...a reasonable time frame defined by the configuration property `lra.tck.recovery.timeout`.

mmusgrov: ie I think this PR is in a state where it can be committed - please could you add your approval?

PR #158, opened 2019-05-27
TCK participant metrics refactoring

xstefank:
Refactoring of TCK participant metrics (counters of Compensate, Complete,...) to CDI bean instead of JAX-RS. This removes duplication of counters tracking in individual participant resources and as CDI service can be injected to both resources and tests this also lowers the number of JAX-RS calls.


mmusgrov:
@xstefank It looks good from an initial skim - I will try to find the time tomorrow to review it in detail.

xstefank:
> Looks good but why didn't you also applying the same pattern [of using metrics] to the LraController and ParticipatingTckResource resource classes?

Sorry, the result of doing multiple things at the same time. I will add them asap. Thanks for the review!

xstefank:
@mmusgrov updated

xstefank:
@rdebusscher thanks! PR updated.

Issue #121, opened 2019-03-19
Annotation descriptions are inconsistent for LRA


mmusgrov:
github does not support moving issues (except in the beta version) so this issue has been transferred from https://github.com/eclipse/microprofile-sandbox/issues/45 which was raised by @nmcl with the description:

> With reference to https://github.com/eclipse/microprofile-sandbox/tree/master/proposals/0009-LRA#java-annotations-for-lras
> 
> The doc spends over a page showing the @lra annotation with examples but then only talks about how the other annotations, such as @compensate, are used. For consistency and so someone can see precisely what annotations are being introduced, I suggest putting in as much detail for these other annotations as there is for @lra.

PR #159, opened 2019-05-27
issue121: Add more examples showing participant compensation behaviour

mmusgrov:
#121

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -414,8 +414,9 @@ or closed (`@Complete` is invoked):
 @ApplicationScoped
 public class SimpleLRAParticipant
 {
-    @Path("/cdi")
     @LRA(LRA.Type.REQUIRES_NEW)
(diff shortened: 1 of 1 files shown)


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -499,29 +499,132 @@ the method finishes and the `end = true` element on the confirmTrip method force
...
+Under failure conditions the system will keep retrying until it is certain that the
 participant has been successfully notified.
 
-An example of how to use this annoation is:
+The LRA model is quite flexible in how business methods perform compensations.
+If a compensating activity is brief then the synchronous model may be appropriate:
 
 [source,java]
 ----
-  @PUT
```
xstefank: Leave the `@PUT` annotation please. The example seems incomplete from JAX-RS perspective when only `@Path` is used on method. (also in other examples please)

mmusgrov:
@rdebusscher I correct the case for the `@Path` annotation. All comments are now resolved so I will now merge this PR.

PR #161, opened 2019-06-05
Non-JAX-RS complete/Compensate: allow String parameters

### LINKED ISSUE #160: Allow String parameters as parameter type for Complete and Compensate Methods in the non JAX-RS case
Through Issue https://github.com/eclipse/microprofile-lra/issues/35 (PR https://github.com/eclipse/microprofile-lra/pull/148) the only allowed parameter type on the Complete and Compensate Methods is `java/net.URI`

    @Compensate
    public void compensateAction(URI lraId) {

However, in the JAX-RS version, the most common parameter types will be String.

    @PUT
    @Path("/compensate")
    @Produces(MediaType.APPLICATION_JSON)
    @Compensate
    public Response compensateWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) String lraId) {

In order to make the spec more uniform, String should also be allowed as parameter types for Complete and Compensate Methods in the non JAX-RS case.

    @Compensate
    public void compensateAction(String lraId) {

rdebusscher:
Fixes #160.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/participant/InvalidLRAParticipantDefinitionException.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/participant/InvalidLRAParticipantDefinitionException.java
@@ -32,9 +32,9 @@ package org.eclipse.microprofile.lra.participant;
  *  - java.util.concurrent.CompletionStage: with the parameter of any of the previously 
  *    defined types
  * 
- * Arguments: up to 2 arguments of types in this order:
- *  - {@link java.net.URI}: representing current LRA context identification
- *  - {@link java.net.URI}: representing potentional parent LRA context identification
+ * Arguments: up to 2 arguments in this order:
+ *  - {@link java.net.URI} or {@link java.lang.String}: representing current LRA context identification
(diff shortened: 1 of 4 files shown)


rdebusscher:
@xstefank Your feedback is also appreciated.

rdebusscher:
@mmusgrov the LRA id **MUST** be an URI, since for the Complete and Compensate methods in the non-jaxrs case, we require URI as parameter type.

mmusgrov:
@rdebusscher That's a shame

rdebusscher:
@mmusgrov We can remove any reference to URI in the API and Spec and replace it with String and define in the spec that LRA id must be a unique String and preferably a URI.

xstefank:
@rdebusscher sorry, I don't agree with this. URI (previously URL) enforces uniqueness. URI can always be transformed to String (in header or in non-JAX-RS with this PR) but not other way around. I think the spec is more robust if the uniqueness is mandated by the API. My 2c.

rdebusscher:
We are going for URI type only. See #163 for the required changes.

Issue #35, opened 2018-10-29
Do not define JAX-RS annotations on @Compensate methods


rdebusscher:
It is up to the implementation on how the Compensate methods are called.

This could be through JAX-RS calls but this should not be the only way (although an implementation will probably only implement one method).
 This makes it easier to use the 'events' way of coordination.

tomjenkinson:
This is the annotation: https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java

tomjenkinson:
This should help with https://github.com/eclipse/microprofile-lra/issues/20

mmusgrov:
I am wondering how this could be implemented in crash recovery scenarios. When a crashed service restarts and the LRA finishes the implementation needs to decide which instance to deliver the callback to. Is it enough to just mandate that, if the @Compensate or @Complete method is not a JAX-RS resource method then, the whole class must be serializable. Adding this requirement means that the implementation can then ensure that it reliably saves the serialized state so on restart it can deserialize it and issue the callback to the correct instance. Note that this requirement is mandated in the description of the the programmatic API: https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#34-java-based-lra-participant-registration-api

rdebusscher:
I do not see why there are any additional requirements needed in the case we do not explicitly set the JAX-RS annotations on the @Compensate methods.
Behind the scenes, it works exactly the same as in the case with the JAX-RS annotations.

If the service is down at the moment the coordinator tries to call the compensate (or complete) method, it will try later on to do this.

Related to this, I disagree that the resource must be serializable when an implementation does not choose to use the JAX-RS communication.

JAX-RS resources are not an island and the classes have references to a lot of other (CDI ) classes. In order to be able to serialize and instantiate the classes at the coordinator, it will need access to almost the complete application, not to mention the classloading and class version mismatches which may occur.

So I advise that we should rethink the requirement of the serialization.

On the other hand, I think we should REQUIRE that each coordinator support the JAX-RS way of calling compensate and complete methods.
This to ensure the interoperability of the different implementations.

If a certain vendor A chooses to only implement a custom protocol but the service running on an MP implementation of Vendor B which automatically exposes only the JAX-RS way of working, they will not be compatible.

tomjenkinson:
[...]

PR #162, opened 2019-06-09
issue35 revisited (Reinstate LRAParticipant)

mmusgrov:
Can we discuss again why the original LRAParticipant interface that we had back in 2017 was insufficient for solving Rudy's issue #35. It had the good features of:

- providing compile time validation that the callback signatures are correct
- use the IDE to tell you the signature rather than having to consult the spec
- is arguably easier for developers
  * no multiple signature and return types to worry about
- is easier to implement
  * eg Narayana already had an implementation for it and some quickstarts

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/participant/AsyncLRAParticipant.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/participant/AsyncLRAParticipant.java
@@ -1,0 +1,33 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2018 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


rdebusscher:
@mmusgrov I think we removed it, not because it was a problem for issue 35 but

- with LRAParticipant interface, every JAX-RS resource which wants to be a participant / have the @LRA MUST be an implementation of the interface.
- The Complete method becomes mandatory to implement (or we need a default method)
- One cannot mix the non-async and the async methods (for example returning Participant for complete and CompletionStage for Compensate)
- During the last (10-15) years, interfaces are removed to indicate a certain type of resource and only annotation is used. So this requirement goes against the trend.
- If we only need LRAParticipant for the non-JAX-RS case, why the unnecessary distinction? They still will need to lookup how they need to write the compensate and complete methods.

So -1 for me on this one.

The implementation of issue 35 is fine as it is. If we do not want the String parameter types (requested in issue160) that is fine for me if:
- We define in the spec that LRA id MUST be an URI (which we already require today, also when we agree on this PR)
- We remove all usages of String as parameter type in the examples
- Only URI methods as parameters (also for the JAX-RS methods) are spec compliant. When implementation allows String parameter on the Complete/Compensate  it is an implementations choice (extension by that implementation)

xstefank:
@tomjenkinson something like this is possible with currently merged solution for issue 35
```java
    @Compensate
    public void cancelBooking(URI lraId) {
        if (//something) {
            throw new RuntimeException(); // results in FailedToCompensate
        } else if (// something else) {
            throw new WebApplicationException(Response.accepted().build()); // indicate Compensating
        }
        // else cancel the booking with id lraId
    }
```

mmusgrov:
Since I am the only one that wants this contribution I will withdraw it.

Issue #171, opened 2019-06-19
A bug in TckContextTests


mmusgrov:
There are two issues with the [resource used by the context tests](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java):

1 some of the methods have been changed to return a URI in PR 165 [1]
2 it contains an ambiguous JAX-RS path [2] and [3]: 

The fix for 1 is to either return a string or go with Martin's suggestion of adding a MessageBodyWriter (I have gone for the latter). 

The fix for 2 is to use POST for the second path.

[1] https://github.com/eclipse/microprofile-lra/pull/165
[2] https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L315
[3] https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L344

PR #172, opened 2019-06-19
issue171 Duplicate resource and missing URI writer

mmusgrov:
Signed-off-by: Michael Musgrove <mmusgrov@redhat.com>

#171 
Fixes a couple of issues with the resource used fo the TckContextTests

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckContextTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckContextTests.java
@@ -43,6 +43,7 @@ import static org.eclipse.microprofile.lra.tck.TckContextTests.HttpMethod.PUT;
 import static org.eclipse.microprofile.lra.tck.participant.api.ContextTckResource.ASYNC_LRA_PATH1;
 import static org.eclipse.microprofile.lra.tck.participant.api.ContextTckResource.ASYNC_LRA_PATH2;
 import static org.eclipse.microprofile.lra.tck.participant.api.ContextTckResource.ASYNC_LRA_PATH3;
+import static org.eclipse.microprofile.lra.tck.participant.api.ContextTckResource.CLEAR_STATUS_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.ContextTckResource.CONTEXT_CHECK_LRA_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.ContextTckResource.LEAVE_PATH;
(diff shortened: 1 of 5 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/UriMessageBodyWriter.java
```diff
@@ -0,0 +1,54 @@
...
+    public long getSize(URI uri, Class<?> aClass, Type type, Annotation[] annotations, MediaType mediaType) {
+        return -1;
+    }
+
+    @Override
+    public void writeTo(URI uri, Class<?> aClass, Type type, Annotation[] annotations, MediaType mediaType,
+                        MultivaluedMap<String, Object> multivaluedMap, OutputStream outputStream) throws IOException, WebApplicationException {
+        outputStream.write(uri.toASCIIString().getBytes());
+//        DataOutputStream dos = new DataOutputStream(outputStream);
+//        dos.writeUTF(uri.toASCIIString());
```
xstefank: probably forgotten comments, can be removed

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckContextTests.java
```diff
@@ -103,7 +103,7 @@ public void testStatus() throws InterruptedException {
         assertEquals(testName.getMethodName() + ": participant is not completing", ParticipantStatus.Completing.name(), status);
 
         // clear the EndPhase override data so that the next status request returns completed or compensated
-        invoke(false, STATUS_PATH, PUT, lra, 200, ContextTckResource.EndPhase.SUCCESS, 200);
+        invoke(false, STATUS_PATH, HttpMethod.POST, lra, 200, ContextTckResource.EndPhase.SUCCESS, 200);
```
xstefank: sorry if I'm missing something here, but this should call [this method](https://github.com/eclipse/microprofile-lra/pull/172/files#diff-4c219933e42e7b38d324e42b907623a2R342) right? I think it should be GET here and in second clear invocation on line 156.

mmusgrov: @xstefank you are correct. I separated out the two "status" method to avoid confusion. I also added some other fixes for the "String to URI" change.

mmusgrov: @xstefank I fixed the PR. I am on PTO tomorrow and need this fix for the narayana implementation so I am going to merge in order to get the snapshot build. I've tested my changes but if you do spot anything please update the issue or PR.

Issue #164, opened 2019-06-12
Include an annotation to allow parties to discover the final outcome of an LRA


mmusgrov:
It may be useful to allow discovery of the [final state](https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc
) of an LRA. The end states FailedToCancel and FailedToClose are likely to be of particular interest to some types of client. And for the FailedToCancel and Cancelled state it may be interesting to know whether or not it was due to the LRA timing out or due to a participant timing out.

mmusgrov:
This topic was requested/discussed in a [forum post](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/microprofile/Ye0wqcegGKY/_Vdd5t6dBQAJ). It's worth pointing out that the feature is not new and is provided in a number of existing transaction standards and proposals such as:

* JTA specification: Section 3.3.2 Transaction Synchronization
* Oasis WS-AT: http://docs.oasis-open.org/ws-tx/wstx-wsat-1.1-spec-pr-01/wstx-wsat-1.1-spec-pr-01.html#_Toc144697711
* Narayana REST-AT: http://narayana.io//docs/project/index.html#_pre_and_post_two_phase_commit_processing

tomjenkinson:
Synchronizations are certainly useful in many transaction situations but personally I think this could go in 1.1

mmusgrov:
Yesterday I included a message on the forum to see if there were any opinions on this feature. If we have no feedback after the next hangout I will mark this for inclusion in milestone 2.

xstefank:
@mmusgrov what is the proposed way of how this information is to be shared with the participant?

rdebusscher:
We can expose the status of all 'known' LRAs (and the participants status) through an URL, like /lra (similar to /openapi and /metrics)

xstefank:
@rdebusscher would this endpoint be present on every participant? And every participant will have knowledge of all LRAs? Even the ones it is not the part of?

rdebusscher:
@xstefank No, we can create a URL which just fetches the statuses from the coordinator. So not for each participant, but for each deployment.

tomjenkinson:
I would think something like:

```
@Synchronization
public void afterCompletion(LRAId lraId, LRAStatus status) {
}
```

Would be sufficient?

tomjenkinson:
If this goes into 1.0 there should not be an RC without it.

[...]

PR #174, opened 2019-06-24
issue164 Support notifications that an LRA has finished

mmusgrov:
#164 

If a resource is invoked in the context of an LRA and it contains a JAX-RS resource method annotated with `@AfterEnd` then it will be notified when the associated LRA is finished (in any of the terminal states).

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
@@ -1,0 +1,82 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.lra.annotation;
+
+import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * <p>
+ * If a JAX-RS resource method is annotated with
+ * {@link LRA} and is invoked in the context of an LRA then
+ * the resource can ask to be notified when the LRA finishes
+ * by marking one of the other methods in the class with
+ * the {@link AfterLRA} annotation.
+ * </p>
+ *
+ * <p>
+ * If the <code>AfterLRA</code> method is also a JAX-RS resource method
+ * then the LRA context is made available to the annotated method
+ * via an HTTP header with the name
+ * {@link LRA#LRA_HTTP_ENDED_CONTEXT_HEADER} and the
+ * final status is passed to the method as plain text
+ * corresponding to one of the {@link LRAStatus} enum values.
+ * For example:
+ * <pre>
+ *     <code>
+ * @PUT
+ * @AfterLRA
+ * public Response afterEnd(@HeaderParam(LRA_HTTP_ENDED_CONTEXT_HEADER) URI lraId,
+ *                          LRAStatus status)
+ *     </code>
+ * </pre>
+ * </p>
+ *
+ * <p>
+ * The implementation will keep resending the notification
+ * until it receives a <code>200 OK</code> status code from the
+ * resource method (which means that the method SHOULD be
+ * idempotent).
+ * </p>
+ *
+ * <p>
+ * If the <code>AfterLRA</code> method is not a JAX-RS resource method
+ * then the id of the LRA and its final status can be obtained
+ * by ensuring that the annotated method conforms to the
+ * signature:
+ *
+ * <code>
+ *     public void onLRAEnd(URI lraId, LRAStatus status)
+ * </code>
+ *
+ * </p>
+ * <p>
+ * The return type is ignored and the method name is not
+ * significant.
+ * </p>
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+public @interface AfterLRA {
+}
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
@@ -23,6 +23,7 @@ package org.eclipse.microprofile.lra.annotation.ws.rs;
 import javax.ws.rs.HeaderParam;
 import javax.ws.rs.core.Response;
 
+import org.eclipse.microprofile.lra.annotation.AfterLRA;
 import org.eclipse.microprofile.lra.annotation.Compensate;
 import org.eclipse.microprofile.lra.annotation.Complete;
 import org.eclipse.microprofile.lra.annotation.Forget;
@@ -102,6 +103,12 @@ public @interface LRA {
     String LRA_HTTP_CONTEXT_HEADER = "Long-Running-Action";
 
     /**
+     * Header name holding the LRA context of an LRA that has finished - to be
+     * used in conjunction with the {@link AfterLRA} annotation.
+     */
+    String LRA_HTTP_ENDED_CONTEXT_HEADER = "Long-Running-Action-Ended";
+
+    /**
      * When a JAX-RS invocation is made with an active LRA which is nested,
      * the parent LRA is made available via an HTTP header field with the
      * following name. The value contains the parent LRA id associated with
@@ -132,6 +139,12 @@ public @interface LRA {
      *     but it does guarantee that they will eventually be sent. Under failure
      *     conditions the system will keep retrying until it is certain that all
      *     participants have been successfully notified.
+     * </p>
+     *
+     * <p>
+     *     If the method is to run in the context of an LRA and the annotated class
+     *     also contains a method annotated with {@link AfterLRA}
+     *     then the resource will be notified of the final state of the LRA.
      * </p>
      *
      * <p>
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -501,6 +501,52 @@ The `end = false` element on the bookTrip method forces the LRA to continue runn
 the method finishes and the `end = true` element on the confirmTrip method forces the LRA
 (started by the bookTrip method) to close the LRA.
 
+[[after-lra]]
+===== Discovering the Outcome of an LRA
+
+As remarked in the previous section, a JAX-RS resource method runs with an active
+context depending upon the value specified in the `@LRA` annotation.
+The final state of this LRA can be discovered by marking one of the other methods
+in the class with the `@AfterLRA` annotation. When the LRA enters a terminal state
+the method will be passed the id of the LRA together with the `LRAStatus` (see the
+javadoc for the `@AfterLRA` annotation for more information). Note that the final
+states of an LRA is defined by <<lra-state-model,the LRA state model>> and it
+would be a specification violation if the implementation calls the method when the
+LRA is not in a final state. Further information about method signatures and
+JAX-RS response codes is given below in section <<jaxrs-participant-methods>>.
+
+Note that the resource does not need to be a participant in order to receive this
+notification. Therefore in the following resource definition, although no method
+is annotated with `@Compensate`, if the method called
+`activityWithLRA` is invoked then the method `notifyLRAFinished` will be called
+when the LRA finishes:
+
+[source,java]
+----
+public class BusinessResource {
+    @PUT
+    @Path("/work")
+    @LRA(value = LRA.Type.REQUIRES_NEW)
+    public Response activityWithLRA(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
+        // perform a business action in the context of lraId
+        return Response.ok().build();
+    }
+
+    @PUT
+    @Path("/after")
+    @AfterLRA // method is called when the LRA associated with the method activityWithLRA finishes
+    public Response notifyLRAFinished(@HeaderParam(LRA_HTTP_ENDED_CONTEXT_HEADER) URI lraId,
+                                      LRAStatus status) {
+
+        switch (status) {
+            case Closed:
+                // the LRA was successfully closed
+            ...    
+        }
+    }
+}
+----
+
 [[compensating-activities]]
 ==== Compensating Activities
 
@@ -545,7 +591,7 @@ If a compensating activity is brief then the synchronous model may be appropriat
 
 On the other hand, the compensating logic may involve concerted activities,
 perhaps even compensating in the context of another LRA. In this case the
-protocol accomodates a more decoupled mode of operation - the following example
+protocol accommodates a more decoupled mode of operation - the following example
 shows how a compensating activity can be started in the background:
 
 [source,java]
@@ -689,7 +735,7 @@ the participant state model>>. These annotations are:
 * `@Compensate` -- a method to be executed when the LRA is cancelled
 * `@Complete` -- a method to be executed when the LRA is closed
 * `@Status` -- a method that allow user to state status of the participant with regards 
-to a paricular LRA
+to a particular LRA
 * `@Forget` -- a method to be executed when the LRA allows participant to 
 clear all associated information
 
@@ -698,9 +744,9 @@ methods associated with the JAX-RS resource method or the methods which are not
 bound to JAX-RS. 
 
 [[jaxrs-participant-methods]]
-===== JAX-RS participant methods
+===== JAX-RS methods
 
-The following table presents expectations that are placed on individual participant 
+The following table presents expectations that are placed on individual
 annotations when associated with JAX-RS resource methods:
 
 [[jaxrs-response-table]]
@@ -727,22 +773,50 @@ annotations when associated with JAX-RS resource methods:
 | 200, 404, 412
 | no expectations
 
+| `@AfterLRA`
+| PUT
+| 200
+| no expectations
+
 |=== 
 
-The presented HTTP methods SHOULD be applied at individual JAX-RS participant methods 
+The presented HTTP methods SHOULD be applied at individual JAX-RS methods
 but they are not enforced by the specification. 
 
-If the annotated method returns an unexpected HTTP status code the implementation MAY
-invoke the same method again.
+If the method annotated with `@AfterLRA` returns an unexpected HTTP status
+or never reaches the caller then the implementation MUST invoke the same method again.
 
-Users are allowed to reuse existing JAX-RS endpoints for participant methods definitions. 
+If the annotated method returns an unexpected HTTP status code the implementation MAY
+invoke the same method again with the following caveat: if there is no `@Status` method
+and the implementation receives an unexpected response code from either of the
+`@Compensate` or `@Complete` invocations then it MUST reinvoke the method. [Note that
+this caveat applies to the situation where the response is lost since the caller will
+not see the correct code].
+
+Users are allowed to reuse existing JAX-RS endpoints for participant methods definitions.
 In this case, LRA implementation MUST ensure that invoking these methods outside of the 
 implementation of the LRA specification will not influence any running LRA.
+
+===== Non-JAX-RS afterLRA method
+
+A method annotated with `@AfterLRA` that is not a JAX-RS resource method MUST accept
+two arguments of type URI and LRAStatus, in that order. The first parameter holds
+an LRA context and the second parameter holds the final status of the LRA. If the
+signature does not conform to this requirement then the behaviour of the implementation
+is unspecified.
+
+An example of a valid signature is:
+
+[source,java]
+----
+@AfterLRA
+public void onLRAEnd(URI lraId, LRAStatus status)
+----
 
 [[non-jaxrs-participant-methods]]
 ===== Non-JAX-RS participant methods
 
-When the participant annotations are applied to the non-JAX-RS resource methods they 
+When the participant annotations are applied to the non-JAX-RS resource methods they
 MUST adhere to these predefined signatures:
 
 * *Return type*: 
@@ -1043,8 +1117,8 @@ participant compensation handlers to run:
 
         final CompletableFuture<Response> response = new CompletableFuture<>();
 
-        excecutorService.submit(() -> {
-            // excecute long running business activity finishing with a NOT_FOUND error
+        executorService.submit(() -> {
+            // execute long running business activity finishing with a NOT_FOUND error
             // which causes the LRA to cancel
             response.completeExceptionally(
                     new WebApplicationException(
@@ -1067,8 +1141,8 @@ the JAX-RS `@Suspended` annotation:
     @POST
     public void asyncResponseLRA(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId,
                          final @Suspended AsyncResponse ar) {
-        excecutorService.submit(() -> {
-            // excecute long running business activity and resume when done
+        executorService.submit(() -> {
+            // execute long running business activity and resume when done
             ar.resume(Response.ok().entity(lraId).build());
         });
     }
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java
@@ -19,8 +19,11 @@
  *******************************************************************************/
 package org.eclipse.microprofile.lra.tck;
 
+import org.eclipse.microprofile.lra.annotation.LRAStatus;
 import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
 import org.eclipse.microprofile.lra.tck.participant.api.GenericLRAException;
+import org.eclipse.microprofile.lra.tck.service.LRAMetricService;
+import org.eclipse.microprofile.lra.tck.service.LRAMetricType;
 
 import javax.ws.rs.client.Entity;
 import javax.ws.rs.client.Invocation;
@@ -80,6 +83,42 @@ public class LRAClientOps {
         int status = tryToEnlistWithAnLRA(lra);
 
         return status == Response.Status.NOT_FOUND.getStatusCode() || status == Response.Status.PRECONDITION_FAILED.getStatusCode();
+    }
+
+    /**
+     * TODO if the current PR is acceptable then delete the old isLRAFinished method
+     * (which tests whether an LRA is active by making an attempt to enlist with it
+     * which some spec implementations report as a stack trace WARNING if the LRA is
+     * no longer active).
+     *
+     * @param lra the LRA to test
+     * @param lraMetricService metrics
+     * @param resourceName name of the resource that the metrics parameter applies to
+     * @return whether or not an LRA has finished
+     */
+    boolean isLRAFinished(URI lra, LRAMetricService lraMetricService, String resourceName) {
+        return getLRAEndStatus(lra, lraMetricService, resourceName) != null;
+    }
+
+    /**
+     *
+     * @param lra the LRA whose end status is being queried
+     * @param lraMetricService the metrics service
+     * @param resourceName the name of the resource whose metrics hold the end state of the LRA
+     * @return the end status of the LRA (or null if the LRA has not yet finished)
+     */
+    private LRAStatus getLRAEndStatus(URI lra, LRAMetricService lraMetricService, String resourceName) {
+        if (lraMetricService.getMetric(LRAMetricType.Closed, lra, resourceName) == 1) {
+            return LRAStatus.Closed;
+        } else if (lraMetricService.getMetric(LRAMetricType.FailedToClose, lra, resourceName) == 1) {
+            return LRAStatus.FailedToClose;
+        } else if (lraMetricService.getMetric(LRAMetricType.Cancelled, lra, resourceName) == 1) {
+            return LRAStatus.Cancelled;
+        } else if (lraMetricService.getMetric(LRAMetricType.FailedToCancel, lra, resourceName) == 1) {
+            return LRAStatus.FailedToCancel;
+        }
+
+        return null;
     }
 
     // synchronize access to the connection since it is shared with the LRA background cancellation code
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
@@ -30,6 +30,7 @@ import org.junit.Test;
 import org.junit.runner.RunWith;
 
 import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.client.WebTarget;
 import javax.ws.rs.core.Response;
 
@@ -55,10 +56,10 @@ public class TckParticipantTests extends TckTestBase {
     }
     
     /**
-     * Test verifies that non-JAX-RS @Complete method is invoked according to the 
-     * LRA protocol and that if {@link javax.ws.rs.WebApplicationException} is 
+     * Test verifies that non-JAX-RS @Complete method is invoked according to the
+     * LRA protocol and that if {@link WebApplicationException} is
      * thrown inside of a non-JAX-RS participant method than {@link Response} it 
-     * is carrying is extracted and acted upon accoding to LRA response handling
+     * is carrying is extracted and acted upon according to LRA response handling
      */
     @Test
     public void validWebApplicationExceptionReturnedTest() {
@@ -68,16 +69,23 @@ public class TckParticipantTests extends TckTestBase {
         Response response = resourcePath.request().get();
         URI lraId = URI.create(checkStatusReadAndCloseResponse(Response.Status.OK, response, resourcePath));
 
-        assertEquals("Non JAX-RS @Complete method throwing WebApplicationException shoud have been called", 
-            1, lraMetricService.getMetric(LRAMetricType.COMPLETE, lraId));
+        assertEquals("Non JAX-RS @Complete method throwing WebApplicationException shoud have been called",
+            1, lraMetricService.getMetric(LRAMetricType.Completed, lraId));
         assertEquals("@Compensate method should not have been called as LRA completed succesfully",
-            0, lraMetricService.getMetric(LRAMetricType.COMPENSATE, lraId));
+            0, lraMetricService.getMetric(LRAMetricType.Compensated, lraId));
 
+        /*
+         * Validate that a resource can receive notifications of the final outcome of an LRA using
+         * the {@link org.eclipse.microprofile.lra.annotation.AfterLRA} annotation on
+         * a non JAX-RS method
+         */
+        assertEquals("@AfterLRA method should have been called",
+                1, lraMetricService.getMetric(LRAMetricType.AfterLRA, lraId));
     }
 
     /**
      * This test verifies chained call of non-JAX-RS participant methods. First 
-     * it starts and cancels a new LRA. @Compensate non-JAX-RS method then returns 
+     * it starts and cancels a new LRA. @Compensate non-JAX-RS method then returns
      * {@link org.eclipse.microprofile.lra.annotation.ParticipantStatus#Compensating} 
      * (see {@link ValidLRAParticipant}) indicating that non-JAX-RS @Status method should 
      * be invoked. The test then waits for recovery and then verifies that @Status method is 
@@ -97,16 +105,16 @@ public class TckParticipantTests extends TckTestBase {
         URI lraId = URI.create(checkStatusReadAndCloseResponse(Response.Status.INTERNAL_SERVER_ERROR, response, resourcePath));
         
         assertEquals("Non JAX-RS @Compensate method should have been called",
-            1, lraMetricService.getMetric(LRAMetricType.COMPENSATE, lraId));
+            1, lraMetricService.getMetric(LRAMetricType.Compensated, lraId));
         assertEquals("@Complete method should not have been called as LRA compensated",
-            0, lraMetricService.getMetric(LRAMetricType.COMPLETE, lraId));
+            0, lraMetricService.getMetric(LRAMetricType.Completed, lraId));
 
         replayEndPhase(ValidLRAParticipant.RESOURCE_PATH);
 
         assertEquals("Non JAX-RS @Status method should have been called", 
-            1, lraMetricService.getMetric(LRAMetricType.STATUS, lraId));
+            1, lraMetricService.getMetric(LRAMetricType.Status, lraId));
         assertEquals("Non JAX-RS @Forget method should have been called",
-            1, lraMetricService.getMetric(LRAMetricType.FORGET, lraId));
+            1, lraMetricService.getMetric(LRAMetricType.Forget, lraId));
 
     }
 
@@ -124,9 +132,9 @@ public class TckParticipantTests extends TckTestBase {
         URI lraId = URI.create(checkStatusReadAndCloseResponse(Response.Status.INTERNAL_SERVER_ERROR, response, resourcePath));
 
         assertEquals("Non JAX-RS @Compensate method with CompletionStage<Void> should have been called",
-            1, lraMetricService.getMetric(LRAMetricType.COMPENSATE, lraId));
+            1, lraMetricService.getMetric(LRAMetricType.Compensated, lraId));
         assertEquals("Non JAX-RS @Complete method should have not been called",
-            0, lraMetricService.getMetric(LRAMetricType.COMPLETE, lraId));
+            0, lraMetricService.getMetric(LRAMetricType.Completed, lraId));
     }
 
     /**
@@ -146,13 +154,13 @@ public class TckParticipantTests extends TckTestBase {
         URI lraId = URI.create(checkStatusReadAndCloseResponse(Response.Status.OK, response, resourcePath));
 
         assertEquals("Non JAX-RS @Complete method with CompletionStage<Response> should have been called",
-            1, lraMetricService.getMetric(LRAMetricType.COMPLETE, lraId));
+            1, lraMetricService.getMetric(LRAMetricType.Completed, lraId));
         assertEquals("Non JAX-RS @Compensate method should have not been called",
-            0, lraMetricService.getMetric(LRAMetricType.COMPENSATE, lraId));
+            0, lraMetricService.getMetric(LRAMetricType.Compensated, lraId));
 
         replayEndPhase(ValidLRACSParticipant.ROOT_PATH);
 
         assertEquals("Non JAX-RS @Status method with CompletionStage<ParticipantStatus> should have been called",
-            1, lraMetricService.getMetric(LRAMetricType.STATUS, lraId));
+            1, lraMetricService.getMetric(LRAMetricType.Status, lraId));
     }
 }
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
@@ -20,12 +20,16 @@
 package org.eclipse.microprofile.lra.tck;
 
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
+import static org.eclipse.microprofile.lra.tck.participant.api.AfterLRAListener.AFTER_LRA_LISTENER_WORK;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraController.ACCEPT_WORK;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraController.LRA_CONTROLLER_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraController.TRANSACTIONAL_WORK_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.JOIN_WITH_EXISTING_LRA_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.JOIN_WITH_EXISTING_LRA_PATH2;
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.TCK_PARTICIPANT_RESOURCE_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.AfterLRAListener.AFTER_LRA_LISTENER_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.AfterLRAParticipant.AFTER_LRA_PARTICIPANT_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.AfterLRAParticipant.AFTER_LRA_PARTICIPANT_WORK;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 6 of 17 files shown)


mmusgrov:
I deliberately excluded non JAX-RS methods. This feature is orthogonal to participants. I have just added an extra test to make that explicit.

rdebusscher:
The @AfterEnd is to 'send' the final state of the LRA to the Participant, correct? Otherwise, I'm obvious not seeing the use case/scenario then.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricType.java
```diff
@@ -20,10 +20,16 @@
...
-    COMPENSATE,
-    COMPLETE,
-    STATUS,
-    FORGET,
-    NESTED
+    // LRA statistics
+    Closed, // an LRA that closed
+    FailedToClose, // an LRA that failed to close
+    Cancelled, // an LRA that was cancelled
+    FailedToCancel, // an LRA that failed to cancel
```
xstefank: this is again kind of subjective, but can we keep the same code style for enum values please? I think either make Complete, Compensate,... or CLOSED, FAILED_TO_CLOSE,...

xstefank:
> I deliberately excluded non JAX-RS methods. This feature is orthogonal to participants. I have just added an extra test to make that explicit.

Can you elaborate on this please? I would think of this functionality as of again something that is called by coordination (like complete, compensate, forget, status) not like something to be called by users (like LRA, Leave)

mmusgrov:
> The @AfterEnd is to 'send' the final state of the LRA to the Participant, correct? Otherwise, I'm obvious not seeing the use case/scenario then.

The specification was motivated by the desire to enable applications to organise work in some kind of business process where the completion of some activities triggers the subsequent ones, to quote our specification:

> Different usage patterns for LRAs are possible, for example LRAs may be used sequentially and/or
concurrently, where the termination of one LRA signals the start of some other unit of work within
an application. ...

And an actual, real world, use case is our own TCK.

tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java
```diff
@@ -82,6 +85,42 @@ boolean isLRAFinished(URI lra) {
...
+            return LRAStatus.Closed;
+        } else if (lraMetricService.getMetric(LRAMetricType.FailedToClose, lra, resourceName) == 1) {
+            return LRAStatus.FailedToClose;
+        } else if (lraMetricService.getMetric(LRAMetricType.Cancelled, lra, resourceName) == 1) {
+            return LRAStatus.Cancelled;
+        } else if (lraMetricService.getMetric(LRAMetricType.FailedToCancel, lra, resourceName) == 1) {
+            return LRAStatus.FailedToCancel;
+        }
+
+        return null;
```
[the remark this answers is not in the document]
mmusgrov: The `AfterLRA` notification is only generated when the LRA has entered a terminal state (ie it does not include `Active`, `Closing` or `Cancelling`);

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/AfterEnd.java
```diff
@@ -0,0 +1,41 @@
...
+ * by marking one of the methods with the {@link AfterEnd}
+ * annotation.
+ *
+ * The LRA context is made available to the annotated method
+ * via an HTTP header with the name
+ * {@link LRA#LRA_HTTP_ENDED_CONTEXT_HEADER}
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.METHOD})
+public @interface AfterEnd {
```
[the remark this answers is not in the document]
mmusgrov: okay

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -101,6 +101,12 @@
      */
     String LRA_HTTP_CONTEXT_HEADER = "Long-Running-Action";
 
+    /**
+     * Header name holding the LRA context of an LRA that has finished - to be
+     * used in conjunction with the {@link AfterEnd} annotation
+     */
+    String LRA_HTTP_ENDED_CONTEXT_HEADER = "Long-Running-Action-Ended";
```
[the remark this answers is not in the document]
mmusgrov: A target use case is to trigger  new actions when a previous one has finished. Therefore we should support something like:

```java
    @PUT
    @Path(AFTER_LRA)
    @AfterLRA // this method will be called when the LRA associated with the method activityWithLRA finishes
    @LRA(value = LRA.Type.REQUIRES_NEW)
    public Response afterEnd(
        @HeaderParam(LRA_HTTP_ENDED_CONTEXT_HEADER) URI endedContext,
        @HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI newContext,
        String endStatus) {
```

so both headers are useful here

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -26,6 +26,9 @@
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.JOIN_WITH_EXISTING_LRA_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.JOIN_WITH_EXISTING_LRA_PATH2;
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.TCK_PARTICIPANT_RESOURCE_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.VolatileListener.VOLATILE_LISTENER_PATH;
```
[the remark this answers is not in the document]
mmusgrov: I forgot to git add the new file

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -208,6 +213,52 @@ public void join() throws WebApplicationException {
...
+                1,
+                lraMetricService.getMetric(LRAMetricType.Closed, lra, VolatileParticipant.class.getName()));
+    }
+
+    /**
+     * TCK test to verify that methods annotated with {@link org.eclipse.microprofile.lra.annotation.ws.rs.AfterEnd}
+     * are notified correctly when an LRA terminates
+     */
+    @Test
+    public void testVolatileListener() throws WebApplicationException, InterruptedException {
```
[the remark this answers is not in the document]
mmusgrov: I forgot to git add the new file

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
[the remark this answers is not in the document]
xstefank: got it, thanks. Please resolve this one.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricType.java
mmusgrov: Agreed, the literals are meant to match up with LRAStatus enums

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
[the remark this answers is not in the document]
mmusgrov: You can resolve conversations too

tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java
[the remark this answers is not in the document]
xstefank: thanks for the clarification. This can be resolved -- I cannot resolve PR conversations because I am not a commiter ;)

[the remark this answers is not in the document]
mmusgrov: Ohm that's poor.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java
```diff
@@ -281,31 +281,31 @@ public Response leave(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
         return Response.ok().entity(lraId.toASCIIString()).build();
     }
 
-    // Compensate methods SHOULD use @PUT but to verify that other HTTP methods are also supported use here HTTP GET
+    // Compensated methods SHOULD use @PUT but to verify that other HTTP methods are also supported use here HTTP GET
```
tomjenkinson: I these methods are better describable as "Compensate" methods

xstefank:
@tomjenkinson maybe this helps https://github.com/eclipse/microprofile-lra/pull/174/files#diff-c196906fd09ae1e04e3c9d20d737ac86R67 ?

@mmusgrov I just noticed if endStatus (or the String response body) is supposed to be LRAStatus how does the participant know whether LRA is in FailedToCompensate because of the timeout or because some other participant failed. Wasn't this the original intent?

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
```diff
@@ -0,0 +1,86 @@
...
+    @LRA(value = LRA.Type.REQUIRED, end = false)
+    public Response activityWithLRA(@HeaderParam(LRA_HTTP_RECOVERY_HEADER) URI recoveryId,
+                                    @HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
+        return Response.ok().build();
+    }
+
+    @PUT
+    @Path(AFTER_LRA)
+    @AfterLRA // this method will be called when the LRA associated with the method activityWithLRA finishes
+    public Response afterEnd(@HeaderParam(LRA_HTTP_ENDED_CONTEXT_HEADER) URI lraId, String endStatus) {
```
rdebusscher: Parameter type should be LRAStatus instead of String. This makes it clearer what accepted values there are.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
```diff
@@ -0,0 +1,86 @@
...
+        LRAStatus status = LRAStatus.valueOf(endStatus);
+
+        if (status.equals(LRAStatus.Closed)) {
```
rdebusscher: Comparing with enum should be with ==

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
```diff
@@ -0,0 +1,86 @@
...
+                    AfterLRAListener.class.getName());
+        } else if (status.equals(LRAStatus.FailedToCancel)) {
+            lraMetricService.incrementMetric(LRAMetricType.FailedToCancel,
+                    lraId, AfterLRAListener.class.getName());
+        } else if (status.equals(LRAStatus.FailedToClose)) {
+            lraMetricService.incrementMetric(LRAMetricType.FailedToClose,
+                    lraId, AfterLRAListener.class.getName());
+        }
+
+        return Response.ok().build();
```
rdebusscher: Describe this in the spec and what happens when endpoint is called within an invalid body (not a LRAStatus)

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -208,6 +214,52 @@ public void join() throws WebApplicationException {
...
+                1,
+                lraMetricService.getMetric(LRAMetricType.Closed, lra, AfterLRAParticipant.class.getName()));
+    }
+
+    /**
+     * TCK test to verify that methods annotated with {@link AfterLRA}
+     * are notified correctly when an LRA terminates
+     */
+    @Test
+    public void testAfterLRAListener() throws WebApplicationException, InterruptedException {
```
rdebusscher: What is the difference with testAfterLRAParticipant()?

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAParticipant.java
```diff
@@ -0,0 +1,116 @@
...
+    public Response compensateWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId, String userData) {
+        lraMetricService.incrementMetric(LRAMetricType.Compensated, lraId, AfterLRAParticipant.class.getName());
+
+        return Response.ok().build();
+    }
+
+    @DELETE
+    @Path("/forget")
+    @Forget
+    public Response forgetWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
```
rdebusscher: Will this ever be called since activityWithLRA always succeeds?

mmusgrov:
> Also, I think we should add adoc changes in this PR too, it would help with my earlier question

If the general feeling is that this PR is a useful contribution then I will go ahead and update the asciidoc

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAParticipant.java
```diff
@@ -0,0 +1,116 @@
...
+    public Response completeWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId, String userData) {
+        lraMetricService.incrementMetric(LRAMetricType.Completed, lraId, AfterLRAParticipant.class.getName());
+        return Response.ok().build();
+    }
+
+    @PUT
+    @Path("/compensate")
+    @Compensate
+    public Response compensateWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId, String userData) {
+        lraMetricService.incrementMetric(LRAMetricType.Compensated, lraId, AfterLRAParticipant.class.getName());
```
rdebusscher: @Compensate is required but actually never be called since activityWithLRA always succeeds. Method can be empty with a comment indicating why it is empty and never been called.

mmusgrov:
> @tomjenkinson maybe this helps https://github.com/eclipse/microprofile-lra/pull/174/files#diff-c196906fd09ae1e04e3c9d20d737ac86R67 ?
> 
> @mmusgrov I just noticed, if endStatus (or the String response body) is supposed to be LRAStatus how does the participant know whether LRA is in FailedToCompensate because of the timeout or because some other participant failed. Wasn't this the original intent?

Personally, I think just knowing the end state is sufficient but if people want me to report a more complex data structure in the `EndLRA` notification then I can add/propose one. Is that the general consensus?

rdebusscher:
@xstefank  How is the time-out of the LRA linked to the `FailedToCancel` (as `FailedToCompensate` is a participant status) of the LRA? 

A timed-out resource can also end with `Compensated`.  

We have no idea why we have the LRA status `FailedToCancel` since we just have the return `FailedToCompensate` but no reason why the participant returned this.

xstefank:
> @xstefank How is the time-out of the LRA linked to the `FailedToCancel` (as `FailedToCompensate` is a participant status) of the LRA?

My bad, typing too fast. I am talking about LRA of course, sorry.

I was talking about this conversation - https://groups.google.com/forum/#!searchin/microprofile/LRA%7Csort:date/microprofile/fdrcpKPYn0w/WONAPNDsAgAJ

`Is there a way for the compensation method to find out, if it was called because of a timeout or any other error?`

mmusgrov:
> Additional remarks on top of the missing support for non-JAX-RS.

@rdebusscher I already responded to your question about the non-JAX-RS method:

> The @AfterEnd is to 'send' the final state of the LRA to the Participant, correct? Otherwise, I'm obvious not seeing the use case/scenario then.

The specification was motivated by the desire to enable applications to organise work in some kind of business process where the completion of some activities triggers the subsequent ones, to quote our specification:

> Different usage patterns for LRAs are possible, for example LRAs may be used sequentially and/or
concurrently, where the termination of one LRA signals the start of some other unit of work within
an application. ...

And an actual, real world, use case is our own TCK.

So a typical use case would be:

A target use case is to trigger  new actions when a previous one has finished. Therefore we should support something like:

```java
    @PUT
    @Path(AFTER_LRA)
    @AfterLRA // this method will be called when the LRA associated with the method activityWithLRA finishes
    @LRA(value = LRA.Type.REQUIRES_NEW)
    public Response afterEnd(
        @HeaderParam(LRA_HTTP_ENDED_CONTEXT_HEADER) URI endedContext,
        @HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI newContext,
        String endStatus) {
```

rdebusscher:
@mmusgrov I agree that it can be important to get feedback that the LRA is finished to get something else started.

But by limiting it to JAX-RS you are excluding implementations which do not use JAX-RS as communication to participants. And makes it essentially a layer on top of Narayana.

rdebusscher:
All we need to do, is define in the spec that the following format is allowed for the method since we have already the proxy participant which handles the callbacks for @Compensate and @Complete and can handle it for this call also.

    @AfterLRA
    public void onLRAEnd(URI lraId, LRAStatus status) {

xstefank:
I agree with Rudy. I asked in this PR already why it should be excluded from non-JAX-RS participants - https://github.com/eclipse/microprofile-lra/pull/174#issuecomment-505297539

mmusgrov:
> @mmusgrov I agree that it can be important to get feedback that the LRA is finished to get something else started.
> 
> But by limiting it to JAX-RS you are excluding implementations which do not use JAX-RS as communication to participants. And makes it essentially a layer on top of Narayana.

If you have a use case I will update the PR to include the non JAX-RS case.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
tomjenkinson: +1

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
tomjenkinson: I agree it would be clearer and comparison for enum can be "=="

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
tomjenkinson: good catch, should be in the spec text - 200 does not seem the right return code

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAParticipant.java
tomjenkinson: +1 that would be clearer, maybe it could generate some assertion failed error in case it is ever called in future reuse situation?

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAParticipant.java
tomjenkinson: if it should not be called, some assertion to catch the problem might be useful

xstefank:
> If you have a use case I will update the PR to include the non JAX-RS case and use it in the spec doc as justification.

Any JAX-RS use case with JAX-RS annotations removed on all methods except for the `@LRA` method.

> Re:JAX-RS, personally I think this is not the same same level as participant API (like `@Compensate` 
and `@Complete`) it is closer to the `@LRA` annotation as it relates to notifications for the business logic

The `@AfterLRA` method is called in the same way as other participant methods (from coordinator) so I cannot really see why it should not be present in non JAX-RS case. Cannot we say that even Compensate, Complete are also notifications for the business logic? After all, they are saying that participant should do something with performed business action -- either perform "undo" business action or clean up performed business action which itself can be also business action again (e.g. order confirmation).

rdebusscher:
@mmusgrov Same case as for the non-JAX-RS case

- LRA ends (with whatever final state - completed, compensated, failedtoCompensate, ...)
- System calls the method annotated with @AfterLRA
- Developer/Business can launch any task/statement/call in response to the end of the LRA. (which can be a call to some endpoints which starts again another LRA)

For example

When Completed/cancelled, Call another service which needs to be informed of the success/failure of the Long Running Action

When FailedToCompensate, perform any required cleanup/consolidation/... (since some participants have completed successful but others didn't)

ochaloup:
I agree here with the Rudy's and Martin's point of view. The `@AfterLRA` is about to be called at the participant. Based on the call the participant decides what is should be done as the next steps in its business processing. And it's the participant that has to expose the endpoint for being called by the coordinator. It should be the framework that simplify this exposure whenever it's possible.

xstefank:
@tomjenkinson thank you, that was very helpful. I can understand your view on this now. 
 
> Service A is annoted with `@LRA` and and `@AfterLRA` but not expected to be a participant and calls Service B which is.

So you want to allow non-participant LRA resources (which do not have `@Compensate` method) to be still able to notified about LRA end. However, that would mean that LRA implemetation will have to somehow manage the link between the LRA and the class where the LRA was started/finished if this class had `AfterLRA` method which basically makes it a kind of participant right? It just enlists with `@AfterLRA` and not with `@Compensate`.

mmusgrov:
> @tomjenkinson thank you, that was very helpful. I can understand your view on this now.
> 
> > Service A is annoted with `@LRA` and and `@AfterLRA` but not expected to be a participant and calls Service B which is.
> 
> So you want to allow non-participant LRA resources (which do not have `@Compensate` method) to be still able to notified about LRA end. However, that would mean that LRA implemetation will have to somehow manage the link between the LRA and the class where the LRA was started/finished if this class had `AfterLRA` method which basically makes it a kind of participant right? It just enlists with `@AfterLRA` and not with `@Compensate`.

There is an example in the PR 

> Thanks for the feedback. It might help for me to explain the way I look at this.
> 
> I see this broadly similar to discussion about a synchronization and how it is distinct from a resource. For instance, please compare [Synchronization](https://docs.oracle.com/javaee/7/api/javax/transaction/Synchronization.html) and [XAResource](https://docs.oracle.com/javase/8/docs/api/javax/transaction/xa/XAResource.html) They are different actors in the system.
> 
> I realise my example draws from enterprise Java but please do bear with me and if I can further clarify please allow me to do so. Within a typical enterprise Java setting you would probably expect the participant (say the database or whatever) to be registered with the transaction differently to how an interested party (such as a JPA implementation or whatever).
> 
> Here is another way to look at this. I would expect the following use case to be applicable.
> 
> Service A is annoted with `@LRA` and and `@AfterLRA` but not expected to be a participant and calls Service B which is.
> 
> Note that in order for that to be valid, some of the wording here about suspend would probably need to be edited: https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L73

The PR does already include an example where there is no participant involved (look for the class AfterLRAListener):

```java
@ApplicationScoped
@Path(AfterLRAListener.AFTER_LRA_LISTENER_PATH)
public class AfterLRAListener {
    public static final String AFTER_LRA_LISTENER_PATH = "after-lra-listener";
    public static final String WORK = "work";

    private static final String AFTER_LRA = "/after";

    @Inject
    private LRAMetricService lraMetricService;

    @PUT
    @Path(WORK)
    @LRA(value = LRA.Type.REQUIRED, end = false)
    public Response activityWithLRA(
            @HeaderParam(LRA_HTTP_RECOVERY_HEADER) URI recoveryId,
            @HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
        return Response.ok().build();
    }

    @PUT
    @Path(AFTER_LRA)
    @AfterLRA // this method will be called when the LRA associated with the method activityWithLRA finishes
    public Response afterEnd(@HeaderParam(LRA_HTTP_ENDED_CONTEXT_HEADER) URI lraId, String endStatus) {
        LRAStatus status = LRAStatus.valueOf(endStatus);

        if (status.equals(LRAStatus.Closed)) {
            lraMetricService.incrementMetric(LRAMetricType.Closed, lraId,
                    AfterLRAListener.class.getName());
        } else if (status.equals(LRAStatus.Cancelled)) {
            lraMetricService.incrementMetric(LRAMetricType.Cancelled, lraId,
                    AfterLRAListener.class.getName());
        } else if (status.equals(LRAStatus.FailedToCancel)) {
            lraMetricService.incrementMetric(LRAMetricType.FailedToCancel,
                    lraId, AfterLRAListener.class.getName());
        } else if (status.equals(LRAStatus.FailedToClose)) {
            lraMetricService.incrementMetric(LRAMetricType.FailedToClose,
                    lraId, AfterLRAListener.class.getName());
        }

        return Response.ok().build();
  
```

xstefank:
If LRA is started in one non-participant resource but closed in another non-participant resource, both having `@AfterLRA` methods will both of them get called? 

@mmusgrov @tomjenkinson thank you, I got it now. But sorry I am still not convinced that this cannot be also in non JAX-RS variant. When `@LRA` method is hit, all of the resource methods needs to be checked for annotations -- and this will just mean check not only for Compensate but also for AfterLRA and if there is any hit in non JAX-RS method just expose proxy. In same sense we need to expose proxy now if there is only `@LRA` and non JAX-RS `@Compensate`. Would this be too complicated?

mmusgrov:
> @tomjenkinson thank you, that was very helpful. I can understand your view on this now.
> 
> > Service A is annoted with `@LRA` and and `@AfterLRA` but not expected to be a participant and calls Service B which is.
> 
> So you want to allow non-participant LRA resources (which do not have `@Compensate` method) to be still able to notified about LRA end. However, that would mean that LRA implemetation will have to somehow manage the link between the LRA and the class where the LRA was started/finished if this class had `AfterLRA` method which basically makes it a kind of participant right? It just enlists with `@AfterLRA` and not with `@Compensate`.

Two remarks:

- the issue title deliberately uses the more general term `party` in favour of `participant`
- the PR contains an example resource that is not a participant

mmusgrov:
> If LRA is started in one non-participant resource but closed in another non-participant resource, both having `@AfterLRA` methods will both of them get called?
> 

yes (the annotation applies whenever the same class has a method that is executed in the context of an LRA)

> @mmusgrov @tomjenkinson thank you, I got it now. But sorry I am still not convinced that this cannot be also in non JAX-RS variant. When `@LRA` method is hit, all of the resource methods needs to be checked for annotations -- and this will just mean check not only for Compensate but also for AfterLRA and if there is any hit in non JAX-RS method just expose proxy. In same sense we need to expose proxy now if there is only `@LRA` and non JAX-RS `@Compensate`. Would this be too complicated?

It is easy to implement, but that is not the point (which is use cases). I think the vote is 2-2 with arguments on both sides. @ochaloup would have swung the vote in your favour (3-2) but his argument is based on the fact that it only applies to participants - but I am sure he can rephrase his argument. @tomjenkinson unfortunately I think we may be outnumbered on this one.

rdebusscher:
The method annotated with @AfterLRA doesn't need to be a participant, I can agree on that.  What we **also** need is the non-JAX-RS variant.

We even can support than the use case where we have a @AfterLRA annotated method in a plain CDI bean.
It can be used to collect **any** failed LRA for example.

ochaloup:
The Tom's point about the distinction on the XAResource and Synchronization and comparing it with the LRA approach is great. I haven't realized this. It's good to have a chance to have `@AfterLRA` to be on non-participant class.

I still think of having both variants - JAX-RS and non-JAX-RS - make much sense. The LRA framework just makes the user to define the endpoint or manage creation of the endpoint in the background. The ease of the plain CDI usage sounds pretty good to me.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: testAfterLRAListener validates that the endLRA notification is delivered to resources that are not participants
testAfterLRAParticipant validates that the endLRA notification is delivered to resources that are participants

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
mmusgrov: It would be a spec violation if the implementation calls the endLRA notification with an invalid status.

The spec is not interested in the result (it is similar to @Forget in this respect). In fact a 200 response would tell the caller that the listener successfully received the notification.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAParticipant.java
mmusgrov: The TCK starts LRAs with a timeout. If that timeout is breached then the compensation will be invoked.

The test is just verifying that when the LRA has finished that the notification is sent - it does not matter whether the LRA was closed or cancelled.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
tomjenkinson: 400?

xstefank: We need to mention in the spec text then, that `@AfterLRA` is on the best-effort basis. This, however, seems that some business may depend on this functionality so maybe we should think more about whether it should be best-effort? Please don't take this as something to stall this PR, we can revisit it in 1.x, I just wanted to point it out.

mmusgrov: Then I will include the following text in the spec for this feature:

 "the notification is guaranteed to eventually be delivered and if the callback returns anything other that 200 then the implementation will keep retrying."

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAListener.java
mmusgrov: I changed the tests to a switch statement (which internally would use something similar to ==)

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/AfterLRAParticipant.java
mmusgrov: I removed the method

mmusgrov:
> The method annotated with @AfterLRA doesn't need to be within a participant, I can agree on that. What we **also** need is the non-JAX-RS variant.
> 
> We even can support than the use case where we have a @AfterLRA annotated method in a plain CDI bean.
> It can be used to collect **any** failed LRA for example.

The lifecycle of an LRA is controlled by invoking JAX-RS methods annotated with `@LRA` so the notification is only generated for the classes that contain such methods and only if they were invoked. 

Is that what you mean?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -208,6 +214,52 @@ public void join() throws WebApplicationException {
...
+        URI lra = lraClient.startLRA(null, lraClientId(), lraTimeout(), ChronoUnit.MILLIS);
+        WebTarget resourcePath = tckSuiteTarget.path(AFTER_LRA_LISTENER_PATH).path(WORK);
```
rdebusscher: Within testAfterLRAParticipant() and testAfterLRAListener() a reference to AfterLRAParticipant.WORK is made (although there is also a AfterLRAListener.WORK)
Both values are the same but can cause some confusing, issues in the future.

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
```diff
@@ -0,0 +1,64 @@
...
+ * then the id of the LRA and its final status can be obtained
+ * by ensuring that the anotated method conforms to the
+ * signature:
+ *
+ * {code}
+ * `    public void onLRAEnd(URI lraId, LRAStatus status)
+ * {/code}
+ *
+ * The return type is ignored and the method name is not
+ * significant.
```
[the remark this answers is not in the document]
mmusgrov: Are you talking about the non JAX-RS case.

The reason we guarantee delivery of the AfterLRA notification is to support crash failures. But if the business method crashes so too will the implementation proxy co-located with it and therefore the method will eventually be reinvoked. This should be the only scenario where the method is reinvoked, ie the business method cannot assume that it will be called again if it throws an exception.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -501,6 +501,20 @@ The `end = false` element on the bookTrip method forces the LRA to continue runn
...
+context depending upon the value specified in the `@LRA` annotation.
+The final state of this LRA can be discovered by marking one of the other methods
+in the class with the `@AfterLRA` annotation. When the LRA enters a terminal state
+the method will be passed the id of the LRA together with the `LRAStatus` (see the
+javadoc for the `@AfterLRA` annotation for more information). Note that the final
+states of an LRA is defined by the <<lra-state-model,the LRA state model>> and it
+would be a specification violation if the implementation calls the method when the
+LRA is not in a final state. Further information about method signatures and
+JAX-RS response codes is given below in section <<jaxrs-participant-methods>>.
+
```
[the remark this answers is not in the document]
mmusgrov: I added a paragraph and an example here to address your comment

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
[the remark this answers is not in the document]
rdebusscher: ok

mmusgrov:
There has been a lot of discussion on this issue but from my perspective and I have addressed everybody's issues. @tomjenkinson @rdebusscher @xstefank @ochaloup are there any further problems with this PR that you would like me to look into?

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
```diff
@@ -0,0 +1,64 @@
...
+ * by marking one of the other methods in the class with
+ * the {@link AfterLRA} annotation.
+ *
+ * If the `AfterLRA` method is also a JAX-RS resource method
+ * then the LRA context is made available to the annotated method
+ * via an HTTP header with the name
+ * {@link LRA#LRA_HTTP_ENDED_CONTEXT_HEADER} and the
+ * final status is passed to the method in json format:
+ *
+ * `@Produces(MediaType.APPLICATION_JSON`
```
xstefank: missing ending parentheses

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -727,22 +773,50 @@ annotations when associated with JAX-RS resource methods:
...
 In this case, LRA implementation MUST ensure that invoking these methods outside of the 
 implementation of the LRA specification will not influence any running LRA.
 
+===== Non-JAX-RS afterLRA method
+
+A method annotated with `@AfterLRA` that is not a JAX-RS resource method MUST accept
+two arguments of type URI and LRAStatus, in that order. The first parameter holds
+an LRA context and the second parameter holds the final status of the LRA. If the
+signature does not conform to this requirement then the behaviour of the implementation
+is unspecified.
```
xstefank: This is more a question: we can catch invalid signature during deployment, would it make sense to add another exception for invalid signatures? Something like InvalidAfterLRADefinitionException? I am also ok with currently stated unspecified behavior as it is not participant-specific and thus the LRA protocol does not rely on it.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -68,16 +69,26 @@ public void validWebApplicationExceptionReturnedTest() {
...
 
-        assertEquals("Non JAX-RS @Complete method throwing WebApplicationException shoud have been called", 
-            1, lraMetricService.getMetric(LRAMetricType.COMPLETE, lraId));
+        assertEquals("Non JAX-RS @Complete method throwing WebApplicationException shoud have been called",
+            1, lraMetricService.getMetric(LRAMetricType.Completed, lraId));
         assertEquals("@Compensate method should not have been called as LRA completed succesfully",
-            0, lraMetricService.getMetric(LRAMetricType.COMPENSATE, lraId));
+            0, lraMetricService.getMetric(LRAMetricType.Compensated, lraId));
 
+        assertEquals("@Compensate method should not have been called as LRA completed succesfully",
```
xstefank: this asserts seems to be duplicated previous one on line 74

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -208,6 +215,52 @@ public void join() throws WebApplicationException {
...
+        lraClient.closeLRA(lra);
+
+        // verify that the LRA is now in one of the terminal states
+        assertTrue("testAfterLRAParticipant: LRA did not finish",
+                lraClient.isLRAFinished(lra, lraMetricService, AfterLRAParticipant.class.getName()));
+
+        // verify that the resource was notified of the final state of the LRA
+        assertEquals("testAfterLRAParticipant: end synchronization was not invoked on resource " + resourcePath.getUri(),
+                1,
+                lraMetricService.getMetric(LRAMetricType.Closed, lra, AfterLRAParticipant.class.getName()));
```
xstefank: this test is missing assert that `@AfterLRA` was invoked. Maybe this assert should be checking LRAMetricType.AfterLRA?

Edit: It seems that `@AfterLRA` methods in After LRA listener and participant increment received status metric in their `@AfterLRA` method. But that means that LRAMetricType.Closed should be equal to 2 - once for /complete call and once for /afterLRA call?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -208,6 +215,52 @@ public void join() throws WebApplicationException {
...
+        lraClient.closeLRA(lra);
+
+        // verify that the LRA is now in one of the terminal states
+        assertTrue("testAfterLRAListener: LRA did not finish",
+                lraClient.isLRAFinished(lra, lraMetricService, AfterLRAListener.class.getName()));
+
+        // verify that the resource was notified of the final state of the LRA
+        assertEquals("testAfterLRAListener: end synchronization was not invoked on resource " + resourcePath.getUri(),
+                1,
+                lraMetricService.getMetric(LRAMetricType.Closed, lra, AfterLRAParticipant.class.getName()));
```
xstefank: same comment as above

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
rdebusscher: Unclear how the @Produces is linked to the paragraph above it? Does it need to be @Consumes? Or more complete example?

    @PUT
    @Consumes(MediaType.APPLICATION_JSON)
    @AfterLRA // this method will be called when the LRA associated with the method activityWithLRA finishes
    public Response afterEnd(@HeaderParam(LRA_HTTP_ENDED_CONTEXT_HEADER) URI lraId, LRAStatus status) {

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -101,6 +102,12 @@
      */
     String LRA_HTTP_CONTEXT_HEADER = "Long-Running-Action";
 
+    /**
+     * Header name holding the LRA context of an LRA that has finished - to be
+     * used in conjunction with the {@link AfterLRA} annotation
```
rdebusscher: First sentence of JavaDoc should have .

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -55,8 +56,8 @@ public static WebArchive deployValidParticipant() {
     }
     
     /**
-     * Test verifies that non-JAX-RS @Complete method is invoked according to the 
-     * LRA protocol and that if {@link javax.ws.rs.WebApplicationException} is 
+     * Test verifies that non-JAX-RS @Complete method is invoked according to the
+     * LRA protocol and that if {@link WebApplicationException} is
      * thrown inside of a non-JAX-RS participant method than {@link Response} it 
      * is carrying is extracted and acted upon accoding to LRA response handling
```
rdebusscher: Typo accoding -> according but not really an issue.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
mmusgrov: These asserts were not introduced by the PR

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: The only code execution path that increments the Closed metric is the one in the AfterLRA notification method ie this test validates that the resource was told that the LRA had been closed via the `AfterLRA` notification.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: The only code execution path that increments the Closed metric is the one in the AfterLRA notification method ie this test validates that the resource was told that the LRA had been closed via the `AfterLRA` notification.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
xstefank: this line (77) is indicated as added by GitHub

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
xstefank: my bad, mixing Closed and Completed, please resolve

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
xstefank: my bad, mixing Closed and Completed, please resolve

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
```diff
@@ -0,0 +1,64 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * If a JAX-RS resource method is annotated with
+ * {@link LRA} and is invoked in the context of an LRA then
+ * the resource can ask to be notified when the LRA finishes
+ * by marking one of the other methods in the class with
+ * the {@link AfterLRA} annotation.
+ *
+ * If the `AfterLRA` method is also a JAX-RS resource method
```
ochaloup:
```suggestion
 * If the <code>AfterLRA</code> method is also a JAX-RS resource method
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
```diff
@@ -0,0 +1,64 @@
...
+ * until it receives a `200 OK` status code from the
+ * resource method (which means that the method SHOULD be
+ * idempotent).
+ *
+ * If the `AfterLRA` method is not a JAX-RS resource method
+ * then the id of the LRA and its final status can be obtained
+ * by ensuring that the annotated method conforms to the
+ * signature:
+ *
+ * {code}
```
ochaloup: The whole javadoc for the interface class should use tags `<p>` to separate paragraps and `<code>` should be used instead of `{code}`
```suggestion
 * <code>
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
mmusgrov: Good points. Actually, it just needs to be a string so that we can avoid writing MessageBodyReaders. I will say that the final status is passed to the method as plain text and include an example.

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
```diff
@@ -0,0 +1,64 @@
...
+ * final status is passed to the method in json format:
+ *
+ * `@Produces(MediaType.APPLICATION_JSON`
+ *
+ * The implementation will keep resending the notification
+ * until it receives a `200 OK` status code from the
+ * resource method (which means that the method SHOULD be
+ * idempotent).
+ *
+ * If the `AfterLRA` method is not a JAX-RS resource method
```
ochaloup:
```suggestion
 * If the <code>AfterLRA</code> method is not a JAX-RS resource method
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java
```diff
@@ -0,0 +1,64 @@
...
+ * If the `AfterLRA` method is also a JAX-RS resource method
+ * then the LRA context is made available to the annotated method
+ * via an HTTP header with the name
+ * {@link LRA#LRA_HTTP_ENDED_CONTEXT_HEADER} and the
+ * final status is passed to the method in json format:
+ *
+ * `@Produces(MediaType.APPLICATION_JSON`
+ *
+ * The implementation will keep resending the notification
+ * until it receives a `200 OK` status code from the
```
ochaloup:
```suggestion
 * until it receives a <code>200 OK</code> status code from the
```

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
mmusgrov: No I mean the assert was already there. What changed was the capitalisation of the enum constant. But I will update the text.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
tomjenkinson: @xstefank I wonder who would catch this InvalidAfterLRADefinitionException?

I think unspecified is fine here, either the vendor might choose to log the error or to fail the deployment or indeed perhaps some other option (I am not sure what exactly)

xstefank: @tomjenkinson indeed, it would be runtime exception failing deployment. I am not sure we want to promise that notification will be delivered but runtime reflective invocation will fail. The business logic may depend on this functionality to start a new LRA for instance.

It would be the same case as for [InvalidLRAParticipantDefinitionException](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/participant/InvalidLRAParticipantDefinitionException.java)

tomjenkinson: I wonder if we should remove https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/participant/InvalidLRAParticipantDefinitionException.java for the same reason?

xstefank: I would prefer to fail as soon as possible. Why would we allow invalid signatures to be deployed when we are sure they will fail when they are called for the first time?

tomjenkinson: I am not saying a vendor should not be allowed to choose to fail to deploy, but my question is if we have to define the exception in the spec. I.e. couldn't a vendor just throw a RTE and get the same behaviour. Let's take Narayana as an example in something like Thorntail. If we through an RTE("Invalid definition....") won't weld treat it the same way as if we throw this specialised exception?

xstefank: yes, that would be the same behavior. Specifying that only some RTE should be thrown during deployment time would enough. But what is the issue with a custom exception in the specification? Is it a problem that it is a part of the API? If you prefer, I can raise a PR that will remove it.

tomjenkinson: The issue is that we add an unnecessary maintenace element into the spec without a use case - or in other words, I do not think that the additional exception will be treated any differently to a RTE by any caller? Perhaps there are some cases you can think of where a specialised exception would be expected to be caught and dealt with differently from a RTE and if that is so we can use the same argument here?

xstefank: @tomjenkinson I've raised https://github.com/eclipse/microprofile-lra/issues/180

tomjenkinson: Great - thanks @xstefank !

tomjenkinson: I think we should make this consistent with #180 and expect vendors to fail deployment and remove the wording about behaviour being unspecified.

xstefank:
@tomjenkinson if this gets merged before #182 I will update #182 accordingly :)

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: So can I resolve this conversation?

xstefank: for me, yes

mmusgrov:
The merge includes #182

Issue #173, opened 2019-06-19
Spec clarification about the responsibilities of JAX-RS compensators


mmusgrov:
When a JAX-RS compensation method cannot compensate immediately then it MUST report a 202 Accepted HTTP status code. I thought this was already clear in the spec but section `3.2.6. Eventual compensations` introduced some ambiguity so I would like to add some text to make this participant responsibility crystal clear.

PR #176, opened 2019-06-28
issue173 Clarify how to report in progress compensations

mmusgrov:
#173 

The current section on Eventual Compensations is not clear about participant responsibilities.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
@@ -27,42 +27,137 @@ import java.lang.annotation.Target;
 import java.time.temporal.ChronoUnit;
 
 /**
- * When a resource method executes in the context of an LRA and if the containing
- * class contains a method annotated with `@Compensate` then the resource will
- * be used as a participant for the LRA. If this associated LRA is subsequently
- * cancelled then the method that this annotation is applied to will be invoked
- * (if the annotation is present on more than one method then an arbitrary one
- * will be chosen). The spec makes no guarantees about when it will be invoked,
- * just that is will eventually be called.
+ * <p>
+ * If a resource method executes in the context of an LRA and if the containing
+ * class has a method annotated with <code>&#64;Compensate</code> then this
+ * method will be invoked if the LRA is cancelled. The resource should attempt to
+ * compensate for any actions it performed in the context of the LRA.
+ * If the annotation is present on more than one method then an arbitrary one
+ * will be chosen. The LRA specification makes no guarantees about when
+ * Compensate method will be invoked, just that it will eventually be called.
+ * </p>
  *
- * If the annotated method is a JAX-RS resource method the id of the currently
+ * <p>
+ * If the annotation is applied to a JAX-RS resource method then the id of the currently
  * running LRA can be obtained by inspecting the incoming JAX-RS headers. If
  * this LRA is nested then the parent LRA MUST be present in the header with the name
  * {@link org.eclipse.microprofile.lra.annotation.ws.rs.LRA#LRA_HTTP_PARENT_CONTEXT_HEADER}
- * and value is of type {@link java.net.URI}.
+ * and the header value will be of type {@link java.net.URI}.
+ * </p>
  *
- * If the annotated method is not a JAX-RS resource method the id of the currently
- * running LRA can be obtained by adhering to a predefined method signature as
- * defined in the LRA specification document. Similarly the method may determine
- * whether or not it runs with a nested LRA by providing a parameter to hold the parent id.
+ * <p>
+ * If the annotated method is not a JAX-RS resource method then the id of the currently
+ * running LRA and its parent LRA (if it is nested) can be obtained by adhering to
+ * predefined method signatures as defined in the LRA specification document.
  * For example,
+ * </p>
+ *
  * <pre>
  *     <code>
- *          &#64;Compensate
- *          public void compensate(URI lraId, URI parentId) { ...}
+ *        &#64;Compensate
+ *        public void compensate(URI lraId, URI parentId) { ...}
  *     </code>
  * </pre>
+ *
+ * <p>
  * would be a valid compensation method declaration. If an invalid signature is detected 
  * the implementation of this specification MUST prohibit successful startup of the application
  * (e.g. with a runtime exception).
+ * </p>
  *
- * Note that, according to the state model {@link LRAStatus} once an LRA has been
- * asked to cancel it is no longer possible to join with it as a participant.
- * Therefore in JAX-RS, combining this annotation with an `@LRA` annotation that does not
- * start a new LRA will result in a `412 PreCondition Failed` status code and is
- * not advised. On the other hand, combining it with an `@LRA` annotation that
- * begins a new LRA can in certain use case make sense, but in this case the LRA
+ * <p>
+ * If the participant cannot compensate immediately then it must report that the
+ * compensation is in progress by either returning a future (such as
+ * {@link java.util.concurrent.CompletionStage}) which will eventually report
+ * one of the final states, or a <code>202 Accepted</code> JAX-RS response code or,
+ * in the case of non JAX-RS resource methods, by returning
+ * {@link ParticipantStatus#Compensating} (see the specification
+ * document for more details).
+ * </p>
+ *
+ * <p>
+ * Note that, according to the state model defined by {@link LRAStatus}, it is not possible
+ * to receive compensation notifications after an LRA has been asked to cancel.
+ * Therefore combining this annotation with an <code>&#64;LRA</code> annotation that does not
+ * start a new LRA will result in a <code>412 PreCondition Failed</code> JAX-RS response
+ * code. On the other hand, combining it with an <code>&#64;LRA</code> annotation that
+ * begins a new LRA can in certain uses case make sense, but in this case the LRA
  * that this method is being asked to compensate for will be unavailable.
+ * </p>
+ *
+ * <p>
+ * If the method is a JAX-RS resource method (or is a non JAX-RS method
+ * annotated with <code>&#64;Compensate</code> with return type
+ * <code>javax.ws.rs.core.Response</code>) then the following are the only
+ * valid response codes:
+ * </p>
+ *
+ *<table border="0" cellpadding="3" cellspacing="0"
+ *   summary="Valid JAX-RS compensation response codes">
+ * <caption>
+ *     <span>JAX-RS Compensation Response Codes</span>
+ *     <span>&nbsp;</span>
+ * </caption>
+ * <tr>
+ *   <th scope="col">Code</th>
+ *   <th scope="col">Response Body</th>
+ *   <th scope="col">Meaning</th>
+ * </tr>
+ * <tr>
+ *   <td scope="row">200</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource has successfully compensated</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">202</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource is still attempting compensation</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">404</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource does not know about the LRA</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">500</td>
+ *   <td scope="row">{@link ParticipantStatus} enum value</td>
+ *   <td scope="row"><p>The resource has failed to compensate.
+ *   The payload contains the reason for the failure.
+ *   A participant MUST remember this state until its
+ *   {@link Forget} method is called.</p>
+ *   <p>The actual value is not important but it MUST
+ *   correspond to a valid {@link ParticipantStatus} enum value. For example,
+ *   if compensation was not possible because the resource already
+ *   completed (without being asked to) then a value such as
+ *   {@link ParticipantStatus#Completed} would be appropriate or
+ *   if it was due to a generic failure then
+ *   {@link ParticipantStatus#FailedToCompensate} would be valid.
+ *   </p>
+ *   <p>
+ *   Note that the
+ *   actual state as reported by the {@link Status} method MUST
+ *   be {@link ParticipantStatus#FailedToCompensate}</p></td>
+ * </tr>
+ * </table>
+ *
+ * <p>
+ * If any other code is returned (or, in the 500 case, the body does not
+ * correspond to a valid state) then the implementation SHOULD either keep
+ * retrying or attempt to discover the status by calling the
+ * {@link Status} method if present or a combination of both.
+ * If the implementation stops retrying then it SHOULD log a warning.
+ * An example scenario where the implementation might attempt to invoke the
+ * compensate method twice and the status method is as follows:
+ * </p>
+ *
+ * <ol>
+ * <li>The implementation invokes the compensate method via JAX-RS.</li>
+ * <li>The JAX-RS server returns a 500 code (ie the notification does not reach the participant).</li>
+ * <li>If there is a status method then the implementation uses that to get the current
+ * state of the participant. If the status is Active then the implementation may
+ * infer that the original request never reached the participant so it is safe to
+ * reinvoke the compensate method.</li>
+ * </ol>
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD})
@@ -71,7 +166,7 @@ public @interface Compensate {
      * The period for which the participant will guarantee it will be able
      * to compensate for any work that it performed during the associated LRA.
      * When this period elapses the LRA that it joined becomes eligible for
-     * cancellation. The units are specified in the {@link Compensate#timeUnit()}
+     * cancellation. The units are specified in the {@link #timeUnit()}
      * attribute.
      *
      * A value of zero indicates that it will always be able to compensate.
@@ -82,7 +177,7 @@ public @interface Compensate {
     long timeLimit() default 0;
 
     /**
-     * @return the unit of time that the {@link Compensate#timeLimit()} attribute is
+     * @return the unit of time that the {@link #timeLimit()} attribute is
      * measured in.
      */
     ChronoUnit timeUnit() default ChronoUnit.SECONDS;
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
@@ -27,44 +27,139 @@ import java.lang.annotation.Target;
 import java.time.temporal.ChronoUnit;
 
 /**
- * When a resource method executes in the context of an LRA and if the containing
- * class contains a method annotated with `@Compensate` then the resource will
- * be used as a participant for the LRA. If, in addition, it also contains a method
- * annotated with `@Complete` then the method will be invoked when the associated
- * LRA is later closed. The spec makes no guarantees about when it will be invoked,
- * just that is will eventually be called.
- *
+ * <p>
+ * If a resource method executes in the context of an LRA and if the containing
+ * class has a method annotated with <code>&#64;Complete</code> (as well as method
+ * annotated with <code>&#64;Compensate</code>) then this Complete
+ * method will be invoked if the LRA is closed. The resource should attempt to
+ * perform any clean up activities relating to any actions it performed in the
+ * context of the LRA.
  * If the annotation is present on more than one method then an arbitrary one
- * will be chosen.
+ * will be chosen. The LRA specification makes no guarantees about when the
+ * Complete method will be invoked, just that it will eventually be called.
+ * </p>
  *
- * If the annotated method is a JAX-RS resource method the id of the currently
+ * <p>
+ * If the annotation is applied to a JAX-RS resource method then the id of the currently
  * running LRA can be obtained by inspecting the incoming JAX-RS headers. If
  * this LRA is nested then the parent LRA MUST be present in the header with the name
  * {@link org.eclipse.microprofile.lra.annotation.ws.rs.LRA#LRA_HTTP_PARENT_CONTEXT_HEADER}
- * and value is of type {@link java.net.URI}.
+ * and the header value will be of type {@link java.net.URI}.
+ * </p>
  *
- * If the annotated method is not a JAX-RS resource method the id of the currently
- * running LRA can be obtained by adhering to a predefined method signature as
- * defined in the LRA specification document. Similarly the method may determine
- * whether or not it runs with a nested LRA by providing a parameter to hold the parent id.
+ * <p>
+ * If the annotated method is not a JAX-RS resource method then the id of the currently
+ * running LRA and its parent LRA (if it is nested) can be obtained by adhering to
+ * predefined method signatures as defined in the LRA specification document.
  * For example,
+ * </p>
+ *
  * <pre>
  *     <code>
- *          &#64;Complete
- *          public void complete(URI lraId, URI parentId) { ...}
+ *        &#64;Complete
+ *        public void complete(URI lraId, URI parentId) { ...}
  *     </code>
  * </pre>
- * would be a valid completion method declaration. If an invalid signature is detected 
+ *
+ * <p>
+ * would be a valid completion method declaration. If an invalid signature is detected
  * the implementation of this specification MUST prohibit successful startup of the application
  * (e.g. with a runtime exception).
+ * </p>
  *
- * Note that, according to the state model {@link LRAStatus} once an LRA has been
- * asked to close it is no longer possible to join with it as a participant.
- * Therefore in JAX-RS, combining this annotation with an `@LRA` annotation that does not
- * start a new LRA will result in a `412 PreCondition Failed` status code and is
- * not advised. On the other hand, combining it with an `@LRA` annotation that
- * begins a new LRA can in certain use case make sense, but in this case the LRA
+ * <p>
+ * If the participant cannot complete immediately then it must report that
+ * completion is in progress by either returning a future (such as
+ *  * {@link java.util.concurrent.CompletionStage}) which will eventually report
+ * one of the final states, or a <code>202 Accepted</code> JAX-RS response code or,
+ * in the case of non JAX-RS resource methods, by returning
+ * {@link ParticipantStatus#Completing} (see the specification
+ * document for more details).
+ * </p>
+ *
+ * <p>
+ * Note that, according to the state model defined by {@link LRAStatus}, it is not possible
+ * to receive completion notifications after an LRA has been asked to close.
+ * Therefore combining this annotation with an <code>&#64;LRA</code> annotation that does not
+ * start a new LRA will result in a <code>412 PreCondition Failed</code> JAX-RS response
+ * code. On the other hand, combining it with an <code>&#64;LRA</code> annotation that
+ * begins a new LRA can in certain use cases make sense, but in this case the LRA
  * that this method is being asked to complete for will be unavailable.
+ * </p>
+ *
+ * <p>
+ * If the method is a JAX-RS resource method (or is a non JAX-RS method
+ * annotated with <code>&#64;Complete</code> with return type
+ * <code>javax.ws.rs.core.Response</code>) then the following are the only
+ * valid response codes:
+ * </p>
+ *
+ *<table border="0" cellpadding="3" cellspacing="0"
+ *   summary="Valid JAX-RS completion response codes">
+ * <caption><span>JAX-RS Completion Response Codes</span><span>&nbsp;</span></caption>
+ * <tr>
+ *   <th scope="col">Code</th>
+ *   <th scope="col">Response Body</th>
+ *   <th scope="col">Meaning</th>
+ * </tr>
+ * <tr>
+ *   <td scope="row">200</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource has successfully completed</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">202</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource is still attempting completion</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">404</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource does not know about the LRA</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">500</td>
+ *   <td scope="row">{@link ParticipantStatus} enum value</td>
+ *   <td scope="row"><p>The resource has failed to complete.
+ *   The payload contains the reason for the failure.
+ *   A participant MUST remember this state until its
+ *   {@link Forget} method is called.</p>
+ *   <p>The actual value is not important but it MUST
+ *   correspond to a valid {@link ParticipantStatus} enum value. For example,
+ *   if completion was not possible because the resource already
+ *   compensated (without being asked to) then a value such as
+ *   {@link ParticipantStatus#Compensated} would be appropriate or
+ *   if it was due to a generic failure then
+ *   {@link ParticipantStatus#FailedToComplete} would be valid.
+ *   If the response body does not contain a valid status then the
+ *   implementation MUST either reinvoke the method or discover the
+ *   status using the {@link Status} annotation if present.
+ *   </p>
+ *   <p>
+ *   Note that the
+ *   actual state as reported by the {@link Status} method MUST
+ *   be {@link ParticipantStatus#FailedToComplete}</p></td>
+ * </tr>
+ * </table>
+ *
+ * <p>
+ * If any other code is returned (or, in the 500 case, the body does not
+ * correspond to a valid state)  then the implementation SHOULD either keep
+ * retrying or attempt to discover the status by calling the
+ * {@link Status} method if present or a combination of both.
+ * If the implementation stops retrying then it SHOULD log a warning.
+ * An example scenario where the implementation might attempt to invoke the
+ * complete method twice and the status method is as follows:
+ * </p>
+ *
+ * <ol>
+ * <li>The implementation invokes the complete method via JAX-RS.</li>
+ * <li>The JAX-RS server returns a 500 code (ie the notification does not reach the participant).</li>
+ * <li>If there is a status method then the implementation uses that to get the current
+ * state of the participant. If the status is Active then the implementation may
+ * infer that the original request never reached the participant so it is safe to
+ * reinvoke the complete method.</li>
+ * </ol>
  */
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD})
@@ -73,7 +168,7 @@ public @interface Complete {
      * The period for which the participant will guarantee it will be able
      * to complete for any work that it performed during the associated LRA.
      * When this period elapses the LRA that it joined becomes eligible for
-     * cancellation. The units are specified in the {@link Complete#timeUnit()}
+     * cancellation. The units are specified in the {@link #timeUnit()}
      * attribute.
      *
      * A value of zero indicates that it will always be able to complete.
@@ -84,7 +179,7 @@ public @interface Complete {
     long timeLimit() default 0;
 
     /**
-     * @return the unit of time that the {@link Complete#timeLimit()} attribute is
+     * @return the unit of time that the {@link #timeLimit()} attribute is
      * measured in.
      */
     ChronoUnit timeUnit() default ChronoUnit.SECONDS;
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Forget.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Forget.java
@@ -28,46 +28,92 @@ import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
 /**
+ * <p>
  * If a participant is unable to complete or compensate immediately
  * (ie it has indicated that the request has been accepted and is
  * in progress) or because of a failure (ie will never be able to finish)
  * then it must remember the fact (by reporting it when asked for its'
- * {@link Status})) until explicitly told that it can clean
- * up using this <em>@Forget</em> annotation.
+ * {@link Status}) until explicitly told that it can clean
+ * up using this <code>&#64;Forget</code> annotation.
+ * </p>
  *
+ * <p>
  * A similar remark applies if the participant was enlisted in a
- * nested LRA {@link LRA#value()}. Actions performed in the context
+ * nested LRA {@link LRA.Type#NESTED}. Actions performed in the context
  * of a nested LRA must remain compensatable until the participant
- * is explicitly told it can clean up using this <em>@Forget</em>
+ * is explicitly told it can clean up using this <code>&#64;Forget</code>
  * annotation.
+ * </p>
  *
+ * <p>
  * If the annotated method is a JAX-RS resource method the id of the currently
  * running LRA can be obtained by inspecting the incoming JAX-RS headers. If
  * this LRA is nested then the parent LRA MUST be present in the header with the name
- * {@link org.eclipse.microprofile.lra.annotation.ws.rs.LRA#LRA_HTTP_PARENT_CONTEXT_HEADER}
+ * {@link LRA#LRA_HTTP_PARENT_CONTEXT_HEADER}
  * and value is of type {@link java.net.URI}.
+ * </p>
  *
+ * <p>
  * If the annotated method is not a JAX-RS resource method the id of the currently
  * running LRA can be obtained by adhering to a predefined method signature as
  * defined in the LRA specification document. Similarly the method may determine
  * whether or not it runs with a nested LRA by providing a parameter to hold the parent id.
  * For example,
+ * </p>
+ *
  * <pre>
  *     <code>
  *          &#64;Forget
  *          public void forget(URI lraId, URI parentId) { ...}
  *     </code>
  * </pre>
+ *
+ * <p>
  * would be a valid forget method declaration. If an invalid signature is detected 
  * the implementation of this specification MUST prohibit successful startup of the application
  * (e.g. with a runtime exception).
+ * </p>
  *
- *
+ * <p>
  * Since the participant generally needs to know the id of the LRA in order
  * to clean up there is generally no benefit to combining this annotation
- * with the `@LRA` annotation (though it is not prohibited).
+ * with the <code>&#64;LRA</code> annotation (though it is not prohibited).
+ * </p>
  *
+ * <p>
  * Related information is provided in the javadoc for the {@link Status}
+ * </p>
+ * <p>
+ * If the method is a JAX-RS resource method (or is a non JAX-RS method
+ * annotated with <code>&#64;Forget</code> with return type
+ * <code>javax.ws.rs.core.Response</code>) then the following are the only
+ * valid response codes:
+ * </p>
+ *
+ *<table border="0" cellpadding="3" cellspacing="0"
+ *   summary="Valid JAX-RS response codes for Forget methods">
+ * <caption><span>JAX-RS Response Codes For Forget Methods</span><span>&nbsp;</span></caption>
+ * <tr>
+ *   <th scope="col">Code</th>
+ *   <th scope="col">Response Body</th>
+ *   <th scope="col">Meaning</th>
+ * </tr>
+ * <tr>
+ *   <td scope="row">200</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The participant may have removed all knowledge of the LRA</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">404</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The method does not know about the LRA</td>
+ * </tr>
+ * </table>
+ *
+ * <p>
+ * If any other code is returned then the implementation SHOULD keep retrying.
+ * If it stops retrying a warning message SHOULD be logged.
+ * </p>
  */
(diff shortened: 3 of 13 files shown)


api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -56,6 +56,11 @@
  * the {@link org.eclipse.microprofile.lra.participant.InvalidLRAParticipantDefinitionException} 
  * will be thrown during the application startup.
  *
+ * If the participant cannot compensate immediately then is must report that the
```
xstefank:
```suggestion
 * If the participant cannot compensate immediately then it must report that the
```

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
```diff
@@ -58,6 +58,11 @@
  * the {@link org.eclipse.microprofile.lra.participant.InvalidLRAParticipantDefinitionException} 
  * will be thrown during the application startup.
  *
+ * If the participant cannot complete immediately then is must report that the
```
xstefank: then it must report

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -816,17 +816,20 @@ exposed method from unauthorized access. The specific security details are not s
...
-- return `ParticipantStatus.Compensating`
-- return a `202 Accepted` HTTP status code
-- mark the method as a JAX-RS asynchronous method (using the
-  `javax.ws.rs.container.Suspended` annotation).
+- Return a CompletionStage or mark the method as asynchronous
+  (using the `javax.ws.rs.container.Suspended` annotation). The future must
+  deliver the final status when the stage completes. Please refer to the section
+  <<reactive-support,about reactive support>> for more details.
+- A JAX-RS method can return a `202 Accepted` HTTP status code
+- A non JAX-RS method can return `ParticipantStatus.Compensating`.
```
xstefank: This seems to be restricting usage of ParticipantStatus to non JAX-Rs methods? How can the participant indicate FailedToCompensate?

mmusgrov: The paragraph begins with:

> If a resource cannot perform a compensation activity immediately then ...

So if it wants to report FailedToCompensate then it means the activity has finished. In other  words this text only applies to compensation activities that are in progress.

mmusgrov: This text is in a section called "Eventual compensations" and the paragraph begins with the text:

>  If a resource cannot perform a compensation activity immediately then ...

it can return Compensating. If it has FailedToCompensate (one of the final states) then then this text is not applicable - ie the text only applies to compensations that are in progress.

xstefank: Exactly, so 

```java
return Response.ok(ParticipantStatus.FailedToCompensate).build()
```
 is allowed but 

```java
return Response.ok(ParticipantStatus.Compensating).build
``` 
isn't. I think this will be confusing for our users.

mmusgrov: LRA is JAX-RS based and therefore should conform to the expected REST norms and this will be particularly true when we add HTTP based interop in milestone 1.x.

When a participant is asked to compensate it should attempt to transition the resource into a final stated (Compensated or FailedToCompenstate). If the resource cannot make the transition immediately then the only reasonable REST based status code is 202:

> HTTP Status 202 indicates that request has been accepted for processing, but the processing has not been completed.

If the resource, or any intermediary, returns another response code then the caller cannot infer that the request is in progress.

xstefank: Why the caller cannot read the request body? It must read it for final states to distinguish between Compensated and FailedToCompensate. I understand your reasoning, but we should decide then if we want to follow the REST-based approach and allow JAX-RS participants to provide their state with HTTP statuses then we should follow it for everything. That means that HTTP OK "ParticipantStatus.FailedToCompensate" shouldn't be allowed as the participant has failed and some appropriate HTTP status (4xx/5xx) should be returned instead and it should not matter what the response body is (same as for 202 and Compensating case). This inconsistency as described in my previous comment will be confusing.

tomjenkinson: I think the crux is whether FailedToComplete is a succesful outcome in our protocol: https://tools.ietf.org/html/rfc7231#section-6.3.1

As an aside, I definitely think a 4xx would be wrong as it would indicate a problem with the client call

tomjenkinson: Also, reading 202 (https://tools.ietf.org/html/rfc7231#section-6.3.3) I think it should probably be 202 and a body containing `ParticipantStatus.Compensating` (or whatever)

xstefank: @tomjenkinson yes, you are right, sorry 5xx would be the right choice here.

My 2c: I would say all ParticipantStatus-es are valid outcomes because a participant is knowingly supplying the response. 200 is an indication of a successful REST request for compensate method, not to the LRA protocol.

mmusgrov: 202 with a body of  `Compenstaing` to hold the status is fine (although superfluous).

Reaching an end state of `FailedToCompensate`, `Compensated` or `Completed` are all valid states for the resource to end with and therefore 200 is fine for those (and the caller can determine which state by reading the body).

As to responding with 200 for the state `Compensating`: the client is not asking the resource to move to some intermediate state, it wants the resource to move to a final state. If the request was "please start compensating" then 200 might be appropriate but the request is "please compensate" and responding with 200 would be wrong if it cannot immediately put the resource into a final state.

mmusgrov: > 202 with a body of `Compenstaing` to hold the status is fine (although superfluous).

Actually the body does make sense if the resource is already trying to complete when asked to compensate!

mmusgrov: > Why the caller cannot read the request body? It must read it for final states to distinguish between Compensated and FailedToCompensate. I understand your reasoning, but we should decide then if we want to follow the REST-based approach and allow JAX-RS participants to provide their state with HTTP statuses then we should follow it for everything. That means that HTTP OK "ParticipantStatus.FailedToCompensate" shouldn't be allowed as the participant has failed and some appropriate HTTP status (4xx/5xx) should be returned instead and it should not matter what the response body is (same as for 202 and Compensating case). This inconsistency as described in my previous comment will be confusing.

So if the outcome is anything other than `Compensated` or `Compensating` would returning `409 Conflict` with payload the actual status resolve your issue.

@tomjenkinson WDYT does section https://tools.ietf.org/html/rfc7231#section-6.5.8 apply here?

mmusgrov: So @xstefank,  for `@Compensate` notifications we would have:

> 200 if the status is `Compensated`
> 202 if the status is `Compensating`
> 409 if the status is `Active`, `Completed`, `FailedToCompensate` or `FailedToCompelete` (with the payload containing the actual status)

and similarly for response to the `@Complete` notification.

xstefank: > So if the outcome is anything other than `Compensated` or `Compensating` would returning `409 Conflict` with payload the actual status resolve your issue.

Not entirely. I am thinking we should either allow HTTP 200 with ParticipantStatus for all states OR don't allow it at all and allow only HTTP status codes for instance 200 == Compensated/Completed, 202 == Compensating/Completing, 5xx == FailedTo*. 

I actually don't see any issue with your original proposal from the sandbox where accepted is declared as 202 OR 200 with ParticipantStatus.Compensating/Completing OR @Suspended. This would only add 202 and @Suspended on top of what is already expected I think.

mmusgrov: So we have reached an impasse. `202 Accepted` should be the only valid response code to indicate that the request to compensate is in progress.

tomjenkinson: I don't think `Active` is valid at all for the compensator asked to `@compensate` the rest of the codes I liked WRT https://tools.ietf.org/html/rfc7231#section-6.5.8

xstefank: 2> So we have reached an impasse. `202 Accepted` should be the only valid response code to indicate that the request to compensate is in progress.

@rdebusscher @ochaloup WDYT?

rdebusscher: Not very familiar with asynchronous REST, but from what I know and read, when using @Suspended the method should return void.

The CompletionStage/Future is then used on the client side when such an async REST is called.

JAX-RS case
-------------

- return status 202 (if this is required and not handled out of the box by JAX_RS if we use @Suspended)
- AsyncResponse.resume(ParticipantStatus.Compensated) or AsyncResponse.resume(ParticipantStatus.FailedToCompensated) are allowed

non JAX-RS case
-----------------

- Method returns a CompletionStage
- returned value should be _ParticipantStatus.Compensated_ or _ParticipantStatus.FailedToCompensated_


The same logic applies for the Complete scenario with the appropriate ParticipantStatus.

ochaloup: I don't think the `202 Accepted` is the only valid response. I can see the perspective  that the `Compensating` response could be rated as a success. This response is used by a user - it's not handled by framework - so from "API" design perspective it's good to take into account how consistent is the use of the return values. If the proposed rules are comprehensible.

Nevertheless, I think differentiating the response by the HTTP codes is what the REST api should strive for.

I understand `ParticipantStatus.FailedToCompensate` changes the the state to a final state but it's an error state. I don't think the `Response.ok(ParticipantStatus.FailedToCompensate).build()` is valid (if we talk about HTTP error codes to be used). If it's valid then the `Response.ok(ParticipantStatus.Compensating).build()` should be considered to be valid as well.
From that perspective using http status `409` sounds good to me.

mmusgrov: From what I am reading the views are:

- @tomjenkinson is okay with 202 for compensating, 200 for compensated and 409 for failure states. For terminal states the body contains the status. For non-terminal states we have a choice: the status URL can be placed in the body or location header (we had this option when we were in the sandbox) or we leave the body empty or we include the status..

- @rdebusscher is okay with 202 for `Compensating`. Use of `@Suspended` is not an issue since a synchronous client is blocked until the server async method completes (ie it will only return when a final state is reached). And an async client will gather the completed result when the server makes it available so again it should never see an intermediate state.

- @ochaloup I am not sure what you want, you seem to be arguing a case for all options, ie you are seeing pros and cons for each alternative.

- @xstefank is okay with 202 for `Compensating`, 200 for `Compensated` and 5xx for the failure states. He is also okay with 200 + the status in the body for all states. But he did add a third alternative of supporting two ways to report `Compensating` depending upon whether the body is present or not.

- @mmusgrov JAX-RS is for REST based applications and the recommendation for reporting asynchronous/in-progress operations is to use the code `202 Accepted`. The status code returned by a participant that has reached a final state is a matter of taste between 200 or 409. 409 means "The request could not be completed due to a conflict with the current state of the target resource" and that matches this situation (the client asks for state Compensated but the resource is in a state where that is not possible - ie there is a conflict between what the client is asking for and what the resource state is).

So from the above feedback the only set of codes that everyone has some degree of agreement with is:

> 202 for compensating, 200 for compensated and either 200 or 409 or 5xx for failure states

Can we agree on 200 or 409 for failure states (using my argument in the previous paragraph for 409). I do not see a case for 5xx codes (which equate to s/w or h/w or configuration failures).

mmusgrov:
There has been a lot of discussion on this issue but from my perspective and I have addressed everybody's issues. @tomjenkinson @rdebusscher @xstefank @ochaloup are there any further problems with this PR that you would like me to look into?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: @mmusgrov ok, so let's go with 200 Compensated, 202 Compensating and 409/5xx(user exceptions) FailedToCompensate and ignore the body of the response. This is a clear distinction.

rdebusscher: Not sure about 409

>This code is used in situations where the user might be able to resolve the conflict and resubmit the request.

But if everyone else agree, I'm fine with it

xstefank: It doesn't matter to me what will be the error code but it just must be unique. @rdebusscher do you have something else in mind, please?

ochaloup: @mmusgrov yes, I was not able to clearly decide as the cons and pros are quite in balance.
Distinction based on the error codes sounds fine to me.

tomjenkinson: 200 = Reached the expected state
202 = In progress
But as to whether it is 200 or 500 (not 5xx, looking at the list of these the only one I could argue for would be [500](https://tools.ietf.org/html/rfc2616#section-10.5.1)) for FailedTo<> depends on your point of view but I think 500

tomjenkinson: I was thinking that 200 would be correct as we do ask the participant to `compensate` and valid outcome states are either `compensated` or `failedtocompensate` and these should be stable outcomes until `forget` is called IIUC.

However then I think I found the definitive position that it should be 500 due to the wording "any other exception" in 3.2.5 in on https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc.

Therefore, I wonder if we can agree on 200, 202 and 500?

mmusgrov: > Not sure about 409
> 
> > This code is used in situations where the user might be able to resolve the conflict and resubmit the request.
> 
> But if everyone else agree, I'm fine with it

It does say "might". It also says the payload SHOULD tell the caller what to do.

In our case the coordinator would read the payload, discover the reason for the conflict, and see that there is no point in resubmitting the request.

I am still okay with 200 or 409 for the failure states.

mmusgrov: The IETF description of 409 does have the nice feature of saying the server SHOULD include enough information to enable the caller to recognise the reason for the conflict. And our spec, as it stands, already provides this information by returning `FailedToCompensate` in the payload.

But if people think 409 is wrong then I am inclined to agree with Tom even though the section he cites refers to non JAX-RS compensation methods so JAX-RS codes are not so applicable. 500 does not provide any reason for the failure and expecting a semantically meaningful payload is probably wrong. Which brings me back to 200. This is fun.

tomjenkinson: I was thinking that the reason it was similar is the non-JAX-RS version reports FailedToComplete as an _exception_ rather than a valid outcome:

In other words rather than having something like:

```public ParticipantCompensateStatus compensate(....) {}```

We went for: 

```public void compensate() {
// If I can't do this
throw new RuntimeException("I failed to compensate");
}```

I did like your point about [409](https://tools.ietf.org/html/rfc2616#section-10.4.10) and returning enough information to recongize the source of the conflict but then after reading some more I am not sure if we would be fulfilling this part:
> "This code is only allowed in situations where it is expected that the user might be able to resolve the conflict and resubmit the request" 
as I don't know that the user (LRA coordinator) can resolve the conflict and resubmit the request unless we think that by calling @Forget on the resource and then calling @Compensate and getting a 404 (which allows the user - LRA coordinator - know things are OK) is an acceptable way to "resolve the conflict" - perhaps it is...

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: > I did like your point about [409](https://tools.ietf.org/html/rfc2616#section-10.4.10) and returning enough information to recongize the source of the conflict but then after reading some more I am not sure if we would be fulfilling this part:
> 
> > "This code is only allowed in situations where it is expected that the user might be able to resolve the conflict and resubmit the request"
> 
> as I don't know that the user (LRA coordinator) can resolve the conflict and resubmit the request unless we think that by calling @forget on the resource and then calling @compensate and getting a 404 (which allows the user - LRA coordinator - know things are OK) is an acceptable way to "resolve the conflict" - perhaps it is...

I addressed that concern in my response to Rudy, [see comment](https://github.com/eclipse/microprofile-lra/pull/176#discussion_r301199890)

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: But I don't think we will be able to reach any consensus on this so would taking a poll help to make progress. The options for reporting a failure to compensate are as follows:

1. 200 + body of `FailedToCompensate`
2. 409 + body of `FailedToCompensate`
3. 500 + body of `FailedToCompensate`
4. ??? (if you want to vote for something else)

The requirement for a payload in option 3 is to handle the case where the response was returned by an intermediary.

If the body does not indicate the reason then the implementation MUST retry the request or use the `Status` URI.

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: And if voting is a viable resolution please add a +1 or -1 for each option so that we can pick a winner.

Comment (tomjenkinson) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: > 
> 
> > I did like your point about [409](https://tools.ietf.org/html/rfc2616#section-10.4.10) and returning enough information to recongize the source of the conflict but then after reading some more I am not sure if we would be fulfilling this part:
> > > "This code is only allowed in situations where it is expected that the user might be able to resolve the conflict and resubmit the request"
> > 
> > 
> > as I don't know that the user (LRA coordinator) can resolve the conflict and resubmit the request unless we think that by calling @forget on the resource and then calling @compensate and getting a 404 (which allows the user - LRA coordinator - know things are OK) is an acceptable way to "resolve the conflict" - perhaps it is...
> 
> I addressed that concern in my response to Rudy, [see this comment](https://github.com/eclipse/microprofile-lra/pull/176#discussion_r301199890)

That comment didn't address the "only allowed" element, just the "might" and "resubmit the request" was not addressed either. But I think my answer proposed sequence does address it more concretely as they client can resubmit (after calling forget), obtain the 404 and treat that differently.

Anyway, I do agree it is time for a poll.

I vote:

1. +0
2. +1
3. +1

Comment (tomjenkinson) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: /cc @rdebusscher @xstefank @ochaloup please can you all provide a vote on the four options?

Comment (tomjenkinson) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: > 
> 
> And if voting is a viable resolution please add a +1 (yes) or -1 (no) or 0 (abstain) for each option so that we can pick a winner. So my votes are:
> 
>     1. +1
> 
>     2. +1
> 
>     3. 0

I forget:
4. 0

Comment (ochaloup) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: 1. -1
2. +1
3. +1

Comment (rdebusscher) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: 1. +1 (when developer explicitly wants to set the FailedToCompensate status)
2. -1
3. +1 (When it failed due to an error)
4. -1

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: 1. -1
2. -1
3. +1

Please let's use either HTTP codes **or** ParticipantStatus response body. If we are going to mix them, we should also allow 200 Compensating which was the original issue. So I don't really have an issue with whether it is 409 or 500, but it must be only one of them and the body is ignored. @rdebusscher developer can also explicitly set 500 HTTP status code, wouldn't that be the same as setting FailedToCompensate response body?

Comment (rdebusscher) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: @xstefank Yes, but I prefer

> public ParticipantStatus doCompensate(...)

over

> public Response doCompensate(...)

or

> throw new WebApplicationException(500)

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: I was thinking we are talking only JAX-RS participants here.

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: I would say option 3 "500 + body of FailedToCompensate" is the winner. I reached that conclusion as follows:

Counting votes gives the result:

1. 0
2. 3
3. 4
4. -1

We nearly got an unambiguous winner with option 3: "500 + body of FailedToCompensate" but Martin has muddied the waters somewhat by voting for a fourth option since he added a caveat after his vote which means he is actually voting for -1 on questions 1, 2 and 3 and +1 on option 4 which would produce the following results:

1. 0
2. 2
3. 3
4. 0

Even with this interpretation of the voting, option 3 is still the winner - can we agree to go with that then?

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: @xstefank 

> The requirement for a payload in option 3 is to handle the case where the response was returned by an intermediary

ie voting for option 3 requires a payload. Also this vote is only for reporting failure states.

All of the other states can be reported without a payload.

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: @mmusgrov sorry for the misunderstanding. I am standing by 500 even if I don't agree with the necessity of payload. But I was overvoted so let's see how this plays out.

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: I am interested to know why you think a payload isn't required for the 500 case since if we can avoid it then the requirements on JAX-RS participants would be much cleaner.

But if we do agree on option 3 then when a JAX-RS participant is asked to compensate it MUST respond, depending upon its state, with:

State  HTTP Code Payload
-----    ---------------- -------

Compensated 200 Empty
Compensating 202 Empty
FailedToCompensate 500 "FailedToCompensate"

(400 is valid if the participant does not exist on the target server).
In the failure case, if the body is not FailedToCompensate the implementation MUST retry until it has the final state.
Once it has the final failure state it MUST call `Forget` on the participant.

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: > In the failure case, if the body is not FailedToCompensate the implementation MUST retry until it has the final state.

What about `@Status` method? Is it respected in this case?

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: Question: In the FailedToCompensate state do we want to include in the payload an indication of why it failed (using one of the ParticipantStatus enum values: Completed, Completing, FailedToComplete). Is that something that would ever be of interest? It could be used to enhance the `@AfterLRA` notification in milestone 1.x

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: > > In the failure case, if the body is not FailedToCompensate the implementation MUST retry until it has the final state.
> 
> What about `@Status` method? Is it respected in this case?

Yes it is, sorry for the ambiguity. I will make that clear when I update the PR.

But can you indicate how we can avoid a payload when the participant is reporting `FailedToCompensate`. I am concerned that the HTTP server could return 500 even though the participant is just fine.

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: Personally, I would prefer to stay with just payload of ParticipantStatus values and 200 HTTP status code. But since we've already decided that we will map the LRA protocol on top of HTTP status codes then we should have enough information in the status meaning - 200 Compensated, 202 Compensating, 500 FailedToCompensate. For 500 it doesn't matter whether the user explicitly passed 500: `return Response.status(Response.Status.INTERNAL_SERVER_ERROR).build();` or if some error was thrown in the compensate method and framework returned 500 (or similar).

I can see where we got confused, I am not expecting that we will call the compensate again when there is failure on the participant side (5xx). In that case, some distinction is necessary.

Comment (tomjenkinson) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: I would be concerned about relying purely on status code 500. Is it possible that the /compensate method executes fine and returns a 200 but some intermediary server (proxy?) fails and returns the 500?

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: That is true @tomjenkinson. In that case wouldn't the response body be overridden with the new error?

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: I thought we were investigating ways to avoid having to look at the response body in the 500 case. If we could do that then that would be great but there is also the case where the request is intercepted by some processing entity before it reaches the participant. A JAX-RS server for example has an internal error that means it cannot deliver the message to the resource and returns 500. In this case the participant is still in the active state (not FailedToCompensate).

Comment (tomjenkinson) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: That would be my concern. I think we have to check it is 500 and the body contains specific content to prevent the situation. I still think it is valid to consider 500 as a reasonable response code even if you need to look in the body on the client side given the definition of 5xx includes "Except when responding to a HEAD request, the
   server SHOULD include an entity containing an explanation of the
   error situation, and whether it is a temporary or permanent
   condition."

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: When we have resolved this current thread I will prepare an update to the PR [according to the voting results.](https://github.com/eclipse/microprofile-lra/pull/176#discussion_r301567046)

Comment (tomjenkinson) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: I am acking the results are OK with me

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: @mmusgrov agreed. If we are moving to HTTP status codes, we will have to have the response body for 500. 

So the final outcome would be:

200 - Compensated/Completed
202 - Compensating/Completing
500 + FailedToCompensate, 500 + FailedToComplete - FailedToCompensate/FailedToComplete
500 (no response  body) - Compensating/Completing

------------------------------------------------------------------------------------------------------------------------
This is just to document my view, feel free to ignore.

I still think that stating participant status through HTTP response code is not a good way of handling this because:
* mainly: LRA is a protocol layer on top of HTTP/REST. HTTP/REST is for us only the transport protocol.
* this effectively cannot be reused if we ever introduce different transport protocols 
* it introduces the possibility that users will forget about including response bodies (in 500 cases) which will result in unexpected behavior
* it forces users to use javax.ws.rs.Response object as return type so something like this is no longer possible: 
```
    @PUT
    @Path("/compensate")
    @Produces("text/plain")
    @Compensate
    public ParticipantStatus compensate(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
        return isOk(lraId) ? ParticipantStatus.Compensated : ParticipantStatus.FailedToCompensate;
    }
```
-------------------------------------------------------------------------------------------------------------------------

so +1 on moving this forward with the decided solution

Comment (tomjenkinson) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: I think the reason we can't support that code is we got -1 from two voters on allowing 200 for FailedToCompensate? But maybe I misunderstood?

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: > I think the reason we can't support that code is we got -1 from two voters on allowing 200 for FailedToCompensate? But maybe I misunderstood?

@tomjenkinson please ignore that, I just wanted to document why I think it would be better to stick with 200. But please ignore that, I am +1 on the voted solution.

Comment (xstefank) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: @mmusgrov this can be resolved from my side. Please update the PR with the voted results.

Comment (mmusgrov) [reply spec/src/main/asciidoc/microprofile-lra-spec.adoc]: > * mainly: LRA is a protocol layer on top of HTTP/REST. HTTP/REST is for us only the transport protocol.

- The LRA spec is for JAX-RS resources to be able to participate in LRAs and JAX-RS is for REST. Therefore using JAX-RS response codes should be the norm
- We would like to use HTTP for interop between implementations

> * this effectively cannot be reused if we ever introduce different transport protocols

It's a good point but the spec is JAX-RS specific

> * it introduces the possibility that users will forget about including response bodies (in 500 cases) which will result in unexpected behavior

True. If there is status method that will remove any ambiguity and if that isn't available then all I can suggest is that the implementation logs something or the developer has a test suite to catch the problem.

> * it forces users to use javax.ws.rs.Response object as return type so something like this is no longer possible:
> 
> ```
>     @PUT
>     @Path("/compensate")
>     @Produces("text/plain")
>     @Compensate
>     public ParticipantStatus compensate(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
>         return isOk(lraId) ? ParticipantStatus.Compensated : ParticipantStatus.FailedToCompensate;
>     }
> ```
> 

Yes that is a shame since it forces the developer to use a particular coding style. On the plus side using a javax.ws.rs.Response should be natural for JAX-RS developers.

> so +1 on moving this forward with the decided solution

**Comment by mmusgrov** [comment]
@rdebusscher @xstefank  @tomjenkinson @ochaloup : I updated the PR to reflect this discussion. The PR also clarifies the various javadocs that document JAX-RS response codes.

File: api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java [hunk 4]
```diff
@@ -27,42 +27,131 @@
...
+ * start a new LRA will result in a <code>412 PreCondition Failed</code> JAX-RS response
+ * code. On the other hand, combining it with an <code>&#64;LRA</code> annotation that
+ * begins a new LRA can in certain uses case make sense, but in this case the LRA
  * that this method is being asked to compensate for will be unavailable.
+ * </p>
+ *
+ * <p>
+ * If the method is a JAX-RS resource method (or is a non JAX-RS method
+ * annotated with <code>&#64;Compensate</code> with return type
+ * <code>javax.ws.rs.core.Response)</code> then the following are the only
```
xstefank: last closing parenthesis shoud be placed after the <code> block

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -27,42 +27,131 @@
...
+ * </tr>
+ * <tr>
+ *   <td scope="row">404</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource does not know about the LRA</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">412</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource cannot join the LRA because the LRA is no
```
xstefank: This is Compensate annotation. We shouldn't be mentioning joining here I think.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -27,42 +27,131 @@
...
+ *   actual state as reported by the {@link Status} method MUST
+ *   be {@link ParticipantStatus#FailedToCompensate}</p></td>
+ * </tr>
+ * </table>
+ *
+ * <p>
+ * If any other code is returned (or, in the 500 case, the body does not
+ * correspond to a valid state) then the implementation MUST either keep
+ * retrying or attempt to discover the status by calling the
+ * {@link Status} method if present or a combination of both.
```
xstefank: when the retry for `@Compensate` and calls to `@Status` method can be combined? The status method is saying that when it's present then `@Compensate` cannot be invoked multiple times or maybe I am missing something. I would suggest removing "or a combination of both".

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
```diff
@@ -27,44 +27,132 @@
...
+ * </tr>
+ * <tr>
+ *   <td scope="row">404</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource does not know about the LRA</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">412</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource cannot join the LRA because the LRA is no
```
xstefank: same comment as above. I belevie this should be defined as a part of `@LRA` annotation, not `@Complete`

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
```diff
@@ -27,44 +27,132 @@
...
+ *   actual state as reported by the {@link Status} method MUST
+ *   be {@link ParticipantStatus#FailedToComplete}</p></td>
+ * </tr>
+ * </table>
+ *
+ * <p>
+ * If any other code is returned (or, in the 500 case, the body does not
+ * correspond to a valid state)  then the implementation MUST either keep
+ * retrying or attempt to discover the status by calling the
+ * {@link Status} method if present or a combination of both.
```
xstefank: also would remove "or a combination of both" unless this is really possible

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
```diff
@@ -27,44 +27,132 @@
...
+ * start a new LRA will result in a <code>412 PreCondition Failed</code> JAX-RS response
+ * code. On the other hand, combining it with an <code>&#64;LRA</code> annotation that
+ * begins a new LRA can in certain use cases make sense, but in this case the LRA
  * that this method is being asked to complete for will be unavailable.
+ * </p>
+ *
+ * <p>
+ * If the method is a JAX-RS resource method (or is a non JAX-RS method
+ * annotated with <code>&#64;Complete</code> with return type
+ * <code>javax.ws.rs.core.Response)</code> then the following are the only
```
xstefank: parenthesis in <code> block

api/src/main/java/org/eclipse/microprofile/lra/annotation/Forget.java
```diff
@@ -28,46 +28,91 @@
...
+ * with the <code>&#64;LRA</code> annotation (though it is not prohibited).
+ * </p>
  *
+ * <p>
  * Related information is provided in the javadoc for the {@link Status}
+ * </p>
+ * <p>
+ * If the method is a JAX-RS resource method (or is a non JAX-RS method
+ * annotated with <code>&#64;Forget</code> with return type
+ * <code>javax.ws.rs.core.Response)</code> then the following are the only
```
xstefank: parenthesis in <code> block

api/src/main/java/org/eclipse/microprofile/lra/annotation/Forget.java
```diff
@@ -28,46 +28,91 @@
...
+ * </tr>
+ * <tr>
+ *   <td scope="row">200</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The participant may have removed all knowledge of the LRA</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">404</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The method does not know about the LRA</td>
```
xstefank: does that distinction 200 vs 404 have any meaning for the caller? This is not important, just unnecessary for the developer to think about the returned HTTP response code when it's not being used.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
```diff
@@ -35,40 +36,86 @@
...
  * Since the participant generally needs to know the id of the LRA in order
  * to report its status there is generally no benefit to combining this
- * annotation with the `@LRA` annotation (though it is not prohibited).
+ * annotation with the <code>&#64;LRA</code> annotation (though it is not prohibited).
+ * </p>
+ *
+ * <p>
+ * If the method is a JAX-RS resource method (or is a non JAX-RS method
+ * annotated with <code>&#64;Status</code> with return type
+ * <code>javax.ws.rs.core.Response)</code> then the following are the only
```
xstefank: parenthesis in <code> block

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -892,18 +895,20 @@ exposed method from unauthorized access. The specific security details are not s
...
+is still in progress using one of the following options:
 
-- return `ParticipantStatus.Compensating`
-- return a `202 Accepted` HTTP status code
-- mark the method as a JAX-RS asynchronous method (using the
-  `javax.ws.rs.container.Suspended` annotation).
+- Return a `CompletionStage` or mark the method as asynchronous
+  (using the `javax.ws.rs.container.Suspended` annotation). The future must
+  deliver the final status when the stage completes. Please refer to the section
+  <<reactive-support,about reactive support>> for more details.
```
xstefank: Should we also mention what should happen when CompletionStage delivers Compensating?

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
mmusgrov: The reason I added "or a combination of both" is to allow the following behaviour:

1. the coordinator receives an invalid JAX-RS status code from the compensate notification
2. the coordinator sees that it can use the status method to ask for the state of the compensator
3. the status method reports that the compensator is still in the `Active` state
4. the coordinator reissues the compensate notification

ie the coordinator has resolved the ambiguity by using a combination of both strategies. Hence the bit of extra text is required.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
mmusgrov: I removed it.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
mmusgrov: Same response as before

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
xstefank: Very interesting point indeed. There are two ways of how we can look at this. I will open a new issue for 1.x, if you don't mind.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Forget.java
mmusgrov: No but it has advantages. All the other methods use 404 and one could imagine a generic precondition check `if  LRA does not exist then throw WebApplicationException with a 404 response code`. Without it the developer needs to think I don't know about this LRA what response should I return so he still needs to think about it.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
mmusgrov: Copy and paste propagates errors, shame we can't have modular javadoc

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
mmusgrov: Good catch. I removed it from the Complete javadoc too.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
mmusgrov: I see any need for an issue

mmusgrov: There is other text in the spec that covers this

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: I added some text

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
xstefank: I may be confused, can you respond to #187, please? We can resolve this conversation.

mmusgrov: Section [3.2.4. Compensating Activities](https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#324-compensating-activities) for example includes the text:

> If there is no @Status then the @Compensate or @Complete methods will continue to be invoked until the implementation knows it has the final status.

There are other places where I talk about things being retried "... knows it has the final status".

rdebusscher: The scenario described here https://github.com/eclipse/microprofile-lra/pull/176/files#r302883356 is only possible when the first Compensate request never reached the participant. Otherwise the call to status would never be Active.

In that sense, the code of the Compensate will only be executed once when there is a Status method.

Of course, when developer mess things up, by returning Active from the Status method even after the Complete method is already called, multiple calls can happen.

I'm fine with the text "or a combination of both" but then we should maybe add the scenario as an example so that users and implementors know the background.

mmusgrov: @rdebusscher I made two changes:

- include the scenario at the end of the javadoc
- changed MUST keep retrying to SHOULD

This is too avoid the situation where the implementation keeps retrying forever.

mmusgrov: @rdebusscher I have also just updated the @Status and @Forget javadoc to use SHOULD for the same reason with some text about logging a warning if it decides to stop retrying. Notice that compensation violations are allowed in the spec provided that the system writes a log section as described in the [section 3.1](https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#31-the-model):

> ... take recovery actions in the event of failure, but allowing always the possibility that recovery isn’t possible which must be logged or flagged for the administrator and manual intervention may be necessary to restore an application’s consistency.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
```diff
@@ -27,44 +27,138 @@
 import java.time.temporal.ChronoUnit;
 
 /**
- * When a resource method executes in the context of an LRA and if the containing
- * class contains a method annotated with `@Compensate` then the resource will
```
xstefank: sorry, I noticed just now. We should keep the reference to `@Compensate` because otherwise without it `@Complete` won't get invoked.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
```diff
@@ -35,40 +36,86 @@
...
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The resource is attempting to determine the status and
+ *   the caller should retry later</td>
+ * </tr>
+ * <tr>
+ *   <td scope="row">404</td>
+ *   <td scope="row">Empty</td>
+ *   <td scope="row">The method does not know about the LRA</td>
+ * </tr>
+ * </table>
```
xstefank: we can also add something like 

```
<p>
 * If any other code is returned then the implementation SHOULD keep retrying.
 * If it stops retrying a warning message SHOULD be logged.
 * </p>
```

mmusgrov: +1

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -27,42 +27,136 @@
...
+ * start a new LRA will result in a <code>412 PreCondition Failed</code> JAX-RS response
+ * code. On the other hand, combining it with an <code>&#64;LRA</code> annotation that
+ * begins a new LRA can in certain uses case make sense, but in this case the LRA
  * that this method is being asked to compensate for will be unavailable.
+ * </p>
+ *
+ * <p>
+ * If the method is a JAX-RS resource method (or is a non JAX-RS method
+ * annotated with <code>&#64;Compensate</code> with return type
+ * <code>javax.ws.rs.core.Response</code> then the following are the only
```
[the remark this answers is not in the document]
mmusgrov: I think adding approval subject to changes is okay too - ie in this case the person who raised the PR just makes the change and does not need to ask for another review. I will address your comments in this PR.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
mmusgrov: Actually I don't think including this text has any meaning since it is covered implicitly in the javadoc for Compensate/Complete. If you don't mind I will back out this paragraph.

mmusgrov: There are no restrictions on how often the Status method can be called so adding the text does not add anything to the spec.

mmusgrov: Note that if the status cannot be obtained and the implementation stops retrying then the javadoc for Compensate/Complete already indicates that a warning will be logged (ie if we added this text two warnings would be logged).

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
mmusgrov: Resolving here but do note that the text is added back due to a latter comment from the reviewer.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
rdebusscher: @mmusgrov Ok, thanks.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
xstefank: I don't mind not mentioning it here, it is kind of implicit. Can be resolved.

mmusgrov:
The test TckContextTests#testForget tests that participants that return invalid codes are handled correctly. We used 500 for that but #173 now says that 500 is a valid response (to indicate FailedToComplete).

I have [just updated the test](https://github.com/eclipse/microprofile-lra/pull/176/commits/a0063c2128d249750212b7812d469457040f99b8) to use 503 instead. I think it is not contentious to go ahead and push the change without another round of reviews.

PR #181, opened 2019-07-09
Issue 180 Replace InvalidLRAParticipantDefinitionException with any u…

### LINKED ISSUE #180: Replace InvalidLRAParticipantDefinitionException with further unspecified deployment/start failure
@tomjenkinson had a good idea that InvalidLRAParticipantDefinitionException may be an unnecessary burden in the future because we don't really care what RTE will be throw. It is only important that it will fail the deployment/start of the application. So we can remove InvalidLRAParticipantDefinitionException from the specification and update the text accordingly.

xstefank:
…nspecified RuntimeException

resolves #180 


xstefank:
I believe it was. The issue is that signatures are specified in the specification, so anything else will result in RTE when the invalid participant method is called for the first time. Why would someone require such behavior?

xstefank:
@tomjenkinson We can say in the spec that in case the return type is not specified (void, ParticipantStatus, Response or CompletionStage right now) then we treat it as if it was void. Is that what you are asking?

xstefank:
Exactly Tom, that is my original thinking, fail as soon as you detect something is wrong. 

> I am thinking we should maybe consider not mandating that vendors throw an RTE and rather categorise this cause the deployment to fail?

This can certainly be done. I can create another PR after #181 is merged.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckInvalidParticipantSignaturesTests.java
```diff
@@ -120,7 +119,6 @@ public void invalidArgumentTypeInParticipantMethodTest() {
     private void testInvalidDeployment(String deploymentName) {
         deploymentNameRule.deploymentName = deploymentName;
         expectedException.expect(DeploymentException.class);
-        expectedException.expectMessage(InvalidLRAParticipantDefinitionException.class.getSimpleName());
```
tomjenkinson: I suppose you should verify that the exception was a RTE as that would match the wording on line 47

xstefank: But if the implementation throws any subclass of RuntimeException the text "RuntimeException" won't be present in the stacktrace. This is why I removed the check.

xstefank: or @tomjenkinson I will rewrite the issue and PR to include your suggestion about only failing the deployment (thus not mentioning RuntimeException) Thanks for looking into this!

tomjenkinson: No problem - I would be happy with either approach - as you say a subclass of RTE might need some code to do with instanceof and construct it by classname (and expect the Jar in the classpath and so on) only perhaps to be removed if we agree to go for the wording to just fail the deployment :)

PR #182, opened 2019-07-10
issue180 Replace InvalidLRAParticipantDefinitionException with furthe…

### LINKED ISSUE #180: Replace InvalidLRAParticipantDefinitionException with further unspecified deployment/start failure

xstefank:
…r unspecified deployment/start failure

resolves #180 


api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -53,8 +53,8 @@
  *     </code>
  * </pre>
  * would be a valid compensation method declaration. If an invalid signature is detected 
- * the {@link org.eclipse.microprofile.lra.participant.InvalidLRAParticipantDefinitionException} 
- * will be thrown during the application startup.
+ * the implementation of this specification MUST prohibit successful startup of this application
```
rdebusscher: successful startup of **this** application -> successful startup of **the** application

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -53,8 +53,8 @@
...
+ * (e.g. with the runtime exception).
```
rdebusscher: with **the** runtime exception -> with **a** runtime exception

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
```diff
@@ -55,8 +55,8 @@
  *     </code>
  * </pre>
  * would be a valid completion method declaration. If an invalid signature is detected 
- * the {@link org.eclipse.microprofile.lra.participant.InvalidLRAParticipantDefinitionException} 
- * will be thrown during the application startup.
+ * the implementation of this specification MUST prohibit successful startup of this application
```
rdebusscher: same remark

api/src/main/java/org/eclipse/microprofile/lra/annotation/Forget.java
```diff
@@ -59,8 +59,8 @@
  *     </code>
  * </pre>
  * would be a valid forget method declaration. If an invalid signature is detected 
- * the {@link org.eclipse.microprofile.lra.participant.InvalidLRAParticipantDefinitionException} 
- * will be thrown during the application startup.
+ * the implementation of this specification MUST prohibit successful startup of this application
```
rdebusscher: same remark

api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
```diff
@@ -57,8 +57,8 @@
  *     </code>
  * </pre>
  * would be a valid status method declaration. If an invalid signature is detected 
- * the {@link org.eclipse.microprofile.lra.participant.InvalidLRAParticipantDefinitionException} 
- * will be thrown during the application startup.
+ * the implementation of this specification MUST prohibit successful startup of this application
```
rdebusscher: same remark

xstefank:
@rdebusscher Rudy, thanks for the review. Comments addressed in the latest force push.

xstefank:
@mmusgrov TCK is testing that the deployment will fail. This was already agreed upon and merged as part of #148. Why would we allow an application to start if it will fail during the first invocation?

ochaloup:
I think the deployment should be forcibly stopped when does not follow the spec. I think MUST is correct here. I prefer the fail fast practice where it is possible.

mmusgrov:
> @mmusgrov TCK is testing that the deployment will fail. This was already agreed upon and merged as part of #148. Why would we allow an application to start if it will fail during the first invocation?

You are assuming a specific implementation approach for this feature. For example, one implementation might delay checking resource classes until the resource endpoint is accessed so in this case the application would already have deployed.

rdebusscher:
A JAX-RS application fails to deploy when there are 2 resources with the same Path (and not that you have an error when you try to access the URL which is defined twice)

A CDI application fails to deploy when a requested CDI bean doesn't exist (not when you try to use it)

So there are enough other situations where deployment fails when not compliant.

>For example, one implementation might delay checking resource classes until the resource endpoint is accessed so in this case the application would already have deployed

This is what we want to avoid that this is possible. No point of allowing something in production which is known to be failing!

The use of MUST is for the benefit of the developer/user of the spec here.

mmusgrov:
So I was outvoted 3 to 1 so I approved the PR

mmusgrov:
But I can see this being tricky to implement in a pure JAX-RS server.

xstefank:
> But I can see this being tricky to implement in a pure JAX-RS server.

Out of curiosity, I tried with Thorntail, Quarkus and Payara Micro. This fails the deployment in all of them:
```java
@WebListener
public class ContextListener implements ServletContextListener {

    @Override
    public void contextInitialized(ServletContextEvent sce) {
        throw new RuntimeException("ServletContextListener context initialized -- POSSIBLE LRA SCAN");
    }
}
```

rdebusscher:
@xstefank However, pure MicroProfile implementation doesn't need to have Servlets.

@mmusgrov 
>But I can see this being tricky to implement in a pure JAX-RS server.

As mentioned, deployment should fail also when more than 1 JAX-RS resource use the same path. So there is a possibility.
And since it is MicroProfile, you have always CDI available where you can fail the deploy in a CDI extension.

mmusgrov:
> @xstefank However, pure MicroProfile implementation doesn't need to have Servlets.

Section 2.3.2 Servlet of the JAX-RS specification says:

> A JAX-RS application is packaged as a Web application in a .war file.

rdebusscher:
But we aren't allowed to depend on the servlet spec in MicroProfile.

xstefank:
Sorry about that. You are right that is a servlet, not JAX-RS and we cannot rely on it even if it's commonly the base for JAX-RS. 

I agree with Rudy that we should be depending only on existing MicroProfile specifications (umbrella rule). This means that through CDI we can definitely catch the application startup (extensions, `@PostContruct` and `@ApplicationScoped`). However, it seems that in JAX-RS itself doesn't really have a facility to execute code on the application startup. So I will raise this as a new issue and we should decide in the community whether or not we will also require CDI or we will change the wording and the TCK to expect failure on application deployment or on the first request.

However, I would also note that the implementation of LRA will probably always be tied to some container that provides between other bootstrap and implementation of JAX-RS. Cannot we expect that such a container provides a way how modules/libraries can tie into application bootstrap?

Issue #185, opened 2019-07-11
Should LRA depend also on CDI in 1.0


xstefank:
The LRA specification should depend only on existing MicroProfile specifications (umbrella rule). Currently, we depend only on JAX-RS.

As a part of the discussion for #182, we've come to a conclusion that the requirement of failing the deployment/startup of the MicroProfile application when any invalid non-JAX-RS method signature is detected may be problematic in containers where there is only JAX-RS.

The option of execution of some application code during the startup/deployment of the application is available through the CDI. We can definitely catch the application startup, for instance with the CDI extensions or @PostContruct method in the @ApplicationScoped bean. 

The question is whether the LRA specification should depend also on CDI.

The pros I can see right now are:

* it is a part of MicroProfile
* we will be reintroducing LRAClient as an injectable interface in 1.x (probably 1.1)
* most MicroProfile containers, if not all, provide CDI implementation
* there is also a possibility to use CDI in Java SE - https://aboullaite.me/cdi-20-java-se/

The other option would be to not require CDI for LRA 1.0 and depend only on JAX-RS. In that case, the JAX-RS itself doesn't seem to provide any facilities to execute code on the application deployment/startup. 

However, most of the well-known and used JAX-RS implementations (Jersey, RESTEasy, CXF) are based on the Servlet specification and so they can use `ServletContextListener` to execute code on the application startup/deployment. This is not bringing Servlet dependency to LRA specification. It would be an implementation detail of these JAX-RS implementations. I would also note that the implementation of LRA will probably always be tied to some container that provides (between other things) application bootstrap and the implementation of JAX-RS. In the case that JAX-RS implementation is not based on Servlet, cannot we expect that such a container provides a way how modules/libraries can tie into the application bootstrap?

In case we decide that it is not possible to bring CDI in LRA 1.0 I will update the text and the TCK that expects deployment failures to expect deployment failure OR failure on the first request.

tomjenkinson:
I don't see why we can't depend on specifications that are in the MicroProfile platform.

> most MicroProfile containers, if not all, provide CDI implementation

If CDI is in the MicroProfile platform baseline requirement how is it possible to not provide a CDI implementation. Please see [this](https://github.com/eclipse/microprofile/blob/3.0/spec/src/main/asciidoc/required-apis.asciidoc) and [this](https://github.com/eclipse/microprofile/blob/3.0/pom.xml#L214)

I would therefore change that to:

[...]

PR #191, opened 2019-07-15
Fix specification wording with regards to deployment phase

xstefank:
no issue

small fix for the discussion we had with @tomjenkinson in https://github.com/eclipse/microprofile-lra/issues/185#issuecomment-511371730.


xstefank:
in annotations we use "If an invalid signature is detected the implementation of this specification MUST prohibit successful startup of the application (e.g. with a runtime exception)." -- I believe that's ok :)

PR #192, opened 2019-07-16
issue166 Validate if incoming LRA is active

### LINKED ISSUE #166: Spec: Validate if incoming LRA is active
When request contains a correct header indicating a LRA Context (LRA id), the LRA **MUST** have LraStatus.Active in the case of LRA.REQUIRES and LRA.MANDATORY.

Otherwise the participant joins an inactive LRA.

xstefank:
resolves #166 

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
@@ -24,6 +24,7 @@ import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
@@ -215,6 +216,30 @@ public class TckLRATypeTests extends TckTestBase {
     public void neverWithoutEndLRA() {
         resourceRequest(NEVER_WITH_END_FALSE_PATH, false, 200, MethodLRACheck.NOT_PRESENT, false);
     }
+    
+    @Test
+    public void requiredWithClosedLRA() {
+        String lraId = resourceRequest(REQUIRED_PATH, false, 200, MethodLRACheck.NONE, false);
+     
+        // manually call REQUIRED LRA endpoint with the id of already finished LRA
+        Response response = tckSuiteTarget.path(TCK_LRA_TYPE_RESOURCE_PATH).path(REQUIRED_PATH).request()
+            .header(LRA.LRA_HTTP_CONTEXT_HEADER, lraId).get();
+
+        Assert.assertNotEquals("REQUIRED LRA endpoint called with inactive LRA context should " +
+            "start a new LRA", lraId, response.readEntity(String.class));
+    }
+    
+    @Test
+    public void mandatoryWithClosedLRA() {
+        String lraId = resourceRequest(REQUIRED_PATH, false, 200, MethodLRACheck.NONE, false);
+        
+        // manually call MANDATORY LRA endpoint with the id of already finished LRA
+        Response response = tckSuiteTarget.path(TCK_LRA_TYPE_RESOURCE_PATH).path(MANDATORY_PATH).request()
+            .header(LRA.LRA_HTTP_CONTEXT_HEADER, lraId).get();
+        
+        Assert.assertEquals("MANDATORY LRA method should return 412 when called with inactive LRA context",
+            Response.Status.PRECONDITION_FAILED.getStatusCode(), response.getStatus());
+    }
 
     /**
      * Perform a JAX-RS resource request and check the resulting status and whether or not it ran with
@@ -228,7 +253,7 @@ public class TckLRATypeTests extends TckTestBase {
      * @param methodLRAShouldBeActive if true the LRA started by the invoked method should still
      *                               be active after the resource invocation completes
      */
-    private void resourceRequest(String path, boolean startLRA, int expectedStatus, MethodLRACheck lraCheckType,
+    private String resourceRequest(String path, boolean startLRA, int expectedStatus, MethodLRACheck lraCheckType,
                                  boolean methodLRAShouldBeActive) {
         Invocation.Builder target = tckSuiteTarget.path(TCK_LRA_TYPE_RESOURCE_PATH)
                 .path(path).request();
@@ -241,8 +266,8 @@ public class TckLRATypeTests extends TckTestBase {
         Response response = target.get();
 
         try {
-            String methodLRA; // Make it a String as we do some String based tests afterwards
-            String incomingLRA = lra == null ? "" : lra.toASCIIString();
+            String methodLraId; // Make it a String as we do some String based tests afterwards
+            String incomingLraId = lra == null ? "" : lra.toASCIIString();
 
             assertEquals(testName.getMethodName() + ": Unexpected status", expectedStatus, response.getStatus());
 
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -286,10 +311,14 @@ private void resourceRequest(String path, boolean startLRA, int expectedStatus,
...
             }
+            
+            return methodLRA;
         } catch (Throwable e) {
             LOGGER.warning(e.getMessage());
         } finally {
             response.close();
         }
+        
+        return null;
```
tomjenkinson: This might be wrong but I am wondering, should the "return null;" be moved under LOGGER.warning(e.getMessage()); as I think that is the only branch when you would return null?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -286,10 +311,14 @@ private void resourceRequest(String path, boolean startLRA, int expectedStatus,
             if (lra != null) {
                 lraClient.closeLRA(lra);
             }
+            
+            return methodLRA;
```
tomjenkinson: It somewhat contravenes my other comment but how about changing this to the return incomingLRA (I assume this is an ID) and only return it if you call lraClient.closeLRA(lra); (i.e move it a few lines up).

I am not sure if you agree but I think it might help readability of the new tests.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -215,6 +216,30 @@ public void neverWithEndRA() {
     public void neverWithoutEndLRA() {
         resourceRequest(NEVER_WITH_END_FALSE_PATH, false, 200, MethodLRACheck.NOT_PRESENT, false);
     }
+    
+    @Test
+    public void requiredWithClosedLRA() {
+        String closedLRA = resourceRequest(REQUIRED_PATH, false, 200, MethodLRACheck.NONE, false);
```
tomjenkinson: I think it might aid readability if this was:
> String lraId = resourceRequest(....
Of course, I might be wrong about what you are expecting here but I think it is an ID?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -215,6 +216,30 @@ public void neverWithEndRA() {
...
+        Response response = tckSuiteTarget.path(TCK_LRA_TYPE_RESOURCE_PATH).path(REQUIRED_PATH).request()
+            .header(LRA.LRA_HTTP_CONTEXT_HEADER, closedLRA).get();
+
+        Assert.assertNotEquals("REQUIRED LRA endpoint called with inactive LRA context should " +
+            "start a new LRA", closedLRA, response.readEntity(String.class));
+    }
+    
+    @Test
+    public void mandatoryWithClosedLRA() {
+        String closedLRA = resourceRequest(REQUIRED_PATH, false, 200, MethodLRACheck.NONE, false);
```
tomjenkinson: I think it might aid readability if this was:
> String lraId = resourceRequest(....
Of course, I might be wrong about what you are expecting here but I think it is an ID?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
xstefank: I can change it to incomingLRA, they are the same just methodLRA is returned from the resource in the response body. But I cannot change to return it only when the lra is closed (I guess you refer to [this](https://github.com/eclipse/microprofile-lra/pull/192/files#diff-33b883a0ff0f67ccf92bc5322022c6cfR305)) because we want to return it even in case the LRA was closed when the REST method finished (automatically). Sorry if I am missing what you meant.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
xstefank: can do, thanks

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
xstefank: will do, thank you

tomjenkinson: thanks!

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
tomjenkinson: I was thinking about [this](https://github.com/eclipse/microprofile-lra/pull/192/files#diff-33b883a0ff0f67ccf92bc5322022c6cfR312) but I can appreciate your comment here. I think the main thing is to make it clear what is being returned (I am thinking from the point of view of the review of this PR I could not immediately decipher what might be in "methodLRA" and was thinking it would only be required in that case I link to earlier in this comment but I could be wrong there.

perhaps
> return methodLRA; // expected to contain the LRA ID

Would be enough?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
tomjenkinson: Thanks!

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
xstefank: ah, now I understand :) thanks, will update this all in one go later today

xstefank:
@tomjenkinson updated, thanks for the review!

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -241,53 +266,56 @@ private void resourceRequest(String path, boolean startLRA, int expectedStatus,
...
                 // otherwise it should be finished
-                assertTrue(lraClient.isLRAFinished(URI.create(methodLRA)));
+                assertTrue(lraClient.isLRAFinished(URI.create(methodLraId)));
             }
 
             if (lra != null) {
                 lraClient.closeLRA(lra);
             }
+            
+            return methodLraId;
```
rdebusscher: I would place this at the end of the method because now it seems that method never returns something (also requires the variable definition to be outside of the try {} )
But personal opinion.

tomjenkinson: If we go with that we should put methodLraId = null on line 318 I think and go with a single return statement?

xstefank: I can look at this again when I'll do final sweep/cleaning before Final release :)

mmusgrov: The method initialises and returns the return value in the try block.
The method returns null in the catch block.

ie the code is safe and is readable so what is the problem with it.

rdebusscher: No real problem, just strange that the method has no return statement as last statement (one have to look for the location where the actual return happens)

mmusgrov: Okay, but you would still need to look back to see where the actual assignments to the return value occur. Returning the value in situ is arguably more readable than setting values and returning at the end.

Issue #187, opened 2019-07-12
Idempotent Compensations in case of participant failures when Status method is present


xstefank:
@mmusgrov raised and interesting [scenario](https://github.com/eclipse/microprofile-lra/pull/176#discussion_r302883356):

* the coordinator receives an invalid JAX-RS status code from the compensate notification
* the coordinator sees that it can use the status method to ask for the state of the compensator
* the status method reports that the compensator is still in the Active state
* the coordinator reissues the compensate notification
ie the coordinator has resolved the ambiguity by using a combination of both strategies. 

This seems to contradict [section 3.2.4](https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#324-compensating-activities):

> If the compensation and completion methods are not idempotent then there MUST be a method annotated with @Status which reports the status. Otherwise the compensation and completion methods should return the status. 

So we may call the not idempotent Compensation/Completion more times in case of failures (the compensate method runs successfully and status code is changed by the JAX-RS framework because of some error in the response). This is an edge case scenario, but maybe it should be documented in the specification -- meaning that in case of errors even the idempotent compensation can be replayed. Or is there any option for how we can mitigate it?

mmusgrov:
If the developer does not follow the spec (eg does not report the status correctly when it returns 500 when asked to compensate) then its fine to add caveats.

Why not just drop the requirement for idempotency. Does anyone have opinions about that proposal?

mmusgrov:
But the wording in section 3.2.4 that you reference does need updating in PR #176 - I will add a clarification now ...

mmusgrov:
@xstefank I added the extra wording is the last commit on PR #176 in commit https://github.com/eclipse/microprofile-lra/pull/176/commits/d35143f34654f824dd26ebac2e94b18ad7133848

tomjenkinson:
I don't think we should drop idempotency requirements

mmusgrov:
> I don't think we should drop idempotency requirements

Any particular reason why not? All REST design guidelines that I've read advocate idempotency so if we say the compensate callback MUST be idempotent then we can remove the `@Status` annotation which has to be a good thing.

mmusgrov:
So if we go back to using `PUT` for the compensate and complete notifications then we can justify the idempotency requirement by [quoting the IETF spec.](https://tools.ietf.org/html/rfc7231#section-8.1.3).

[...]

PR #198, opened 2019-07-25
issue146 Specifiy JAX-RS request methods on annotations

mmusgrov:
#146 

This PR clarifies which JAX-RS request methods can be used on annotations.

@rdebusscher I added a note about including an optional status URL in `202 Accepted` responses to cover the situation where there is no `@Status` method. If you accept this text then I don't think we need a new issue for it.

I also did not change any text that discussed idempotency since we are still discussing that on issue #187

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -755,7 +755,7 @@ annotations when associated with JAX-RS resource methods:
 
 [[jaxrs-response-table]]
 |===
-| Annotation | Prefered HTTP method | Expected status codes | Response
+| Annotation | Required HTTP method | Expected status codes | Response
 
 | `@Compensate`
 | PUT
@@ -786,9 +786,6 @@ annotations when associated with JAX-RS resource methods:
 
 Please refer to the javadoc for each annotation for the description of the
 conditions under which the various JAX-RS response codes are returned.
-
-The presented HTTP methods SHOULD be applied at individual JAX-RS methods
-but they are not enforced by the specification. 
 
 If the method annotated with `@AfterLRA` returns an unexpected HTTP status
 or never reaches the caller then the implementation MUST invoke the same method again.
@@ -911,7 +908,15 @@ is still in progress using one of the following options:
   it exists, and if there is no such method it will reinvoke the `@Compensate`
   method). Please refer to the section
   <<reactive-support,about reactive support>> for more details.
-- A JAX-RS method can return a `202 Accepted` HTTP status code
+- A JAX-RS method can return a `202 Accepted` HTTP status code.
+  If there is no `@Status` method then the response MAY provide
+  a status URL in the `HTTP Location` header field so that the
+  implementation can discover the final outcome.
+  This URL, if present, MUST obey the requirements specificed in
+  the javadoc for the <<source-Status,Status annotation>>.
+  If the developer has not provided an `@Status` method nor a status
+  URL then the implementation MUST reinvoke the `@Compensate` method
+  (ie it MUST be idempotent).
 - A non JAX-RS method can return `ParticipantStatus.Compensating`.
 
 The `@Status` method, if present, MUST report the progress of the compensation.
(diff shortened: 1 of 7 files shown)


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -786,7 +786,7 @@ annotations when associated with JAX-RS resource methods:
 Please refer to the javadoc for each annotation for the description of the
 conditions under which the various JAX-RS response codes are returned.
 
-The presented HTTP methods SHOULD be applied at individual JAX-RS methods
+The presented HTTP methods MUST be applied at individual JAX-RS methods
```
mmusgrov: Oops I will update this paragraph.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java
```diff
@@ -296,8 +295,7 @@ public Response compensateWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId,
         return getEndPhaseResponse(false);
     }
 
-    // Complete methods SHOULD use @PUT but to verify that other HTTP methods are also supported use here HTTP POST
-    @POST
+    @PUT
     @Path("/complete")
     @Complete
     public Response completeWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId,
```
xstefank: I can't comment on lines that haven't been changed, but you missed `@Status` method on line [313](https://github.com/eclipse/microprofile-lra/pull/198/files#diff-4c219933e42e7b38d324e42b907623a2R313) which is still defined with `@PUT` and not `@GET`

mmusgrov: Well spotted

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -910,7 +907,15 @@ is still in progress using one of the following options:
   it exists, and if there is no such method it will reinvoke the `@Compensate`
   method). Please refer to the section
   <<reactive-support,about reactive support>> for more details.
-- A JAX-RS method can return a `202 Accepted` HTTP status code
+- A JAX-RS method can return a `202 Accepted` HTTP status code.
+  If there is no `@Status` method then the response MAY provide
+  a status URL in the `HTTP Location` header field so that the
+  implementation can discover the final outcome.
```
rdebusscher: Ok, During the discussions I always had the impression that the location Header would be made available in the case when there IS @Status method available.

In this scenario, the Location Header becomes an alternative for the @Status method and reinvoking the Compensate method and makes sense.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -910,7 +907,15 @@ is still in progress using one of the following options:
...
+  This URL, if present, MUST obey the requirements specificed in
+  the javadoc for the <<source-Status,Status annotation>>.
+  If the developer has not provided an `@Status` method nor a status
+  URL then the implementation may reinvoke the `@Compensate` method
```
rdebusscher: `may`? It MUST reinvoke as it otherwise has no way of retrieving the final status.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -910,7 +907,15 @@ is still in progress using one of the following options:
...
+  (ie it MUST be idempotent).
```
rdebusscher: applies also for @Complete

mmusgrov: The text after the paragraph describing Compensate says:

> Similarly if the resource cannot perform a completion activity immediately.

mmusgrov: I think I've resolved everything here so @rdebusscher I will merge the PR at close of business today unless you have any further issues with it?

PR #204, opened 2019-07-31
Fix REST invocation in TckContextTests

xstefank:
Fix TCK errors introduced in #198 with regards to correct REST HTTP methods.


mmusgrov:
This is a bug fix so I think it is okay to merge without a review from @rdebusscher

Issue #206, opened 2019-08-01
The LRA state model diagram is missing an allowed transition


mmusgrov:
Nested LRAs are allowed to transition from the state Closed to the state Cancelling. This transition is missing from the [LRA state model diagram](https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/images/lra-state-model.png).

mmusgrov:
@xstefank thanks for pointing this out. Could you review it please?

PR #207, opened 2019-08-01
issue206 Missing state transition in the LRA state model diagram

mmusgrov:
#206 
This bug fix adds the missing transition

#208 is the TCK test for this behaviour.


mmusgrov:
@xstefank thanks for pointing this out. Could you review it please?

spec/src/main/asciidoc/images/lra-state-model.txt
```diff
@@ -56,6 +56,8 @@ state LRAContext {
   ActiveLRA --> Closing : close
   Closing --> Closed
   Closing --> FailedToClose
+
+  Closed --> Cancelling : cancel (transition is only allowed if LRA is nested)
```
tomjenkinson: How does a nested participant know that it can release resources? It seems it cannot release resources when it is closed so I presume some other callback is issued but I don't see such a notion mentioned in 3.2.7

xstefank: I agree. We can require that nested LRA participants must declare Forget methods. In that sense:
* if nested LRA closes, its participants must keep resources (same as current behavior)
* if nested LRA cancels, its participant can release resources
* if parent LRA closes, Forget method must be called even on nested LRA participants

mmusgrov: `@AfterLRA` could be used

mmusgrov: The participant knows whether or not he is executing in a nested context.

If he is enlisted with a nested LRA then he would hold onto any required state until it got the AfterLRA notification.
If he is enlisted with a top level LRA then he would release any required state when he has processed the compensate or complete notification correctly.

mmusgrov: Okay let's go with `@Forget` then. If there are no other opinions then I will push a spec update.

tomjenkinson: +1 to `@Forget`

mmusgrov: @tomjenkinson and @xstefank I have added some clarifying remarks about the use of `@Forget`

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -933,7 +933,9 @@ will start a new LRA whose behaviour is as follows:
 1. A nested LRA can close or cancel independently of its parent.
 2. A nested LRA which has closed must retain the ability to cancel the
    effects if the the parent cancels. This requirement must be enforced
-   by participants.
+   by participants. If the participant has a Forget method then it MUST
+   be invoked if the parent LRA is closed. The Forget method is described
+   in the section <<forgetting-an-lra,Forgetting an LRA>>
```
xstefank: missing comma at the end

mmusgrov:
@rdebusscher I am merging this one since it is a bug fix and has two approvers,

PR #213, opened 2019-08-04
Stronger wording on @LRA placement

rdebusscher:
and explicitly indicating where @LRA can be used.


rdebusscher:
@xstefank as I can't add you as reviewer

PR #215, opened 2019-08-05
Fix some TCK tests of #205 and improve 

rdebusscher:
Fix: LRA Status was reported with Class name but consulted without from MetricService
Improve: correct deployment name, better assert messages, ...

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckUnknownStatusTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckUnknownStatusTests.java
@@ -50,7 +50,7 @@ public class TckUnknownStatusTests extends TckTestBase {
 
     @Deployment(name = "tckunkownstatus")
     public static WebArchive deploy() {
-        return TckUnknownStatusTests.deploy(TckTests.class.getSimpleName().toLowerCase());
+        return TckUnknownStatusTests.deploy(TckUnknownStatusTests.class.getSimpleName().toLowerCase());
     }
 
     @Before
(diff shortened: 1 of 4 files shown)


rdebusscher:
@xstefank Also for you to review

rdebusscher:
I agree that 412 status code is the logical result of the compensate/complete methods but due to another reason.

The @LRA(value = LRA.Type.REQUIRED) starts a new LRA (as mentioned in the spec since there is an id present but an inactive one so REQUIRED must create a new LRA) but method logic (when coded in a real world situation) must return status 412 since it doesn't know about this LRA.

I'mm adjust the location of the @LRA in the test class

rdebusscher:
The behaviour of LRA annotation and JAX-RS participant methods (for example Complete/Compensate), make class level placement of annotation not feasible.

mmusgrov:
When I was running these tests against our implementation I needed to set the consistency delay to something bigger than our "periodic recover scanner" interval. These are the only tests where we need to set a non zero delay. Is it possible for you to introduce a second config property (such as "lra.tck.consistency.delay2") so that implementations can set a different delay for these specific tests. It would mean changing your tests to call `applyConsistencyDelay2()` instead of `applyConsistencyDelay()`. The resulting time difference it makes to our TCK run is huge.

rdebusscher:
@mmusgrov We should call the 2 delays `shortDelay` and `longDelay`.

The **_shortDelay_** needs to be used when we just need to wait for the implementation to call the participant methods
The **_longDelay_** is used when multiple calls to these participant methods needs to be performed.

mmusgrov:
Okay. I added this recommendation in PR https://github.com/eclipse/microprofile-lra/pull/219

Issue #218, opened 2019-08-07
Add a second TCK config property to configure the consistency delay period


mmusgrov:
We have a property called `lra.tck.consistency.delay` which adds a delay into each TCK test. Some implementations only require a delay on the recently added TckUnknownTests in the PR for issue #195 which require multiple passes before the final state is reached. I would like to introduce a second config property called `lra.tck.consistency.delay2` which we can apply to the TckUnknownTests tests. This way we can use a shorter delay for the other tests.

rdebusscher:
I would call the properties `lra.tck.consistency.delay.short` (the current one) and `lra.tck.consistency.delay.long`

mmusgrov:
Even though each implementation can set them to arbitrary values?

mmusgrov:
@xstefank recommends using the [helper method `replayEndPhase()`](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTestBase.java#L157) instead of (or as well as) the method `applyConsistencyDelay()`. Our implementation works better by triggering a recovery scan and it seems yours prefers a delay so perhaps using both is the way to go (delay and replayEndPhase) to cater for both approaches.

xstefank:
@mmusgrov were you able to verify with Narayana that `replayEndPhase()` works in this scenario, please? I believe we should stay consistent then as other similar scenarios, for instance `TckContextTests#testStatus()` rely on `replayEndPhase()`. 

Trigger for recovery scan can be a useful feature which can be introduced in later version of the spec (for instance as a method on the `LRAClient`).

My apologies for not getting a deeper knowledge of these TCK tests, I am extremely busy with other work.

mmusgrov:
@xstefank We have two mechanisms in the TCK for testing eventual consistency:

1. Delaying for an arbitrary period of time.
2. Explicitly requesting a replay of the termination phase of an LRA.

The Narayana implementation of LRA can work with both approaches but I prefer the second since it is less arbitrary and speeds up the TCK. The tests introduced by @rdebusscher only allow an implementation to use approach 1 and this is why I asked for an option to use a shorter delay for these particular tests. But I would prefer that the test explicitly requests a replay of the end phase so that the Narayana TCK run does not have to depend on timeouts.

xstefank:
@mmusgrov I agree, replayEndPhase() should be the easier solution.

And I also think that standardizing the trigger for recovery can be one of the features in 1.x.

[...]

PR #219, opened 2019-08-08
issue218 Define long and short consistency delay config properties

mmusgrov:
#218 
This PR implements the recommendation on issue https://github.com/eclipse/microprofile-lra/pull/215#issuecomment-519186466


rdebusscher:
`consistencyDelay * 2 ` is an option but I like the greater flexibility of 2 separate delays.

Probably not worth it because it is only the TCK, but when longDelay is not defined, his value is shortDelay * 2.

xstefank:
what I am afraid of is that this may hide some errors. But then it's the question of what the TCK purpose is. I don't like the idea, but I guess I am in minority :) However, I still think we should unify then the tests that rely on `replayEndPhase()` right now to also work with `longDelay` just to stay consistent.

PR #229, opened 2019-08-29
issue228 TckParticipantTests require implementations to block on meth…

### LINKED ISSUE #228: TCK: TckParticipantTests require implementations to block on methods which return CompletionStage
`TckParticipantTests#testNonJaxRsCompletionStageResponseAndParticipantStatus` and `TckParticipantTests#testNonJaxRsCompletionStageVoid` both finish with the invocation of method returning a `CompletionStage`. Some implementations (Narayana now) may choose to not block when the non-JAX-RS participant method returns `CompletionStage`. This means that we should give the implementation time to finish the execution as there is an injected delay in all returned `CompletionStage`s. In other words, the test will finish before the completion stage and this means that the LRA is left active after the test ends.

xstefank:
…ods which return CompletionStage

resolves #228 


xstefank:
@rdebusscher for your review

PR #235, opened 2019-09-13
Issue227 replace TckTestBase#replayEndPhase with an implementation pr…

### LINKED ISSUE #227: TCK: replace TckTestBase#replayEndPhase with an implementation provided recovery trigger
Provide a custom TCK interface whose implementation is supplied to the TCK by the implementor of this specification through the ServiceLoader mechanism. This interface will provide `triggerRecovery(lraId)` and `triggerRecovery()` methods which the implementor will use to trigger repeated Complete, Compensate or Status calls for a particular LRA or all active LRAs in state Completing or Compensating.

xstefank:
…ovided recovery trigger

resolves #227 

This is the initial replacement. I will later update TckUnknown* tests when this PR gets merged.


tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LraRecoveryService.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Iterator;
+import java.util.ServiceLoader;
+import java.util.logging.Logger;
+
+@ApplicationScoped
+public class LraRecoveryService {
+
+    private static final Logger LOG = Logger.getLogger(LraRecoveryService.class.getName()); 
+    
+    private LraRecovery lrarRecovery = lookupLraRecovery();
```
rdebusscher: Usage of @PostConstruct since this is a CDI bean?

rdebusscher: minor: typo lrarRecovery

xstefank: post construct will be run only once the bean is constructed, so I chose to rather use static method which will fail sooner.

tomjenkinson: I wonder if the ServiceLoader is guaranteed to be available during the construction of this class? And assuming it is what guarantees that the service is registered pre the construction of the class - I guess there should be some notes added to the TCK guide about the expectations?

xstefank: @tomjenkinson per discussion on gitter I will replace the service loader in this PR with CDI definition that will be supplied to the Arquillian directly.

tomjenkinson: Good plan, thanks @xstefank

xstefank:
@rdebusscher, @tomjenkinson I remove ServiceLoader approach. I also removed artificial one second delays in ValidLRACSParticipant as they would force implementor to wait for CompletionStages to end. I had to add one delay to the testNonJaxRsCompletionStageResponseAndParticipantStatus because in this scenario we depend on the outcome of the CS from Complete call.

tck/running_the_tck.asciidoc
```diff
@@ -17,6 +17,12 @@
 
 The TCK uses `JUnit 4` and `Arquillian`.
 
+== SPI
+
+In order to run TCK, the implementors must provide an implementation of
+`org.eclipse.microprofile.lra.tck.service.spi.LraRecoveryService` and declare
+it as a CDI bean so it can be used in the TCK environment.
```
tomjenkinson: Thanks!

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -161,11 +165,24 @@ public void testNonJaxRsCompletionStageResponseAndParticipantStatus() throws Int
...
-        replayEndPhase(ValidLRACSParticipant.ROOT_PATH);
+        lraRecoveryService.triggerRecovery(lraId);
+    }
+
+    /**
+     * Artificially wait for CompletionStage in {@link ValidLRACSParticipant} to finish
+     */
+    private void waitForCompletionStage() {
+        try {
+            Thread.sleep(1000);
```
tomjenkinson: @xstefank - is there nothing else we can do here except the sleep? I find these to be unreliable myself, wdyt @rdebusscher ?

xstefank: @tomjenkinson we should wait for the completion stage to finish before we run the recovery. There is really just a small window, more of a context switch, which may or may cause race condition (depending on how the implementor triggers recovery). We need to wait for the return value (Response Accepted) in this scenario, so I don't know whether we can do anything else here.

xstefank:
@rdebusscher, @tomjenkinson after some reruns I was able to pass TCK with Narayana even without any timeouts so I removed them. The PR is ready for review.

PR #236, opened 2019-09-15
Remove timeLimit and timeUnit for @Compensate and @Complete

### LINKED ISSUE #233: Remove TimeLimit and TimeUnit on @Complete and @Compensate
@Complete is optional and thus it makes more sense that developer specifies timeout on @Compensate. 

It also makes more sense `When timeLimit is reached, the @Compensate is called` (so they have defined the time limit on the method which get called) which is not the case when defined on @Complete because the @Compensate will be called in that case.

rdebusscher:
Fixes #233

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
@@ -172,23 +172,5 @@ import java.time.temporal.ChronoUnit;
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD})
 public @interface Compensate {
-    /**
-     * The period for which the participant will guarantee it will be able
-     * to compensate for any work that it performed during the associated LRA.
-     * When this period elapses the LRA that it joined becomes eligible for
-     * cancellation. The units are specified in the {@link #timeUnit()}
-     * attribute.
-     *
-     * A value of zero indicates that it will always be able to compensate.
-     *
-     * @return the period for which the participant can guarantee it
-     * will be able to compensate when asked to do so
-     */
-    long timeLimit() default 0;
 
-    /**
-     * @return the unit of time that the {@link #timeLimit()} attribute is
-     * measured in.
-     */
-    ChronoUnit timeUnit() default ChronoUnit.SECONDS;
 }
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
@@ -37,10 +37,9 @@ import java.lang.annotation.Target;
  * </p>
  *
  * <p>
- * In the case the ability to compensate the Long Running Action is time bounded, this is, the action can only be
- * undone within a certain time limit when the action has started, you can limit the lifespan of the
- * Long Running action with the {@see LRA.timeLimit} and {@see LRA.timeUnit}. With these parameters you can define
- * The maximum lifetime of the Long Running Actions and makes it eligible for automatic cancellation.
+ * In the case where the ability to compensate the Long Running Action is time bounded, you can limit the lifespan
+ * of the Long Running action by providing values for the {@see LRA.timeLimit} and {@see LRA.timeUnit} attributes.
+ * When the time limit is reached the LRA becomes eligible for automatic cancellation.
  * </p>
  *
  * <p>
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
@@ -24,7 +24,6 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
-import java.time.temporal.ChronoUnit;
 
 /**
  * <p>
@@ -35,6 +34,13 @@ import java.time.temporal.ChronoUnit;
  * If the annotation is present on more than one method then an arbitrary one
  * will be chosen. The LRA specification makes no guarantees about when
  * Compensate method will be invoked, just that it will eventually be called.
+ * </p>
+ *
+ * <p>
+ * In the case the ability to compensate the Long Running Action is time bounded, this is, the action can only be
+ * undone within a certain time limit when the action has started, you can limit the lifespan of the
+ * Long Running action with the {@see LRA.timeLimit} and {@see LRA.timeUnit}. With these parameters you can define
+ * The maximum lifetime of the Long Running Actions and makes it eligible for automatic cancellation.
  * </p>
  *
  * <p>
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
@@ -39,10 +39,9 @@ import java.lang.annotation.Target;
  * </p>
  *
  * <p>
- * In the case the ability to complete the Long Running Action is time bounded, this is, the action can only be
- * confirmed within a certain time limit when the action has started, you can limit the lifespan of the
- * Long Running action with the {@see LRA.timeLimit} and {@see LRA.timeUnit}. With these parameters you can define
- * The maximum lifetime of the Long Running Actions and makes it eligible for automatic cancellation.
+ * In the case where the ability to complete the Long Running Action is time bounded, you can limit the lifespan
+ * of the Long Running action by providing values for the {@see LRA.timeLimit} and {@see LRA.timeUnit} attributes.
+ * When the time limit is reached the LRA becomes eligible for automatic cancellation.
  * </p>
  *
  * <p>
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
@@ -174,23 +174,5 @@ import java.time.temporal.ChronoUnit;
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD})
 public @interface Complete {
-    /**
-     * The period for which the participant will guarantee it will be able
-     * to complete for any work that it performed during the associated LRA.
-     * When this period elapses the LRA that it joined becomes eligible for
-     * cancellation. The units are specified in the {@link #timeUnit()}
-     * attribute.
-     *
-     * A value of zero indicates that it will always be able to complete.
-     *
-     * @return the period for which the participant can guarantee it
-     * will be able to complete when asked to do so
-     */
-    long timeLimit() default 0;
 
-    /**
-     * @return the unit of time that the {@link #timeLimit()} attribute is
-     * measured in.
-     */
-    ChronoUnit timeUnit() default ChronoUnit.SECONDS;
 }
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
@@ -24,7 +24,6 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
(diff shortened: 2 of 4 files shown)


rdebusscher:
@xstefank FYI

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1009,26 +1009,36 @@ example, to indicate that an LRA should automatically cancel after 100 milliseco
...
-the LRA is cancelled and participants will be notified via their compensation
-(i.e. the method annotated with `@Compensate`). To set such a time limit add the
-`timeLimit` and `timeUnit` element to the `@Compensate` and `@Complete`
-annotations. If different time limits are set on these two methods then the
-earliest of the two will take effect. An example of how a participant could
-indicate that its' ability to compensate is limited to 100 milliseconds could be:
+When the timeout is reached, the LRA moves to the state `Cancelling` and all registered participants receive a `Compensate`
+call. Just as if the Long Running Action would be closed due to an error.
+
+When a participant joins an existing transaction using the `@LRA(value = LRA.Type.MANDATORY)` annotation, the timeout of
```
xstefank: MANDATORY may be misleading here as this also applies for REQUIRED and SUPPORTS for instance. Can we just say "joins an existing transaction using the `@LRA(...)`" or "joins an existing transaction (for instance using the `@LRA(value = LRA.Type.MANDATORY)`)"?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1009,26 +1009,36 @@ example, to indicate that an LRA should automatically cancel after 100 milliseco
...
+
 ----
 
-In this example after 100 milliseconds has passed the implementation SHOULD
-automatically cancel the LRA which will result in this method being invoked.
+Suppose you have a call to the `someService` endpoint at 10h25, the time out is at that moment calculated at 11h25.
+Suppose that `anotherService` is called for the same Long Running Action, this is the LRAid is in the header for
+the call at the endpoint, at 10h27. The timeout related to this call is 10h28. Since these 2 calls share the same
+Long Running action, the LRA will be eligible for automatic cancellation from 10h28 on although the time out was initially 1 hour,
+it can be cancelled already after 3 minutes.
```
mmusgrov: This text seems to imply that services are using synchronised clocks. I would just say that if two timeout values have been set on an active context then the earliest one that reaches the implementation will trigger cancellation of the LRA.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -172,23 +172,5 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD})
 public @interface Compensate {
-    /**
```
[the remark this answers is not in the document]
rdebusscher: Added a reference to the @LRA.timeUnit and @LRA.timeLimit members in the JavaDoc and mentioned the time-bounded aspect.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Complete.java
```diff
@@ -174,23 +174,5 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.METHOD})
 public @interface Complete {
-    /**
```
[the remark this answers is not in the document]
rdebusscher: Added a reference to the @LRA.timeUnit and @LRA.timeLimit members in the JavaDoc and mentioned the time-bounded aspect.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1009,26 +1009,36 @@ example, to indicate that an LRA should automatically cancel after 100 milliseco
       @HeaderParam(LRA.LRA_HTTP_CONTEXT_HEADER) URI lraId) {...}
 ----
 
-Furthermore, the ability to compensate or complete may be transient capabilities of a
-service so participants can also be timed out. When the time limit is reached
```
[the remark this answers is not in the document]
rdebusscher: Kept sentence from the original spec text.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1009,26 +1009,36 @@ example, to indicate that an LRA should automatically cancel after 100 milliseco
...
-the LRA is cancelled and participants will be notified via their compensation
-(i.e. the method annotated with `@Compensate`). To set such a time limit add the
-`timeLimit` and `timeUnit` element to the `@Compensate` and `@Complete`
-annotations. If different time limits are set on these two methods then the
-earliest of the two will take effect. An example of how a participant could
-indicate that its' ability to compensate is limited to 100 milliseconds could be:
+When the timeout is reached, the LRA moves to the state `Cancelling` and all registered participants receive a `Compensate`
+call. Just as if the Long Running Action would be closed due to an error.
```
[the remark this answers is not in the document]
rdebusscher: I guess this is about the terminology used by the people (and thus why a terminology section in the beginning would be useful)
- Closing LRA -> Ending/Stopping the LRA (no longer active)
- Completing LRA -> successful closing LRA and @Complete will be called
- Cancelling LRA -> Since we can only cancel the LRA today by having an HTTP status 4xx or 5xx (by default, configurable) this means that an LRA is cancelled only after an 'error' occurred.

I reworded the last sentence and no longer used 'error'.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
rdebusscher: reworded

mmusgrov: The text is still using absolute times and would require the two services to have to synchronise their clocks. The impossibility of synchronising clocks across entities in a distributed system was discussed in one of Lamport's papers. I would just say that if two timeout values have been set on an active context then the earliest one calculated by the implementation will trigger cancellation of the LRA.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -38,6 +37,13 @@
  * </p>
  *
  * <p>
+ * In the case the ability to compensate the Long Running Action is time bounded, this is, the action can only be
+ * undone within a certain time limit when the action has started, you can limit the lifespan of the
+ * Long Running action with the {@see LRA.timeLimit} and {@see LRA.timeUnit}. With these parameters you can define
+ * The maximum lifetime of the Long Running Actions and makes it eligible for automatic cancellation.
```
mmusgrov: The grammar could be improved (for example there is a new sentence without a full stop). Better would be something like:
> In the case where the ability to compensate the Long Running Action is time bounded you can limit the lifespan of the Long Running action by providing values for the {@see LRA.timeLimit} and {@see LRA.timeUnit} attributes. When the time limit is reached the LRA becomes eligible for automatic cancellation.

I have a similar comment regarding the javadoc for the Complete annotation below.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1009,26 +1009,38 @@ example, to indicate that an LRA should automatically cancel after 100 milliseco
...
-service so participants can also be timed out. When the time limit is reached
-the LRA is cancelled and participants will be notified via their compensation
-(i.e. the method annotated with `@Compensate`). To set such a time limit add the
-`timeLimit` and `timeUnit` element to the `@Compensate` and `@Complete`
-annotations. If different time limits are set on these two methods then the
-earliest of the two will take effect. An example of how a participant could
-indicate that its' ability to compensate is limited to 100 milliseconds could be:
+Furthermore, the ability to compensate or complete may be transient capabilities of a service so participants can also be timed out.
+When the timeout is reached, the LRA moves to the state `Cancelling` and all registered participants receive a `Compensate`
+call. The same actions are performed then the case where the LRA is canceled.
```
mmusgrov: The grammar is odd, better is:

> The same actions are performed as in the case when an LRA is cancelled.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: For example if the message that carries the timeout values is delayed in the network then this would throw off the absolute time calculations.

mmusgrov:
How does this issue relate to issue #233

spec/src/main/asciidoc/microprofile-lra-spec.adoc
rdebusscher: > if two timeout values have been set on an active context then the earliest one calculated by the implementation will trigger cancellation of the LRA.

This also assumes synchronization as it needs to know the 'earliest one'

rdebusscher: So if we do not have any possibility to reliably determine the time, then we should change the spec and indicate that defining another timeout on subsequent participants is forbidden.

mmusgrov: All posix compliant clocks are monotonic so the concept of earliest does not depend on synchronised clocks. A clock which ticks backwards would brake all kinds of stuff. Therefore a local clock can reliably order time. 

If the implementation is distributed across nodes then the first node whose time scheduler fires will trigger cancellation of the LRA.

rdebusscher: Let me elaborate more on what I meant (I was at a conference and replied too condensed)

The example I have put in the spec is from the point of view of the end user. He has 'absolute' timings, namely from his perspective, there are things happening on the services as a response to his actions.
So when he is watching his clock, a request can hit service A at 10h25, and Service B at 10h27. And this regardless of the timings on the Servers running Service A and B.

And the 'earliest' still assumes some kind of synchronization of the clocks.  200 years from now in the western calendar (say the year 2220) is already a few hundred years in the past according to the Buddhist Calendar today.

Clocks can go back in time! When servers are using the time Services, adjustments can happen so that clock jumps back. And even a more obvious example, when we switch from Summer to Winter time, clocks go back 1 hour.
 
So this discussion has highlighted some other cases we need to explain in the specification.
- The timeouts are durations as you mention, but they result in some calculated end time. In relation to which system do they get calculated? (the participant or in case there is a central component)
- In which frame of reference are those 'end times' calculated. A difference in timezone (but also using the Buddhist Calendar for example) will not result in the expected outcome. And what to do with the switches from Winter to Summer Time and vice versa.
We need to mention that, although an absolute synchronization between clocks is not possible, they need to be 'aligned'. When a participant wants to join a transaction but the clock has a difference with the other services, it is possible that joining will fail as according to this service the LRA is already timed-out.

I'll create another ticket for the above 'timing' issues.

rdebusscher: What if I add a sentence that the clock values are from the users perspective and that no absolute clock  synchronization is assumed between the services?

Would that be OK?

mmusgrov: Not really. You are adding complexity where none is necessary.

A service joins and a timer is started. Similarly on a second server. Here it is irrelevant whether or not two clocks read different times since the implementation just needs to wait for a period to elapse and the two periods should be identical since all practical clocks tick at the same rate.

The only circumstance where we would ever need to use a time base is for crash recovery purposes and the use of UTC time would side step any concerns over timezones.

What is wrong with the simple short sentence that I suggested?

tomjenkinson: I don't think a link was provided here so the ticket that Rudy created is this one I think: https://github.com/eclipse/microprofile-lra/issues/237 (I added a comment on there)

api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
rdebusscher: Updated

spec/src/main/asciidoc/microprofile-lra-spec.adoc
rdebusscher: >What is wrong with the simple short sentence that I suggested?

It doesn't make it very clear that although a participant has defined a timeout, it can be overruled when within the same LRA another endpoint is called with another timeout.

But removed example and absolute timings and added the simple sentence.

PR #254, opened 2019-11-14
issue249 Replace short and long delays with calls to LRARecoveryService

### LINKED ISSUE #249: TCK: Replace short and long delays with calls to LRARecoveryService
Long delays are present because of repeated recovery calls. This can now be replaced by calls to LraRecoveryService.

The use of short delays in the TCK should also be reviewed with the aim of providing a less brittle mechanism.

xstefank:
Also, clean up unused replayEndPhase recovery handling.

The TckRecoveryTests require that the LRARecoveryService is injectable through the Arquillian resource injection. So it needs to be a CDI bean and be injectable be Arquillian.

resolves #249 

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
@@ -1,0 +1,85 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.lra.tck.service.spi;
+
+import java.net.URI;
+import java.util.logging.Logger;
+
+/**
+ *     The LRA model guarantees eventual consistency but does not say when
+ *     participants will be brought into a consistent state. When a request
+ *     to close/cancel an LRA is received, the implementation is allowed to
+ *     delay sending out the complete/compensate callbacks (the specification
+ *     only requires that an implementation must eventually send them).
+ *
+ *     This SPI provides a mechanism for an implementation to indicate when
+ *     it knows that callbacks have been sent and responses received or when
+ *     callbacks have been issued, all participants have reached an end state,
+ *     and when all listener notifications have been successfully delivered.
+ *
+ */
+public interface LRARecoveryService {
+
+    /**
+     * Wait for the delivery of Complete and Compensate participant callbacks.
+     * When this method returns the caller can be certain that the callbacks were
+     * sent and responses received (including error responses).
+     *
+     * @param lraId the LRA context
+     * @throws LRACallbackException the implementation was unable to determine whether
+     * or not the callbacks were received by all participants
+     */
+    void waitForCallbacks(URI lraId) throws LRACallbackException;
+
+    /**
+     * Wait for all participants to reach an end state and for all
+     * {@link org.eclipse.microprofile.lra.annotation.AfterLRA} notifications to be successfully delivered.
+     * 
+     * The default implementation iterates {@link LRARecoveryService#waitForEndPhaseReplay(URI)} until
+     * all participants reach a final state and for all listeners notifications to be successfully delivered.
+     *
+     * @param lraId the LRA context
+     * @throws LRACallbackException the implementation was unable to determine whether
+     * or not all participants have reached an end state and whether or not all listeners have been
+     * successfully notified
+     */
+    default void waitForRecovery(URI lraId) throws LRACallbackException {
+        Logger log = Logger.getLogger(LRARecoveryService.class.getName());
+        int counter = 0;
+
+        do {
+            log.info("Recovery attempt #" + ++counter);
+        } while (!waitForEndPhaseReplay(lraId));
+        log.info("LRA " + lraId + "has finished the recovery");
+    }
+
+    /**
+     * Wait for one replay of the end phase of the LRA (the calls to Status, Complete, Compensate, and Forget
+     * methods of all Compensating/Completing participants. If the LRA is finished as a result of this call
+     * all listeners must also be successfully notified before this method returns.
+     * 
+     * @param lraId the LRA context
+     * @return true if after the recovery all participants reached an end state and all AfterLRA listeners were
+     * successfully notified, false otherwise
+     * @throws LRACallbackException the implementation was unable to determine whether
+     * or not the callbacks were received by all participants
+     */
+    boolean waitForEndPhaseReplay(URI lraId) throws LRACallbackException;
+}
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
@@ -50,10 +50,11 @@ public interface LRARecoveryService {
 
     /**
      * Wait for all participants to reach an end state and for all
-     * {@link org.eclipse.microprofile.lra.annotation.AfterLRA} notifications to be successfully delivered.
-     * 
+     * {@link org.eclipse.microprofile.lra.annotation.AfterLRA} notifications to be successfully delivered
+     * (AfterLRA methods return HTTP 200).
+     *
      * The default implementation iterates {@link LRARecoveryService#waitForEndPhaseReplay(URI)} until
-     * all participants reach a final state and for all listeners notifications to be successfully delivered.
+     * all participants reach a final state and all AfterLRA listeners notifications are successfully delivered.
      *
      * @param lraId the LRA context
      * @throws LRACallbackException the implementation was unable to determine whether
@@ -71,15 +72,19 @@ public interface LRARecoveryService {
     }
 
     /**
-     * Wait for one replay of the end phase of the LRA (the calls to Status, Complete, Compensate, and Forget
+     * Wait for one replay of the end phase of the LRA (the callback calls to Status, Complete, Compensate, and Forget
      * methods of all Compensating/Completing participants. If the LRA is finished as a result of this call
-     * all listeners must also be successfully notified before this method returns.
-     * 
+     * all listeners must also be successfully notified before this method returns. The callback calls must be attempted
+     * but do not have to be successful (e.g. implementation tries to call Compensate which returns connection refused
+     * is a valid invocation of this method).
+     *
      * @param lraId the LRA context
-     * @return true if after the recovery all participants reached an end state and all AfterLRA listeners were
-     * successfully notified, false otherwise
-     * @throws LRACallbackException the implementation was unable to determine whether
-     * or not the callbacks were received by all participants
+     * @return true if the implementation successfully issued callback requests and received responses that indicate
+     * that final state was achieved and all AfterLRA listeners were successfully notified (AfterLRA methods returned
+     * HTTP 200), or false if the implementation successfully issued callback requests but it did not receive all
+     * responses or the received responses indicate that the final state is not reached yet
+     * @throws LRACallbackException the implementation has no knowledge of this LRA or it was unable to retry
+     * the requests to all participants so it does not make sense to trigger this method with the same argument again
      */
     boolean waitForEndPhaseReplay(URI lraId) throws LRACallbackException;
 }
--- a/tck/running_the_tck.asciidoc
+++ b/tck/running_the_tck.asciidoc
@@ -109,6 +109,12 @@ but that provide a way to deploy and undeploy deployments. That container to be 
 <arquillian>
 ----
 
+== SPI
+
+* The TCK provides an SPI `LRARecoveryService` which is retrived through the
+`ServiceLoader` mechanism. The implmenter is required to provide an
+implementation of this interface in order to run the TCK.
+
 == Debugging tests
 
 Debugging is dependent on the TCK implementor. The implementor configures Arquillian to use particular runtime
@@ -116,8 +122,3 @@ to deploy the test deployments and run the LRA client side implementation there.
 the runtime needs to define java debug properties `-Xdebug -Xrunjdwp:transport=dt_socket,address=8787,server=y,suspend=y`.
 Then debugger may be connected to port 8787 and to track the test execution.
 
-== Additional notes
-
-* `TckTestBase#replayEndPhase(String)` method, used for instance in `@Status` and 
-`@Forget` call verifications, expects that recovery happens on all active LRAs in 
-a reasonable time frame defined by the configuration property `lra.tck.recovery.timeout`.
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
@@ -24,8 +24,6 @@ import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
 
 import org.eclipse.microprofile.config.inject.ConfigProperty;
-import org.eclipse.microprofile.lra.annotation.LRAStatus;
-import org.eclipse.microprofile.lra.annotation.ParticipantStatus;
 
 @ApplicationScoped
 public class LraTckConfigBean {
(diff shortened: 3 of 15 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,51 @@
...
+ * define what actions can be taken to make the recovery (replay of the Compensate, Complete or Status) calls happen
+ * when the TCK requires to perform such actions {@link LRARecoveryService#waitForRecovery(URI)}. This allows to not
+ * wait for the mechanisms like periodic recovery and thus it makes the TCK runs faster.
+ */
+public interface LRARecoveryService {
+
+    /**
+     * Waits for the delivery of all participant callbacks (Complete, Compensate, AfterLRA, ...). When
+     * this method returns the results are verified so the implementation must make sure that the method
+     * doesn't return before the validation can happen.
```
mmusgrov: I think I understand what the javadoc is saying but only because we have discussed it elsewhere on the issues.
A reader who does not have that context would struggle to understand how he/she should implement these SPI methods.

rdebusscher:
@xstefank 

>So it needs to be a CDI bean and be injectable be Arquillian.

Did you not mention that there was an issue for a certain test that it needed to be a CDI bean. Was ServiceDiscovery not easier/good enough?

xstefank:
@rdebusscher yes, and I can reimplement to use ServiceLoader but for now I just stuck to just requiring that you will define your LRARecoveryService as CDI bean and that you also include arquillian resource provider for it. I agree that ServiceLoader will be easier from the point of implementation and it's also aligned with other specs. I will fix this. Thanks.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,80 @@
...
+import java.net.URI;
+
+/**
+ * This SPI is providing the implementation with the ability to define what actions need to be taken
+ * in case the TCK needs to wait for the the delivery of of participant callbacks (Complete, Compensate, ...) when the
+ * LRA is asked to perform some eventually consistent action {@link LRARecoveryService#waitForCallbacks(URI)} and to
+ * define what actions can be taken to make the recovery (replay of the Compensate, Complete or Status) calls happen
+ * when the TCK requires to perform such actions {@link LRARecoveryService#waitForRecovery(URI)}.
+ */
+public interface LRARecoveryService {
```
rdebusscher: I read this doc as LRARecoveryService being passive (except for the examples). It just wait until the callbacks are done.

But If I remember correct, the the idea of the LRARecoveryService was that it actively forced the callbacks to be performed faster (by poking some internal methods so that callbacks are done)  (at least for the waitForRecovery)

The waitForCallback is indeed just 'waiting' for a short period so that implementation can deliver the callbacks and they are executed before the test method verifies if those callbacks are executed.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,80 @@
...
+     *   completion notification which implies the participant is Completing</li>
+     *   <li>call to {@link LRARecoveryService#waitForRecovery(URI)} with the lraId -- trigger immediate replay
+     * of the Complete (or Status) call and wait for its delivery</li>
+     *   <li>call to {@link LRARecoveryService#waitForRecovery(URI)} returns</li>
+     *   <li>the TCK test verifies that Complete (or Status) have been called (retried)</li>
+     * </ol>
+     * 
+     * @param lraId the LRA id of the LRA which should be recovered when this method exits
+     */
+    void waitForRecovery(URI lraId);
```
rdebusscher: so `waitForRecovery` is not a good name is it actively needs to perform things.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
rdebusscher: This SPI is also allowed to throw Exceptions in case it detects that waiting or forcing recovery is not possible.

That way the TCK test also fails and covers for the case that it should wait indefinitely for a callback to happen which is not going to be performed.

xstefank: good points. I will add this.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
xstefank: can (and should) but doesn't have to. Also long delay would work here.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,88 @@
...
+     *   <li>call to {@link LRARecoveryService#waitForCallbacks(URI)} with lraId -- wait for all Complete calls to be
+     * delivered and optionally perform some implementation specific actions to make the calls return faster</li>
+     *   <li>call to {@link LRARecoveryService#waitForCallbacks(URI)} returns</li>
+     *   <li>the TCK test verifies that all Complete methods have been called</li>
+     * </ol>
+     *
+     * @param lraId the LRA id of the LRA which callbacks should be delivered when this method exits
+     * @throws Exception in case of any error. The exception will be propagated to fail the enclosing TCK test.
+     */
+    void waitForCallbacks(URI lraId) throws Exception;
```
mmusgrov: Can we use a specific exception to indicate that callbacks were not successful and in the tests catch it and assert fail. Using Exception will produce stack traces in the tests and make it harder to differentiate other TCK exceptions from callback failures.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,88 @@
...
+     *   <li>call to {@link LRARecoveryService#waitForRecovery(URI)} returns</li>
+     *   <li>the TCK test verifies that Complete (or Status) have been called (retried)</li>
+     * </ol>
+     * 
+     * @param lraId the LRA id of the LRA which should be recovered when this method exits
+     * @throws Exception in case of any error. The exception will be propagated to fail the enclosing TCK test.
+     *              
+     */
+    void waitForRecovery(URI lraId) throws Exception;
+}
```
mmusgrov: I am finding it difficult to parse the javadoc. Your first attempt was better, I took that a varied it a little- how about the following alternative:

```java
/**
 *     The LRA model guarantees eventual consistency but does not say when
 *     participants will be brought into a consistent state. When a request
 *     to close/cancel an LRA is received, the implementation is allowed to
 *     delay sending out the complete/compensate callbacks (the specification
 *     only requires that an implementation must eventually send them).
 *
 *     This SPI provides a mechanism for implementations to indicate when
 *     it knows that callbacks have been sent and responses received or when
 *     callbacks have been issued and all participants have reached an end state.
 *
 */
public interface LRARecoveryService {
    /**
     * Wait for the delivery of all participant callbacks (Complete or Compensate).
     * When this method returns the caller can be certain that the callbacks were
     * sent and responses received (including error responses).
     *
     * @param lraId the LRA context
     * @throws LRACallbackException the implementation was unable to determine whether
     * or not the callbacks were received by all participants
     */
    void waitForCallbacks(URI lraId) throws CallbackException;

    /**
     * Waits for all participants to reach an end state.
     * Question do we also want the method to wait for any AfterLRA notifications to be successfully delivered?
     *
     * @param lraId the LRA context
     * @throws LRACallbackException the implementation was unable to determine whether
     * or not all participants have reached an end state.
     */
    void waitForRecovery(URI lraId) throws CallbackException;
}
```

xstefank: so I see, we had a misunderstanding - with waitForRecovery you wanted to change the definition  of what triggerRecovery does, not just a rename. Ok, I will use this javadoc and fix your other comment.

mmusgrov: >     /**
>      * ...
>      * Question do we also want the method to wait for any AfterLRA notifications to be successfully delivered?

@xstefank Did you also see my question in the javadoc. Is that something we need and if so can it be implicit in the `waitForRecovery` method?

xstefank: yes, I included also a requirement for afterLRA to be delivered.

xstefank: @mmusgrov but if we are going with your definition (Waits for all participants to reach an end state) we are unable to test intermediate states of LRAs -- check the testForget() TCK test, please.

mmusgrov: So go with some similar to [one of my earlier suggestions](https://github.com/eclipse/microprofile-lra/issues/249#issuecomment-549723582).

mmusgrov: > yes, I included also a requirement for afterLRA to be delivered.

We need good javadoc so that it is clear what the implementer's responsibilities are.

mmusgrov: > @mmusgrov but if we are going with your definition (Waits for all participants to reach an end state) we are unable to test intermediate states of LRAs -- check the testForget() TCK test, please. So, are we going back to triggerRecovery() or should I change the tests?

I did not fully understand your javadoc so was unable to determine what the responsibilities were. That is why I asked if I could propose the alternative text.

Can you explain what we need in order to verify the `testForget()` TCK test.

xstefank: it would be better to check the code directly. testForget closes LRA which goes into Closing state, then we "triggerRecovery" so the Status method is invoked. If we now block until the participant reaches the final state, this test will never end. The state is cleared after som asserts and then the recovery is triggered again. So we can either have `triggerRecovery(URI)` and `waitForRecovery(URI)` or we can replace all `waitForRecovery(URI)` calls with `triggerRecovery(URI); waitForCallbacks(URI);`. Or of course we can change this test, but then we will be always be able to test only end states which I guess is not something we want.

mmusgrov: So don't block

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,84 @@
...
+ *     only requires that an implementation must eventually send them).
+ *
+ *     This SPI provides a mechanism for implementations to indicate when
+ *     it knows that callbacks have been sent and responses received or when
+ *     callbacks have been issued and all participants have reached an end state.
+ *
+ */
+public interface LRARecoveryService {
+
+    Logger LOGGER = Logger.getLogger(LRARecoveryService.class.getName());
```
mmusgrov: Would it be possible to move this into the default implementation since an implementation may not need to log (eg it could use callbacks instead of polling) or may use a different logging framework.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,84 @@
...
+        int counter = 0;
+
+        do {
+            LOGGER.info("Recovery attempt #" + ++counter);
+        } while (!waitForEndPhaseReplay(lraId));
+        LOGGER.info("LRA " + lraId + "has finished the recovery");
+    }
+
+    /**
+     * Wait for one replay of the end phase of the LRA (the calls to Status, Complete, or Compensate methods
```
mmusgrov: There is also `@Forget`

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
xstefank: The implementation can always implement this method. It only makes sense, because this is expected behavior. I can remove logging or just replace it with System.out.println?

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
xstefank: +1

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,84 @@
...
+     * of all Compensating/Completing participants once) of a particular LRA. If all participants after these calls
+     * reach an end state, all {@link org.eclipse.microprofile.lra.annotation.AfterLRA} methods for this LRA are
+     * also expected to be called before this method returns.
```
mmusgrov: If any listener returns anything other that a 200 status then the spec says that the implementation must retry the notification later. In this case does the `waitForEndPhaseReplay` method return after the first failed attempt to send the listener notification. And if the TCK test calls the method again will it reattempt delivery of the AfterLRA notification?

xstefank: Yes. This would be the same as for Status, Complete, Compensate. I will adjust the text to make it clear.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
mmusgrov: meant to move the `getLogger` call into the default method so it would like:

``` java
    default void waitForRecovery(URI lraId) throws LRACallbackException {
         Logger LOGGER = Logger.getLogger(LRARecoveryService.class.getName());
         int counter = 0;

        do {
            LOGGER.info("Recovery attempt #" + ++counter);
        } while (!waitForEndPhaseReplay(lraId));
        LOGGER.info("LRA " + lraId + "has finished the recovery");
    }
```

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,83 @@
...
+/**
+ *     The LRA model guarantees eventual consistency but does not say when
+ *     participants will be brought into a consistent state. When a request
+ *     to close/cancel an LRA is received, the implementation is allowed to
+ *     delay sending out the complete/compensate callbacks (the specification
+ *     only requires that an implementation must eventually send them).
+ *
+ *     This SPI provides a mechanism for implementations to indicate when
+ *     it knows that callbacks have been sent and responses received or when
+ *     callbacks have been issued and all participants have reached an end state.
```
mmusgrov: "and when all listener notifications have been successfully delivered."

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,83 @@
...
+ *
+ */
+public interface LRARecoveryService {
+
+    /**
+     * Wait for the delivery of all participant callbacks (Complete or Compensate).
```
mmusgrov: This should not be all participant callbacks - just Complete and Compensate. The other method is for `all` callbacks.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param lraId the LRA context
+     * @throws LRACallbackException the implementation was unable to determine whether
+     * or not the callbacks were received by all participants
+     */
+    void waitForCallbacks(URI lraId) throws LRACallbackException;
+
+    /**
+     * Wait for all participants to reach an end state and for all
+     * {@link org.eclipse.microprofile.lra.annotation.AfterLRA} notifications to be delivered.
```
mmusgrov: "successfully delivered"

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,83 @@
...
+     * 
+     * The default implementation iterates {@link LRARecoveryService#waitForEndPhaseReplay(URI)} until
+     * all participants reach a final state.
```
mmusgrov: "and for all listeners notifications to be successfully delivered."

We either need a separate method to wait for listener notifications or we need make sure the javadoc accurately reflects the behaviour with respect to the listener callbacks.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,83 @@
...
+     *
+     * @param lraId the LRA context
+     * @throws LRACallbackException the implementation was unable to determine whether
+     * or not all participants have reached an end state.
```
mmusgrov: "and whether or not all listeners have been successfully notified"

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,83 @@
...
+
+        do {
+            log.info("Recovery attempt #" + ++counter);
+        } while (!waitForEndPhaseReplay(lraId));
+        log.info("LRA " + lraId + "has finished the recovery");
+    }
+
+    /**
+     * Wait for one replay of the end phase of the LRA (the calls to Status, Complete, Compensate, and Forget
+     * methods of all Compensating/Completing participants and the possible repeated calls to AfterLRA methods for
```
mmusgrov: The text is vague with respect to the listener callbacks - does the method perform repeated listener notifications in the same pass or does the caller need to reissue the `waitForEndPhaseReplay` repeatedly until no exception is thrown or when true is returned?

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,83 @@
...
+     * finished LRAs) of a particular LRA.
+     * 
+     * @param lraId the LRA context
+     * @return true if after the recovery all participants reached an end state and all AfterLRA methods were called
+     * successfully (the LRA has ended), false otherwise
```
mmusgrov: The LRA has ended before the AfterLRA calls are issued (the text seems to imply that the LRA only ends when all AfterLRA notifications have been successfully delivered.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,83 @@
...
+     * @throws LRACallbackException the implementation was unable to determine whether
+     * or not the callbacks were received by all participants
+     */
+    boolean waitForEndPhaseReplay(URI lraId) throws LRACallbackException;
```
mmusgrov: Why does this method return a result whereas the other two interface methods do not

xstefank: so we can use it in `waitForRecovery`

mmusgrov: Okay

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,85 @@
...
+        int counter = 0;
+
+        do {
+            log.info("Recovery attempt #" + ++counter);
+        } while (!waitForEndPhaseReplay(lraId));
+        log.info("LRA " + lraId + "has finished the recovery");
+    }
+
+    /**
+     * Wait for one replay of the end phase of the LRA (the calls to Status, Complete, Compensate, and Forget
```
rdebusscher: It is not entirely clear for me, also the difference between `waitForRecovery` and `waitForEndPhaseReplay`.

this `waitForEndPhaseReplay` issues the callbacks and wait for a response from them.  In which cases does it return false?

- When the callback returned an error (say status 500)
- When the callback returns an 'in progress' status (like ParticipantStatus.compensating)

So correct me if I'm wrong, but it should only return false in the case the end state is not yet achieved yet, but it makes sense to call this method again (which is what `waitForRecovery` does)

If it doesn't make sense to call the method again

- the callback returned an error (say status 500)
- the callback returned a failed state (ParticipantStatus.FailedToCompensate)
- ...

It must throw an exception.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param lraId the LRA context
+     * @throws LRACallbackException the implementation was unable to determine whether
+     * or not the callbacks were received by all participants
+     */
+    void waitForCallbacks(URI lraId) throws LRACallbackException;
+
+    /**
+     * Wait for all participants to reach an end state and for all
+     * {@link org.eclipse.microprofile.lra.annotation.AfterLRA} notifications to be successfully delivered.
```
rdebusscher: AfterLRA needs to be delivered AND response is HTTP Status 200 (since they can be resend if not the ok status is received)

xstefank: I covered with "successfully" but I will add this to parentheses after that to make it clear.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
xstefank: yes, you got everything correct. If exception is thrown this is unrecoverable state for the SPI (can by handled in TCK test), false means that one round of callbacks was issued but LRA is still not finished (202, or error from one or more of the callbacks) and true means that after the round of callbacks the LRA is in the end state (Closed, Cancelled, FailedTo*). Do you think we should adjust something in the interface to make it more clear?

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
rdebusscher: ok, I see. I interpreted successfully delivered as the call was successfully performed, but it was not checking about the response.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/spi/LRARecoveryService.java
rdebusscher: I would add the above examples to the javadoc so that each implementor knows exactly what it should do in each case. Probably most people will 'guess' it correctly, as I did, but it putting this explicitly in the text is probably easier for implementors.
Thx

xstefank: I hit send too soon.

> If it doesn't make sense to call the method again

> the callback returned an error (say status 500)

No, and I can see how this is misleading. We are testing this scenario in TckRecoveryTests so I would prefer to just return false in this case and keep the exception for really exceptional cases.

> the callback returned a failed state (ParticipantStatus.FailedToCompensate)

No, the LRA reached a final state (FailedToCancel) to it should return true

Issue #237, opened 2019-09-24
Spec: Clarify some requirements around 'time'


rdebusscher:
- The timeouts are durations and result in some calculated end time. In relation to which system do they get calculated? (the participant or in case there is a central component) And who is responsible for determining that a LRA needs to be cancelled due to timeout (as different systems in the environment can have different clocks)
- In which frame of reference are those 'end times' calculated (is Unix epoch usable?). A difference in timezone (but also using the Buddhist Calendar for example) will not result in the expected outcome. And what to do with the switches from Winter to Summer Time and vice versa.
- We need to mention that, although an absolute synchronization between clocks is not possible, they need to be 'aligned'. When a participant wants to join a transaction but the clock has a difference with the other services, it is possible that joining will fail as according to this service the LRA is already timed-out.

tomjenkinson:
I think that text clarification is good here but what I would say is that the clarification shouldn't define the implementation technique and so the wording would be about not relying on these timeouts as so-to-speak alarm clocks but rather a hint to the `LRA`. As you point out, there is no requirement for a central component but the LRA is an entity in the system and so the timeout should probably only be specified in relation to that.

In other words if a participant says "I can do a compensation in the next 1 hour", it is just a hint that it should be available to do that up until 1 hour but even in the case that you do manage to call it back within the next hour it might still have a compensation failure so it's not guaranteed it would be able to do that anyway.

AIUI the timelimit is there for a participant to say, in 1 hour (or whatever) I am going to be compensating so you (i.e. the LRA) might as well tell everyone else to compensate too as there is no point in asking me to complete in 61 minutes. That should be made clear in the text.

mmusgrov:
> * The timeouts are durations and result in some calculated end time. In relation to which system do they get calculated? (the participant or in case there is a central component) And who is responsible for determining that a LRA needs to be cancelled due to timeout (as different systems in the environment can have different clocks)

Since we are talking about timeouts, the value read from different clocks is irrelevant as long as they tick at the same rate. The implementation just needs to know when an interval of time has elapsed, i.e. it does not need to calculate an end time unless it needs to persist the timer (see next comment).

> * In which frame of reference are those 'end times' calculated (is Unix epoch usable?). A difference in timezone (but also using the Buddhist Calendar for example) will not result in the expected outcome. And what to do with the switches from Winter to Summer Time and vice versa.

[...]

PR #257, opened 2019-11-19
issue 237 Clarify some requirements around 'time'

mmusgrov:
#237

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -1028,7 +1028,7 @@ example, to indicate that an LRA should automatically cancel after 100 milliseco
 
 Furthermore, the ability to compensate or complete may be transient capabilities of a service so participants can also be timed out.
 When the timeout is reached, the LRA moves to the state `Cancelling` and all registered participants receive a `Compensate`
-call. The same actions are performed then the case where the LRA is canceled.
+call. The same actions are performed then the case where the LRA is cancelled.
 
 When a participant joins an existing transaction using the `@LRA(value = ...)` annotation, the timeout of
 the already active Long Running Action can be influenced by the `timeLimit` and `timeUnit` defined on this annotation.
@@ -1037,6 +1037,51 @@ the request enters the JAX-RS method and the timeout information found on the `@
 the moment calculated for the LRA at that time, this new moment becomes the time out moment for the LRA.
 So when multiple participants define a timeout period, the earliest one will trigger the cancellation of the LRA.
 
+Since the specification targets microservices running in different JVM's there are
+potential issues around how timing constraints are handled:
+
+A timeout is a duration which results in some calculated end time. If different systems
+use different timezones or clock granularity then there is potential for different parts
+of the system to have different views of the current time.  Therefore implementations
+must be careful to avoid using absolute times and be aware that absolute synchronisation
+between clocks is not possible, and therefore, where possible, rely on a duration instead.
+If an absolute time is required then the recommendation is to use UTC which is a universal
+time standard to which all timezones can be canonically converted. This will provide some
+degree of alignment between clocks (where alignment means that clocks are only
+approximately globally synchronised).
+
+Potential reasons for requiring absolute time values, as opposed to durations, include:
+
+1. When a failed server is restarted the implementation [of this specification] needs to
+   calculate how long is left for timed LRA's to become eligible for cancellation based on 
+   on how much time elapsed before the server failed and how much time elapsed before the
+   failed server was restarted. Such calculations can only be made if the absolute time of
+   expiry of the LRA is known.
+2. In a centralised coordinator based system, if another server takes control of the LRA
+   then the absolute time would be required.
+3. One could envisage a non centralised implementation that relies on absolute time values
+   to determine when LRA's may be cancelled. In such an approach it is even more important
+   that different clocks are aligned. Failure to align clocks in such a system would
+   result in higher failure rates for LRA's.
+
+
+If there are edge cases around timing of actions or if local clocks
+cannot be relied upon then there are risks that different parts of the system may
+perform conflicting actions. The implementation must strive to minimise these risks
+but if they do occur then it MUST detect the conflict and force the LRA into
+<<lra-state-model,one of the failure states>> and it MUST report the conflict
+(the actual mechanism by which it does so is unspecified).
+
+Also note that time limits are never absolute with respect to when LRA's are actually
+cancelled: they are hints to the system that if the time limit is breached then
+compensation may not be possible and an implementation should move the state of the
+LRA to cancelling. It is also possible for an implementation to miss a deadline
+(for whatever reason) and some other part of the system cancels the LRA before it
+notices. In this scenario there would be a potential conflict since one of the
+participants is no longer able to compensate. But as mentioned in the introduction to
+this specification, failure states are permissible but "they must be logged or
+flagged for the administrator and manual intervention may be necessary to restore
+an application’s consistency".
 
 [[leaving-an-lra]]
 ==== Leaving an LRA


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1020,6 +1020,30 @@ the request enters the JAX-RS method and the timeout information found on the `@
...
+Since the specification targets microservices running in different JVM's there are
+potential issues around how timing constraints are handled:
+
+A timeout is a duration which results in some calculated end time. If different systems
+use different timezone or clock granularity then there is potential for different views
+about the current time.  Therefore implementations must be careful to avoid using absolute
+times and be aware that absolute synchronisation between clocks is not possible, and
+therefore, where possible, rely on a duration instead. If an absolute time is required
+then the recommendation is to use UTC which is a universal time standard to which
+all timezones can be canonically converted. This will provide some degree of alignment
```
xstefank: I may be missing the idea in this paragraph but we allow only durations in LRA annotation. Can you remind me why do we need to describe absolute times, please?

mmusgrov: > I may be missing the idea in this paragraph but we allow only durations in LRA annotation. Can you remind me why do we need to describe absolute times, please?

When a crashed server is restarted the implementation must calculate when timed LRAs become eligible for cancellation.

rdebusscher: With `crashed server` do you mean the implementation or the micro-service?

For me, the 'time-out' time is calculated at the moment a participant joins and monitored all the time to see if LRAs become eligible for cancellation. (and callbacks needs to be sent out)

When another participant joins, the time-out needs to be verified and possible redefined (when participant defines also a duration which might be earlier then the current end time).

So I see a role for the UTC time there, which doesn't has anything to do with a restart of a crashed implementation or restart of micro-service.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1020,6 +1020,30 @@ the request enters the JAX-RS method and the timeout information found on the `@
...
+between clocks. If there are edge cases around timing of actions or if local clocks
+cannot be relied upon then there are risks that different parts of the system may
+perform conflicting actions. The implementation must strive to minimise these risks
+but if they do occur then it MUST detect the conflict and force the LRA into
+<<lra-state-model,one of the failure states>>.
+
+Note that time limits are never absolute, they are hints to the system that if the
+time limit is breached then compensation may not be possible and an implementation
+should move the state of the LRA to cancelling. Note that if a time limit is
+breached and the system begins asking some participants to complete before it
```
rdebusscher: >and the system begins asking some participants to complete before it notices then there is potential for conflicts.

Unclear sentence for me.

xstefank:
I still don't think it's necessary to have this paragraph in the spec (but of course it's just more reading) because this is implementation concern.

>If time on Server B is 2 minutes later then on Server A (due to a wrong clock setting). If Participant A (running on Server A), start a LRA with timeout of 1 minute, and then calls Participant B (running on server B) it cannot join the LRA anymore because according to Server B the LRA is already timed out.

I think this is not correct. Participant on Server B can normally join as LRA is still active. The enlisment has nothing to do with Service B local clock. Even if the Service B send a new timelimit on enlistment, this will be only duration that will be compared to timeout time set at Server A.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
tomjenkinson: I agree, it's not clear what the meaning is to me either. I assume it is pointing out that the participant may have made a unilateral decision with prevents a consistent outcome, rather than some confusion in the LRA implementation.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
tomjenkinson: I believe this case is referring to a situation where all actors in the system are shutdown while the LRA is still active. In order that the system can rehydrate successfully it needs to know the expected time that the LRA was expected to automatically cancel.

rdebusscher:
@xstefank

>I think this is not correct. Participant on Server B can normally join as LRA is still active. 

 In a central system like Naraya, it is probably the case, but not in a distributed system as with Payara implementation. There it is the implementation code running on Server B which makes the decision if the LRA is still active.

rdebusscher:
@tomjenkinson 

>clock granularity

For me granularity is the precision, not the correctness.

xstefank:
> In a central system like Naraya, it is probably the case, but not in a distributed system as with Payara implementation. There it is the implementation code running on Server B which makes the decision if the LRA is still active.

Ok. But isn't that then an implementation concern? not something we need to have in the specification? If we eventually define interoperability the timeout format MUST be defined precisely so different implementation can pick up time after some implementation that first started it. In the current state, Payara should be able to align its timing if I understand correctly. So would this paragraph help somebody who is reading/implementing the specification?

rdebusscher:
@xstefank 

>But isn't that then an implementation concern? not something we need to have in the specification?

If you see a spec text as something which should only be read by implementors, then yes it should not be mentioned.

But for me, specs are also for the developers who want to learn how to use it and what they should do.

mmusgrov:
> I still don't think it's necessary to have this paragraph in the spec (but of course it's just more reading) because this is implementation concern.
> 
> > If time on Server B is 2 minutes later then on Server A (due to a wrong clock setting). If Participant A (running on Server A), start a LRA with timeout of 1 minute, and then calls Participant B (running on server B) it cannot join the LRA anymore because according to Server B the LRA is already timed out.
> 
> I think this is not correct. Participant on Server B can normally join as LRA is still active. The enlisment has nothing to do with Service B local clock. Even if the Service B send a new timelimit on enlistment, this will be only duration that will be compared to timeout time set at Server A.

+1 This test case would imply that the spec implementation is incorrect. When an LRA begins to close or cancel any attempt to join with it SHOULD fail. If it does not fail then it is a protocol violation and MUST be logged when the implementation discovers the violation.

I think I see where your issues are coming from: are you are assuming that each node (where the spec implementation is operating) is working independently from other nodes with no agreement between them. If there is no agreement/coordination between the different parts of the system then you will see more conflicts for the outcome of the LRA and it will be difficult (if not impossible) to report protocol violations which would also be a spec violation.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: You are right, I will reword it (I should have used cancel not close). The sentence was meant to refer to the implementation not noticing that the time limit has passed (for whatever reason). If is then asked to cancel the LRA then that would potentially conflict with the participant that said it will be unable to compensate.

rdebusscher:
>+1 This test case would imply that the spec implementation is incorrect. When an LRA begins to close or cancel any attempt to join with it SHOULD fail.

Payara implementation will not allow that Participant B join as a participant since according to the clock of Server B, LRA should be timed out.

But due to the major time difference in clocks of both servers, the application will not behave as expected.

But if I'm the alone one who wants to warn the developers/people about this kind of issue, then it is fine for me. We need to put those things then in the implementation documentation only.

xstefank:
> If you see a spec text as something which should only be read by implementors, then yes it should not be mentioned.
> 
> But for me, specs are also for the developers who want to learn how to use it and what they should do.

@rdebusscher Sorry Rudy, but I don't understand. Developers will probably not have direct access to influence any timing issues. All the developer can do is to pass a duration in the LRA annotation. This kind of issue should be transparent to the users I believe. But we don't need to throw it away, we can add something like additional notes section or similar.

rdebusscher:
>Developers will probably not have direct access to influence any timing issues.

They can set the clocks of the different machines.

But as already mentioned, we will only put a remark in the documentation of the implementors about the clock alignment if that is an issue for that implementation. (and not in the spec)

xstefank:
Ok, I see we are looking differently at it. I agree that we can close this then.

mmusgrov:
> > +1 This test case would imply that the spec implementation is incorrect. When an LRA begins to close or cancel any attempt to join with it SHOULD fail.
> 
> Payara implementation will not allow that Participant B join as a participant since according to the clock of Server B, LRA should be timed out.
> 
> But due to the major time difference in clocks of both servers, the application will not behave as expected.
> 
> But if I'm the alone one who wants to warn the developers/people about this kind of issue, then it is fine for me. We need to put those things then in the implementation documentation only.

@rdebusscher 

I can add more text although the PR does state that "... This will provide some degree of alignment between clocks".

So should I say something like  "... if the clocks are not aligned and an implementation requires global clock synchronisation then failure rates may be higher than expected", or something to that effect?

rdebusscher:
@mmusgrov 

>if the clocks are not aligned and an implementation requires global clock synchronisation then failure rates may be higher than expected

That sound good to me

mmusgrov:
@rdebusscher @xstefank @tomjenkinson This one is ready for re-review.

PR #259, opened 2019-11-28
Define config property to define propagation

### LINKED ISSUE #201: Spec: About propagation of active context
The following is defined in the spec.

>The implementation MUST propagate the `active context` on outgoing JAX-RS
requests.

>If an LRA is propagated to a resource that is not annotated with any of the
annotations defined in this specification then the LRA will not be made
available to the invoked method. However, any outgoing JAX-RS invocations
made during the execution of the method will still carry the original context.

rdebusscher:
by not annotated JAX-RS resources. Fixes #201

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
@@ -72,10 +72,10 @@ import java.time.temporal.ChronoUnit;
  *
  * <p>
  * If an LRA is propagated to a resource that is not annotated with any
- * particular LRA behaviour then the LRA will be suspended (ie the context
- * will not be available to the resource). But if this resource
- * then performs an outgoing JAX-RS request then the suspended LRA must be propagated
- * on this second outgoing request. For example, suppose resource <code>A</code>
+ * particular LRA behavior, then the value of the <code>mp.lra.LRA.value</code> configuration parameter
+ * determines how the LRA is propagated. The default of this parameter is <code>SUPPORTS</code> which
+ * means that the LRA must be propagated to outgoing requests.
+ * For example, suppose resource <code>A</code>
  * starts an LRA and then performs a JAX-RS request to resource <code>B</code> which
  * does not contain any LRA annotations. If resource <code>B</code> then performs a
  * JAX-RS request to a third service, <code>C</code> say, which does contain LRA
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
@@ -72,8 +72,8 @@ import java.time.temporal.ChronoUnit;
  *
  * <p>
  * If an LRA is propagated to a resource that is not annotated with any
- * particular LRA behavior, then the value of the <code>mp.lra.LRA.value</code> configuration parameter
- * determines how the LRA is propagated. The default of this parameter is <code>SUPPORTS</code> which
+ * particular LRA behavior, then the value of the <code>mp.lra.propagation.active</code> configuration parameter
+ * determines how the LRA is propagated. The default of this parameter is <code>true</code> which
  * means that the LRA must be propagated to outgoing requests.
  * For example, suppose resource <code>A</code>
  * starts an LRA and then performs a JAX-RS request to resource <code>B</code> which
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -341,10 +341,9 @@ this identifier is passed as a method parameter.
 The implementation MUST propagate the `active context` on outgoing JAX-RS
 requests.
 
-If an LRA is propagated to a resource that is not annotated with `@LRA` then the implementation behaves
-as if there is an annotation present on the resource with the `type()` defined by the configuration parameter
-`mp.lra.LRA.value`. From a functional point of view, there is no difference notable except for the different
-default for `end()` which is false when there is no `@LRA` defined.
+If an LRA is propagated to a resource that is not annotated with `@LRA` then the implementation looks at the
+configuration parameter `mp.lra.propagation.active` to decide if the LRA context information in the header must
+be propagated with the outgoing requests (value evaluates to `true`, the default) or not (value evaluates to `false`).
 
 The user is allowed to manually assign the `context header` on JAX-RS requests
 and responses:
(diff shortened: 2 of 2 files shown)


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -305,6 +305,27 @@ that it could not until explicitly told that it can clean up by providing a meth
 
 If multiple methods are annotated with the same annotation an arbitrary one will be chosen.
 
+[[config-parameters]]
+==== Configuration parameters
+
+This specification has 1 configuration parameter which is retrieved using the MicroProfile Config principals.
```
xstefank: didn't you mean "principles"?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -305,6 +305,27 @@ that it could not until explicitly told that it can clean up by providing a meth
...
+
+     mp.lra.LRA.value=SUPPORTS
+
+The allowed values for the configuration parameter are `SUPPORTS`, the default value, and `NOT_SUPPORTED`.
```
xstefank: so if we allow only SUPPORTS/NOT_SUPPORTED wouldn't it be better to have only a boolean value mp.lra.propagation.active=true/false? Won't this lead developers to define all possible LRA.Type values?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -305,6 +305,27 @@ that it could not until explicitly told that it can clean up by providing a meth
...
+When an invalid value is defined, the implementation MUST prohibit the successful startup of the application
+(e.g. through a runtime exception).
+
+When a JAX-RS endpoint, or the containing class, is not annotated with @LRA, but it is called on a MicroProfile LRA compliant runtime,
+the system behaves as if there is an annotation present with the `type()` value taken from the configuration parameter.
```
xstefank: this is `value()` in the LRA annotation - https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L169

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -72,10 +72,10 @@
  *
  * <p>
  * If an LRA is propagated to a resource that is not annotated with any
- * particular LRA behaviour then the LRA will be suspended (ie the context
- * will not be available to the resource). But if this resource
- * then performs an outgoing JAX-RS request then the suspended LRA must be propagated
- * on this second outgoing request. For example, suppose resource <code>A</code>
+ * particular LRA behavior, then the value of the <code>mp.lra.LRA.value</code> configuration parameter
+ * determines how the LRA is propagated. The default of this parameter is <code>SUPPORTS</code> which
+ * means that the LRA must be propagated to outgoing requests.
```
mmusgrov: I agree with Martin's comment about using a boolean valued property here. If you agree then the property name needs changing.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: +1

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -305,6 +305,27 @@ that it could not until explicitly told that it can clean up by providing a meth
...
+==== Configuration parameters
+
+This specification has 1 configuration parameter which is retrieved using the MicroProfile Config principals.
+
+     mp.lra.LRA.value=SUPPORTS
+
+The allowed values for the configuration parameter are `SUPPORTS`, the default value, and `NOT_SUPPORTED`.
+When an invalid value is defined, the implementation MUST prohibit the successful startup of the application
+(e.g. through a runtime exception).
+
```
mmusgrov: So if we go with a boolean valued property there is no reason to prohibit startup of the application.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: So if we go with a boolean valued property then

> the system behaves as if there is an annotation present with the value SUPPORTS if the config property is `true` and NOT_SUPPORTED if it is `false`.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -305,6 +305,27 @@ that it could not until explicitly told that it can clean up by providing a meth
...
+When a JAX-RS endpoint, or the containing class, is not annotated with @LRA, but it is called on a MicroProfile LRA compliant runtime,
+the system behaves as if there is an annotation present with the `type()` value taken from the configuration parameter.
+
+Note however that these values, which correspond to the enum values of `org.eclipse.microprofile.lra.annotation.ws.rs.LRA.Type`
```
mmusgrov: If we go with a boolean valued property then this sentence needs to be reworded.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -323,10 +344,10 @@ this identifier is passed as a method parameter.
...
 requests.
 
-If an LRA is propagated to a resource that is not annotated with any of the
-annotations defined in this specification then the LRA will not be made
-available to the invoked method. However, any outgoing JAX-RS invocations
-made during the execution of the method will still carry the original context.
+If an LRA is propagated to a resource that is not annotated with `@LRA` then the implementation behaves
+as if there is an annotation present on the resource with the `type()` defined by the configuration parameter
+`mp.lra.LRA.value`. From a functional point of view, there is no difference notable except for the different
+default for `end()` which is false when there is no `@LRA` defined.
```
mmusgrov: If we go with a boolean valued property this text needs to be reworded.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
rdebusscher: yes agree. I proposed the SUPPORTS/NOT_SUPPORTED since we only discussed based on LRA.Type

spec/src/main/asciidoc/microprofile-lra-spec.adoc
rdebusscher: yes

spec/src/main/asciidoc/microprofile-lra-spec.adoc
rdebusscher: Indeed, I'll adjust

rdebusscher:
@mmusgrov @xstefank I updated PR to use a boolean config parameter.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: @rdebusscher this still references the old configuration parameter mp.lra.LRA.value.

mmusgrov: +1

Issue #242, opened 2019-10-14
Remove tests for active LRAs for non participating resources


mmusgrov:
There are two tests:

    TckLRATypeTests#mandatoryWithClosedLRA
    TckLRATypeTests#requiredWithClosedLRA

that check whether or not an implementation prohibits calling an `@LRA` annotated method (of a non-participating resource) if the LRA is no longer active (note that the tests do not register a participant). This behaviour is problematic for the following reasons:

  - it is unnecessary because it does not guarantee that an LRA will be valid when the JAX-RS method executes since an LRA can be closed at any time before and during the method invocation. The only way to guarantee validity of an LRA during a method invocation is to enlist with it. If the LRA is no longer active then the attempt to enlist with it will fail with 412 and that is good enough to write correct business logic.
  - it is impossible to implement correctly in a distributed environment for the previous reason (enlisting as a participant is the correct approach)
  - it incurs an unnecessary performance penalty due to the extra network call to get the status.
  - it is not a spec requirement (the closest text is in the javadoc for the LRA and we could add a note there to say that an LRA which is active before method entry can become inactive at an arbitrary point in time and that an attempt to enlist with an inactive LRA will fail although I think we already state this elsewhere).

NB the tests do make sense as long as the invoked resource is a participant.

xstefank:
+1 this makes sense. We can remove these two tests but we should add one testing join failure.

tomjenkinson:
+1

I would expect to see the join test added at the same time as it should be prevented. Would that need additional text too?

rdebusscher:
I can understand that the test TckLRATypeTests#requiredWithClosedLRA is incorrect and should be removed.

But TckLRATypeTests#mandatoryWithClosedLRA is testing something valid. When you annotated the JAX-RS method with @LRA(type=MANDATORY) than you need to call it with a valid LRA id in the header, otherwise HTTP Status 412 must be returned (defined in the JavaDoc for LRA.Type.MANDATORY)

rdebusscher:
>If an entity wants to know if the LRA is active then he must enlist as a participant

This is not required, as LRA party (by just using @LRA) you can also enforce things like REQUIRED, MANDATORY, etc ..)

> If he chooses not to enlist then he must be prepared to handle the consequences.

Do we have this somewhere in the spec? (explicitly, because this is a very important aspect of the spec)
[...]

PR #266, opened 2020-01-14
issue242 remove tests for active LRAs for non participating resources

mmusgrov:
#242


xstefank:
@rdebusscher both of them can be on the participant. So I will reintroduce them :). Merging this now.

PR #267, opened 2020-01-16
issue6 Timeout tests

mmusgrov:
#6 

The issue called for 3 different tests. I implemented the first one TckTests#timeLimitWithPreConditionFailed
The second test is very similar to TckTests#timeLimit so I have not added a one for that.
The third test that was asked for was to verify that the time limit is respected after a system crash. But we already have a test for that (see TckRecoveryTests#testCancelWhenParticipantIsUnavailable).

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
@@ -23,6 +23,8 @@ import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT
 import static org.eclipse.microprofile.lra.tck.participant.api.AfterLRAListener.AFTER_LRA_LISTENER_WORK;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.ACCEPT_WORK;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.LRA_RESOURCE_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.TIME_LIMIT;
+import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.TIME_LIMIT_HALF_SEC;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.TRANSACTIONAL_WORK_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.JOIN_WITH_EXISTING_LRA_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.ParticipatingTckResource.JOIN_WITH_EXISTING_LRA_PATH2;
@@ -314,7 +316,7 @@ public class TckTests extends TckTestBase {
 
     @Test
     public void timeLimit() {
-        WebTarget resourcePath = tckSuiteTarget.path(LRA_RESOURCE_PATH).path("timeLimit");
+        WebTarget resourcePath = tckSuiteTarget.path(LRA_RESOURCE_PATH).path(TIME_LIMIT);
         Response response = resourcePath
                 .request()
                 .get();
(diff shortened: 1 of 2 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -344,6 +346,31 @@ public void timeLimit() {
                 1, lraMetricService.getMetric(LRAMetricType.Compensated, lraId, LraResource.class.getName()));
     }
 
+    /**
+     * Service A - Timeout 500 ms
+     * Service B - Type.MANDATORY``
```
xstefank:
```suggestion
     * Service B - Type.MANDATORY
```

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java
```diff
@@ -410,6 +417,34 @@ public Response timeLimit(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
...
+        activityStore.add(new Activity(lraId));
+
+        try {
+            Thread.sleep(1000); // sleep for longer than specified in the timeLimit annotation attribute
+            // force the implementation to notice that the LRA should have timed out
+            lraTestService.waitForCallbacks(lraId);
+            // the next request should fail with a 412 code since the LRA should no longer be active
+            restPutInvocation(lraId, MANDATORY_LRA_RESOURCE_PATH, "");
+        } catch (WebApplicationException wae) {
+            return wae.getResponse();
```
xstefank: can you also log the exception or rethrow it? With the current code, the test fails only on failed assertion on status. The error message can be helpful.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java
```diff
@@ -410,6 +417,34 @@ public Response timeLimit(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
...
+        } catch (InterruptedException e) {
+            LOGGER.log(Level.FINE, "Interrupted because time limit elapsed", e);
+            fail("Interrupted because time limit elapsed");
```
xstefank: Will this work since this is a JAX-RS resource (i.e. not in the test)

tomjenkinson: Good question, fail() (at least in junit4 master, today) seems to translate into an [exception](https://github.com/junit-team/junit4/blob/master/src/main/java/junit/framework/Assert.java#L57) but I am not sure what the calling side receives if anything?

xstefank: 500 I believe since with exception it's failed JAX-RS invocation.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: ok

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java
mmusgrov: It is only an error if the status is not 412.
412 is the expected status and the resource is returning that to the test. The test has the responsibility to do the assertion. Add a log message here would at best provide tracing and we don't normally trace TCK tests. The calling test will report any unexpected status code.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java
mmusgrov: I will return a response code instead.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java
xstefank: sorry, I forgot that this method throws an exception when the status is different than 200.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java
```diff
@@ -410,6 +415,34 @@ public Response timeLimit(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
...
+        activityStore.add(new Activity(lraId));
+
+        try {
+            Thread.sleep(1000); // sleep for longer than specified in the timeLimit annotation attribute
+            // force the implementation to notice that the LRA should have timed out
+            lraTestService.waitForCallbacks(lraId);
+            // the next request should fail with a 412 code since the LRA should no longer be active
+            restPutInvocation(lraId, MANDATORY_LRA_RESOURCE_PATH, "");
+        } catch (WebApplicationException wae) {
+            return wae.getResponse();
```
xstefank:
```suggestion
            return Response.status(wae.getResponse().getStatus()).build();
```

I tested this locally and the response returned in `wae` from `restPutInvocation` is actually closed in this method [1] which fails the underlying JAX-RS implementation at least in Narayana.

[1] https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java#L327

mmusgrov: Thanks. I pushed your recommended fix.

PR #269, opened 2020-01-20
Some small Fixes to spec doc

rdebusscher:


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -519,7 +519,7 @@ returns the particular HTTP status codes identified in the `cancelOn` and
   @POST
   public Response bookTrip(...) { ... }
 
-  @LRA(LRA.Type.SUPPORTS, // if there is an incoming context then it used as the active context
+  @LRA(LRA.Type.SUPPORTS, // if there is an incoming context then it is used as the active context
```
rdebusscher: And I would define this method as LRA.Type.MANDATORY

The confirm of the trip is performed in the same LRA as the booking (so that booking can be reverted if not confirmed or when failure happens)

When booking is within a separate LRA, the types in the example would be different.

The Type.SUPPORTS allow that the JAX-RS method is called outside an LRA which probably does not make sense.

mmusgrov: Do you want to add this to your PR?

rdebusscher: I first wanted to know the opinion of the other team members.

I did a new commit to change that line.

PR #271, opened 2020-01-30
issue214 Add TCK for LRA placement on interface and super classes

### LINKED ISSUE #214: TCK: LRA placement on interface and super classes
We need some TCK tests to prove implemenation picks up the @LRA on a interface and super class.

Proposal: Duplicate `TckLRATypeTests` but once for @LRA on an interface and once for @LRA on superclass.

xstefank:
resolves #214

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
@@ -20,6 +20,9 @@
 package org.eclipse.microprofile.lra.tck;
 
 import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
+import org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckInterface;
+import org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource;
+import org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckSuperclass;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
@@ -33,19 +36,9 @@ import java.net.URI;
 import java.time.temporal.ChronoUnit;
 import java.util.logging.Logger;
 
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.MANDATORY_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.MANDATORY_WITH_END_FALSE_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.NEVER_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.NEVER_WITH_END_FALSE_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.NOT_SUPPORTED_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.NOT_SUPPORTED_WITH_END_FALSE_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.REQUIRED_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.REQUIRED_WITH_END_FALSE_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.REQUIRES_NEW_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.REQUIRES_NEW_WITH_END_FALSE_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.SUPPORTS_PATH;
-import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.SUPPORTS_WITH_END_FALSE_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckInterfaceResource.TCK_LRA_TYPE_INTERFACE_RESOURCE_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckResource.TCK_LRA_TYPE_RESOURCE_PATH;
+import static org.eclipse.microprofile.lra.tck.participant.api.LRATypeTckSuperclassResource.TCK_LRA_TYPE_SUPERCLASS_RESOURCE_PATH;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotEquals;
@@ -77,7 +70,7 @@ import static org.junit.Assert.assertTrue;
 @RunWith(Arquillian.class)
 public class TckLRATypeTests extends TckTestBase {
     private static final Logger LOGGER = Logger.getLogger(TckLRATypeTests.class.getName());
-    
+
     @Deployment(name = "lra-type-tck-tests")
     public static WebArchive deploy() {
         return deploy(TckLRATypeTests.class.getSimpleName().toLowerCase());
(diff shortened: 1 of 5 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckInterface.java
```diff
@@ -0,0 +1,97 @@
...
+
+import javax.ws.rs.GET;
+import javax.ws.rs.HeaderParam;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Response;
+import java.net.URI;
+
+import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
+
+@Path(LRATypeTckInterface.TCK_LRA_TYPE_INTERFACE_RESOURCE_PATH)
```
rdebusscher: @Path on the interface is not supported by JAX-RS.

Section 3.6 Annotation inheritance

>JAX-RS annotations may be used on the methods and method parameters of a super-class or an implemented interface. Such annotations are inherited by a corresponding sub-class or implementation class method provided that the method and its parameters do not have any JAX-RS annotations of their own. Annotations on a super-class take precedence over those on an implemented interface. The precedence over conflicting annotations defined in multiple implemented interfaces is implementation specific. **Note that inheritance of class or interface annotations is not supported.**

rdebusscher: Moving the annotation to LRATypeTckInterfaceResource solves the issue and is keeps all the rest of the code still valid and runnable.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckBean.java
```diff
@@ -0,0 +1,93 @@
...
+    public static final String REQUIRED_WITH_END_FALSE_PATH = "/end-required";
+    public static final String REQUIRES_NEW_WITH_END_FALSE_PATH = "/end-requires-new";
+    public static final String SUPPORTS_WITH_END_FALSE_PATH = "/end-supports";
+    public static final String NOT_SUPPORTED_WITH_END_FALSE_PATH = "/end-not-supported";
+    public static final String MANDATORY_WITH_END_FALSE_PATH = "/end-mandatory";
+    public static final String NEVER_WITH_END_FALSE_PATH = "/end-never";
+
+    // resource methods for each LRA.Type attribute reused in all LRA type tests
+
+    public Response requiredLRA(URI lraId) {
```
rdebusscher: All the methods in this class have the same body and thus can be replaced by a single method. (and even not required to have a method to represent a single simple statement)

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckSuperclass.java
```diff
@@ -0,0 +1,97 @@
...
+import javax.ws.rs.GET;
+import javax.ws.rs.HeaderParam;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.Response;
+import java.net.URI;
+
+import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
+
+@Path(LRATypeTckSuperclass.TCK_LRA_TYPE_SUPERCLASS_RESOURCE_PATH)
+public abstract class LRATypeTckSuperclass {
```
rdebusscher: Same remark as the interface one. @Path on abstract class not supported. move it to LRATypeTckSuperclassResource.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckInterface.java
xstefank: thanks, interestingly it works with RESTEasy. I will update this.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckBean.java
xstefank: I made this decision based on thinking that if we are to change any of the methods you will need to update it in 3 places, but maybe this is an unnecessary optimization. I will revert back to returning response directly.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckConfig.java
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.lra.tck.participant.api;
+
+public class LRATypeTckConfig {
```
mmusgrov: This class makes it difficult to navigate from the test to the resource that the test is using. Before I could be looking at the test in the IDE and click the resource path which takes me to the resource. Adding this level of redirection now means I have to work a lot harder to find the resource. Since the change is not tied to the issue can we go back to the way it was before.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
```diff
@@ -93,145 +95,501 @@ public void after() {
         NONE, NOT_PRESENT, EQUALS, NOT_EQUALS
     }
 
+    // LRA annotations placed on the resource class (see LRATypeTckResource)
+
     @Test
     public void requiredWithLRA() {
-        resourceRequest(REQUIRED_PATH, true, 200, MethodLRACheck.EQUALS, false);
+        requiredWithLRA(TCK_LRA_TYPE_RESOURCE_PATH);
```
mmusgrov: This is odd, you have added an extra level of indirection for no particular reason. Before I could scan the tests and get a good feel for how they differ. Since the change has no bearing on the issue can we go back to how they were originally.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckConfig.java
tomjenkinson: I guess it is because more than one resource now has this path directly? Perhaps this new file could be used to provide a consistently named list of constants and each of the different resources reference it where necessary. e.g. in LraTypeResource:
```public static final String REQUIRED_PATH = LRATypeTckConfig.REQUIRED_PATH;```

File: tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java [hunk 6]
```diff
@@ -244,9 +602,9 @@ public void neverWithoutEndLRA() {
      * @param methodLRAShouldBeActive if true the LRA started by the invoked method should still
      *                               be active after the resource invocation completes
      */
-    private String resourceRequest(String path, boolean startLRA, int expectedStatus, MethodLRACheck lraCheckType,
+    private String resourceRequest(String rootPath, String path, boolean startLRA, int expectedStatus, MethodLRACheck lraCheckType,
```
mmusgrov: This change seems unrelated to the issue.

mmusgrov: Don't forget to update the javadoc for the new parameter

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckConfig.java
mmusgrov: The constants are specific to the resource where they are used (if they happen to be the same string then that is coincidental). Putting them in a separate class just obfuscates the tests.

xstefank: so then TckLRATypeTests class will have 3 imports for each of the path parameters (as Tom mentioned same strings) just to be able to move to the right method from the test in one click? Isn't it a little redundant? I can, for instance, add a comment with:

```
    /**
     * {@link LRATypeTckResource#requiredLRA}
     */
```
to each test which can avoid duplication. Thougts?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
xstefank: the benefit is that all 3 resources are handled in the same way. I don't want to keep the original code because the same logic is used in all 3 cases.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckConfig.java
mmusgrov: What is wrong with separate imports. The problem with your approach is as I stated above, the path belongs with the resource where it is used. Putting it somewhere else is counter intuitive and makes navigation awkward.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckLRATypeTests.java
mmusgrov: Here is an example of where the extra indirection just adds overhead:

```java
    @Test
    public void requiredWithLRA() {
        requiredWithLRA(TCK_LRA_TYPE_RESOURCE_PATH);
    }

    @Test
    public void requiredWithoutLRA() {
        requiredWithoutLRA(TCK_LRA_TYPE_RESOURCE_PATH);
    }

    private void requiredWithLRA(String rootPath) {
        resource Request(rootPath, REQUIRED_PATH, true, 200, MethodLRACheck.EQUALS, false);
    }

    private void requiredWithoutLRA(String rootPath) {
        resourceRequest(rootPath, REQUIRED_PATH, false, 200, MethodLRACheck.NOT_EQUALS, false);
    }
```

and all of the methods you changed do the same. The original code was clear and did not have this obfuscating indirection.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckConfig.java
tomjenkinson: I think one thing to clarify is whether these strings definitely have the exact same purpose. Is it possible two different services might need to vary any of them in the future. e.g "/required" and "/required2" on two different services?

tomjenkinson: Actually if they were different services they would likely have a different root `@Path` to separate them. So it would be whether you might want one service to have two REQUIRED_PATH values.

xstefank: They have different root paths so not really. I am already in the process of changing this through. I will push the upgrade later today.

xstefank:
@mmusgrov @tomjenkinson PR updated to have paths directly in resources.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckResource.java
```diff
@@ -34,6 +34,7 @@
 @ApplicationScoped
 @Path(LRATypeTckResource.TCK_LRA_TYPE_RESOURCE_PATH)
 public class LRATypeTckResource {
+
```
tomjenkinson: I personally prefer the spacing style that is changed in this file and presumably it's not broken the checkstyle? But it seems like something that should be considered for future checkstyle rules to help keep patches clearer. In this case the file is connected to being changed by issue214 but it is just formatting from what I can see?

xstefank: checkstyle passed. There shouldn't be any problems with formatting. But the formatting rules is something that we can revisit.

PR #273, opened 2020-02-01
Issue272 Add TCK for deployment failure when LRA is not combined with either Compensate or AfterLRA

### LINKED ISSUE #272: Add TCK for deployment failure when LRA is not combined with either Compensate or AfterLRA
#244 introduced requirement for Compensate or AfterLRA to be present in each class which has an LRA method. This requirement is supported by the deployment failure [1]. We need to add TCK for this use-case.

[1] https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L44

xstefank:
resolves #272 

* add a new TCK test - `TckInvalidSignaturesTests#invalidLRAResourceWithoutCompensateOrAfterLRATest`
* fix other TCK LRA resources which do not contain any of the Compensate of AfterLRA methods
* ignore TckTests#mixedMultiLevelNestedActivity because it collides with the above requirement (will be reported in the new issue - https://github.com/eclipse/microprofile-lra/issues/274)

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckInvalidSignaturesTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckInvalidSignaturesTests.java
@@ -19,6 +19,7 @@
  *******************************************************************************/
 package org.eclipse.microprofile.lra.tck;
 
+import org.eclipse.microprofile.lra.tck.participant.invalid.LRAResourceWithoutCompensateOrAfteRLRA;
 import org.eclipse.microprofile.lra.tck.participant.nonjaxrs.InvalidAfterLRASignatureListener;
 import org.eclipse.microprofile.lra.tck.participant.nonjaxrs.InvalidArgumentTypesParticipant;
 import org.eclipse.microprofile.lra.tck.participant.nonjaxrs.InvalidReturnTypeParticipant;
@@ -55,6 +56,7 @@ public class TckInvalidSignaturesTests {
     private static final String TOO_MANY_ARGS_DEPLOYMENT = "too-many-args-deploy";
     private static final String INVALID_ARGUMENT_TYPE_DEPLOYMENT = "nonjaxrs-argument-type-deploy";
     private static final String INVALID_AFTER_LRA_SIGNATURE_DEPLOYMENT = "invalid-after-lra-deploy";
+    private static final String INVALID_LRA_RESOURCE_DEPLOYMENT = "invalid-lra-resource-deploy";
 
     @Rule
     public ExpectedException expectedException = ExpectedException.none();
@@ -83,6 +85,11 @@ public class TckInvalidSignaturesTests {
     @Deployment(name = INVALID_AFTER_LRA_SIGNATURE_DEPLOYMENT, managed = false)
     public static WebArchive deployInvalidAfterLRASignatureResource() {
         return createArchive(InvalidAfterLRASignatureListener.class);
+    }
+
+    @Deployment(name = INVALID_LRA_RESOURCE_DEPLOYMENT, managed = false)
+    public static WebArchive deployInvalidLRAResource() {
+        return createArchive(LRAResourceWithoutCompensateOrAfteRLRA.class);
     }
 
     private static WebArchive createArchive(Class<?> resourceClass) {
@@ -128,6 +135,14 @@ public class TckInvalidSignaturesTests {
     @Test
     public void invalidAfterLRASignatureTest() {
         testInvalidDeployment(INVALID_AFTER_LRA_SIGNATURE_DEPLOYMENT);
+    }
+
+    /**
+     * Verify that invalid LRA resource which does not contain any of @Compensate or @AfterLRA methods is detected
+     */
+    @Test
+    public void invalidLRAResourceWithoutCompensateOrAfterLRATest() {
+        testInvalidDeployment(INVALID_LRA_RESOURCE_DEPLOYMENT);
     }
 
     private void testInvalidDeployment(String deploymentName) {
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
@@ -64,6 +64,7 @@ import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
@@ -177,6 +178,7 @@ public class TckTests extends TckTestBase {
     }
 
     @Test
+    @Ignore("https://github.com/eclipse/microprofile-lra/issues/274")
     public void mixedMultiLevelNestedActivity() throws WebApplicationException {
         multiLevelNestedActivity(CompletionType.mixed, 2);
     }
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckResource.java
@@ -19,6 +19,8 @@
  *******************************************************************************/
 package org.eclipse.microprofile.lra.tck.participant.api;
 
+import org.eclipse.microprofile.lra.annotation.AfterLRA;
+import org.eclipse.microprofile.lra.annotation.LRAStatus;
 import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -137,5 +139,10 @@ public class LRATypeTckResource {
     @LRA(value = LRA.Type.NEVER, end = false)
     public Response neverEndLRA(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
         return Response.ok(lraId).build();
+    }
+
+    @AfterLRA
+    public void afterLRA(URI lraId, LRAStatus status) {
+        // no-op, required by the specification (see https://github.com/eclipse/microprofile-lra/pull/265)
     }
 }
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/NonParticipatingTckResource.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/NonParticipatingTckResource.java
@@ -19,6 +19,8 @@
  *******************************************************************************/
 package org.eclipse.microprofile.lra.tck.participant.api;
 
+import org.eclipse.microprofile.lra.annotation.AfterLRA;
+import org.eclipse.microprofile.lra.annotation.LRAStatus;
 import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
 import org.eclipse.microprofile.lra.tck.LRAClientOps;
 
@@ -82,7 +84,7 @@ public class NonParticipatingTckResource {
 
     @PUT
     @Path(SUPPORTS_PATH)
-    @LRA(value = LRA.Type.SUPPORTS)
+    @LRA(value = LRA.Type.SUPPORTS, end = false)
     public Response supports(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
         return Response.ok(lraId).build();
     }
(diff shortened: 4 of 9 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -177,6 +178,7 @@ public void compensateMultiLevelNestedActivity() throws WebApplicationException
     }
 
     @Test
+    @Ignore
```
mmusgrov: Why isn't this included in the PR for issue #274  ?

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckResource.java
```diff
@@ -125,4 +127,9 @@ public Response notSupportedEndLRA(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lra
     public Response neverEndLRA(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
         return Response.ok(lraId).build();
     }
-}
\ No newline at end of file
+
+    @AfterLRA
+    public void afterLRA(URI lraId, LRAStatus status) {
+        // no-op, required by the specification
```
mmusgrov: Since this change should have been included with PR 265 (https://github.com/eclipse/microprofile-lra/pull/265/files#diff-06aca14982a5ee1e4547e9ec14b41bbbR400) maybe you could reference the spec to say why it is required so that we have traceability?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
tomjenkinson: Assuming it is required for this to be ignored to get this PR to pass, might I suggest a comment in the code and on the commit message as to why it is ignored? and associate this PR with both issues (even if it is just to ignore this test and doesn't fix 274)

xstefank: Yes, this is the conflict described in #274. I will do what Tom suggests.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LRATypeTckResource.java
xstefank: ok, I will link the PR in the comment.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckInvalidSignaturesTests.java
```diff
@@ -64,7 +66,7 @@
 
     @ArquillianResource
     private Deployer deployer;
-
```
tomjenkinson: I am not sure if the community has a preference decided here yet but I wonder if formatting of all the sources could be done at the same time and maybe add a checkstyle rule to the build?

mmusgrov: Our maven-checkstyle-plugin rules are [here](https://github.com/eclipse/microprofile-lra/blob/master/pom.xml#L204). When I created the rules I copied them from another MP pom (and they all seemed to be using the same set but I haven't checked since). We don't have any rules around blank lines though (but I think it is doable with a regex).

xstefank: This was a forgotten switch on my side. It's a manual revert so if it's required I will revert it.

tomjenkinson: Although I don't think blank lines were effectively added, lines that would be blank other than whitespace were modified. I think it looks like trailing whitespace was removed? If you look at https://github.com/eclipse/microprofile-lra/pull/273/files you will see what I mean.

As the changes meets the current checkstyle rules of the project then that would suggest (to me, non-committer on this project) it does not need reverting.  I would comment though that even if it is not in the checkstyle rules, it seems MP Config has a .editorconfig for non-.md and non-.adoc files: https://github.com/eclipse/microprofile-config/blob/828ee2fee4ccbef6701943bb96d7be187408fd63/.editorconfig#L12 which I assume might remove the trailing whitespace in at least the .java files in editors that understand the format (just surmising here though as I am not an expert on the .editorconfig file type)

I also haven't tried the following, but I do wonder if preventing trailing whitespace is something the community want then https://checkstyle.org/config_whitespace.html#NoWhitespaceAfter might detect this?

tomjenkinson: Actually, I can't be sure if it does or doesn't meet the checkstyle rules as I don't know if CI is ran on this PR and if so whether it tests the checkstyle?

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/NonParticipatingTckResource.java
```diff
@@ -82,7 +84,7 @@ public Response neverRunWithLRA(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId)
 
     @PUT
     @Path(SUPPORTS_PATH)
-    @LRA(value = LRA.Type.SUPPORTS)
+    @LRA(value = LRA.Type.SUPPORTS, end = false)
```
mmusgrov: Will this change result in any active LRAs when the test finishes. For example, will the LRA in the following call ever be closed:
 
https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L232

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckInvalidSignaturesTests.java
tomjenkinson: CI was not ran on the PR

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/invalid/LRAResourceWithoutCompensateOrAfteRLRA.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.ws.rs.Path;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.core.Response;
+import java.net.URI;
+
+@Path(LRAResourceWithoutCompensateOrAfteRLRA.INVALID_LRA_RESOURCE_PATH)
+public class LRAResourceWithoutCompensateOrAfteRLRA {
+
+    public static final String INVALID_LRA_RESOURCE_PATH = "/invalid-lra-resource-path";
+    public static final String LRA_PATH = "/lra";
```
rdebusscher: Doesn't seems to be used, Line 40 has a constant, not referring to this variabnle. Can be removed

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckInvalidSignaturesTests.java
mmusgrov: We do not run CI on PRs.

The NoWhitespaceAfter rule would be useful.

I would suggest we do not apply the same standard of rigour to pre GA PR's when it comes to formatting, after all the GA release should be clean from that point of view and thereafter we can apply the higher bar for PRs with respect to incidental changes such as cleaning up white space.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/invalid/LRAResourceWithoutCompensateOrAfteRLRA.java
xstefank: fixed

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/NonParticipatingTckResource.java
xstefank: I checked with Narayana and there are no open LRAs after full TCK pass.

xstefank:
@rdebusscher @mmusgrov @ochaloup this is ready for review again.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/NonParticipatingTckResource.java
ochaloup: @xstefank sorry for ignorance but why is needed to change the `SUPPORTS` here for not ending the LRA?

xstefank: hah, refreshing my memory :) So I think that the issue was that in some test the LRA was closed here and then propagated to some LRA resource into which I added the AfterLRA method (you can check by removing this change if you wish to know which test it was). So with AfterLRA the already ended LRA was rejected because only Cancelling/Closing LRAs are allowed with the AfterLRA method present.

mmusgrov: Where is the LRA closed or cancelled?

xstefank: So the affected test is [TckContextTests#testContextAfterRemoteCalls](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckContextTests.java#L255) which only calls [this JAX-RS method](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L214). The test calls the method which we are discussing in the [getActiveLRA()](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L214) method. As you can see in the [test flow](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L214), the test starts a new LRA and then on line [216](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L216) it calls this SUPPORTS_PATH resource method with this LRA where the LRA is currently closed (as it does have end = true which is the default). Then after some checks the `getActiveLRA()` is called again on line [224](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L224) where this SUPPORTS_PATH resource method is called again with already closed LRA.

As I added `@AfterLRA` method to `NonParticipatingTckResource` it is no longer truly non participating as it is an After LRA listener. So the invocation on line [224](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L224) fails as the LRA is already ended and AfterLRA listener requires only Active, Closing or Cancelling LRA.

Adding `end=false` fixes the issue and the LRA is closed at the end of [contextCheck](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L214) method.

xstefank: @mmusgrov ^^

mmusgrov: > Adding `end=false` fixes the issue and the LRA is closed at the end of [contextCheck](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ContextTckResource.java#L214) method.

The end of this method is:

``` java
        // invoke a remote service which runs in the callers context, ie do set the context header
        remote = restPutInvocation(active, TCK_NON_PARTICIPANT_RESOURCE_PATH, SUPPORTS_PATH, "");
        assertEquals("contextCheck4: the remote service should have ran with the active context", remote, active);
        assertEquals("contextCheck4: after calling a remote service the active LRAs is different", active, getActiveLRA());

        return Response.ok().entity(lraId.toASCIIString()).build();
```

so how does it get closed?

xstefank: @mmusgrov the LRA is closed normally at the end of `contextCheck` method which has and LRA annotation -- `@LRA(value = LRA.Type.REQUIRED)`.

xstefank:
@mmusgrov I am going to merge this later today. Do you still have any questions?

xstefank:
@mmusgrov @tomjenkinson we have 2 approvals on this PR. I am going to merge this later today so this is just a notification if you want to take another look.

xstefank:
Pushed only changes suggested by @tomjenkinson in his last comment.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/NonParticipatingTckResource.java
tomjenkinson: I think I follow what @xstefank is saying about `contextCheck`.

xstefank:

Issue #274, opened 2020-02-01
Conflict in TCK: unable to cancel already closed nested LRA


xstefank:
With #244 we require now that every LRA resource must contain either Compensate or AfterLRA method:

* Compensate - requires that received/created LRA is active when the LRA method is invoked
* AfterLRA - requires that received/created LRA is active, Closing, or Cancelling when the LRA method is invoked

In the TCK test `TckTests#mixedMultiLevelNestedActivity` we test the scenario where the nested LRA is first closed and then subsequently canceled. That means that we are invoking LRA method with already closed LRA which together with the requirement of #244 means that such invocations will always fail.

This means that #244 basically removed the possibility of mixed nested outcomes from 1.0. It will be possible to simulate `TckTests#mixedMultiLevelNestedActivity` scenario with the `LRAClient` when it will be introduced later.

For now, I added ignore to `TckTests#mixedMultiLevelNestedActivity` in PR #273.

xstefank:
Possible solutions (I assume we can't change the semantics of Compensate methods):

* remove the requirement from AfterLRA resource on particular LRA status when LRA method is invoked (if LRA is already closed and implementation doesn't know about it anymore, the AfterLRA method will never be invoked)

* update the specification text about mixed nested outcomes (if there are any mentions) and remove `TckTests#mixedMultiLevelNestedActivity` (we can reintroduce this test when we add `LRAClient` interface

rdebusscher:
>- Compensate - requires that received/created LRA is active when the LRA method is invoked
> - AfterLRA - requires that received/created LRA is active, Closing, or Cancelling when the LRA method is invoked

This is about when the LRA method is executed, not about when the @Compensate annotated method is executed.

I also doesn't see any change in #265 which indicates that something has changed regarding when the @Compensate annotated method can be executed.

In the section of `Forgetting an LRA` there is the explicit mention that @Compensate can be called again when the parent LRA closes

>a nested LRA can be closed independently from its parent but
it must retain the ability to compensate until the parent has finished.

So I for me this issue is invalid.

xstefank:
@rdebusscher what do you mean, please? In the copied statement it is said that we are talking about LRA methods (e.g., Compensate - requires that received/created LRA is active **when the LRA method is invoked**).

[...]

PR #276, opened 2020-02-12
issue274 Clarify context validation for nested LRAs

mmusgrov:
issue #274 

The spec update adds a note about validation of nested LRAs.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
@@ -275,10 +275,10 @@ public @interface LRA {
          *     <li>A nested LRA which has closed must retain the ability to cancel the
          *     effects if the the parent cancels. This requirement must be enforced
          *     by participants.
-         *     <li>If a nested LRA cancels then all of its children must cancel (even if
-         *     they closed - see 2).
+         *     <li>If a nested LRA cancels then all of its children must effectively cancel
+         *     (even if they have previously been asked to close - see 2)
          *     <li>If a nested LRA closes then it, and all of its children, must close
-         *     (but retain the ability to later compensate - see 2).
+         *     (but retain the ability to later cancel - see 2).
          * </ol>
          * <p>
          *     Downstream LRAs will only be part of this nesting hierarchy if the
@@ -292,8 +292,18 @@ public @interface LRA {
          *     there is no longer any outstanding work in need of completion.
          * </p>
          * <p>
-         *     On the other hand it does make sense to cancel a closed nested LRA since
-         *     the work has been done so there is something that can be compensated for.
+         *     On the other hand it does make sense to cancel the effects of a
+         *     closed nested LRA since the work has been done and there is
+         *     something that can be compensated for. In this case the LRA
+         *     method invocation is allowed to proceed only if the participant
+         *     is already enlisted with the LRA, otherwise the method invocation
+         *     is rejected using a <code>412 Precondition Failed</code> HTTP
+         *     status code. If the method invocation causes the LRA to close
+         *     then the {@link Complete} annotated method, if present,
+         *     <i>is not</i> called again. If the LRA method causes the LRA
+         *     to cancel then the nested LRA is moved to the
+         *     {@link org.eclipse.microprofile.lra.annotation.LRAStatus#Cancelling}
+         *     state and the {@link Compensate} callback will be invoked.
          * </p>
          * <p>
          *     Therefore, as a consequence of requirement 2, any activities performed in
@@ -304,12 +314,16 @@ public @interface LRA {
          *     will be told to compensate. This implies that the nested participants
          *     must be aware that they are nested and the JAX-RS header with the
          *     name {@value #LRA_HTTP_PARENT_CONTEXT_HEADER} is guaranteed to hold
-         *     the parent context whenever a nested LRA is active.
+         *     the parent context whenever a nested LRA is being propagated.
          * </p>
          *
          * <p>
-         *     A participant which has closed can determine when the top level
+         *     A participant which has completed can determine when the top level
          *     parent has closed by providing a {@link Forget} callback handler.
+         *     When this method is called there is no longer a requirement to
+         *     maintain the ability to reverse the effects of a closed nested LRA.
+         *     This can be used by the participant to clean up any resources it used
+         *     to implement this guarantee.
          * </p>
          *
          * <p>
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -406,8 +406,17 @@ when the LRA reaches one of <<lra-state-model,the final LRA states>>.
 In the case that the class contains a `@Compensate` annotated method, this means that before
 executing the `@LRA` annotated method, the implementation must be able to guarantee that the
 participant has been or will eventually be registered with the LRA.
-This implies that the LRA is effectively validated as being active (since
-it is not possible to enlist with an inactive LRA).
+This implies that the LRA is effectively validated as being active since it is not possible
+to enlist with an inactive LRA. There is an exception to this statement if the LRA is nested
+and in the `Closed` state (since the transition `Closed` -> `Cancelling` is permitted). In this
+case the method invocation will be allowed to proceed provided that the participant is already
+registered with the LRA (please refer to the <<source-LRA,javadoc>> for the `NESTED`
+value of the `LRA#value()`
+element for more information about this situation, in particular if the method invocation
+causes the LRA to close then the `@Complete` annotated method, if present, *is not* called
+again). And if the participant is not registered with the LRA then the caller of the LRA
+method will receive a `412` status response as described in the javadoc.
+
 In the case that the class contains a `@AfterLRA` annotated method, this means that before
 executing the `@LRA` annotated method, the implementation must be able to guarantee that the
 listener will be notified about the outcome of the LRA.


api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -275,8 +275,8 @@
          *     <li>A nested LRA which has closed must retain the ability to cancel the
          *     effects if the the parent cancels. This requirement must be enforced
          *     by participants.
-         *     <li>If a nested LRA cancels then all of its children must cancel (even if
-         *     they closed - see 2).
+         *     <li>If a nested LRA cancels then all of its children must cancel (or compensate
+         *     if they are already closed - see 2).
```
xstefank: This is a little confusing to me. Both LRAs and participants need to cancel/compensate even if they previously closed/completed. We shouldn't mix the terminology (LRA -- close/cancel, participant -- complete/compensate).

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -408,6 +408,8 @@ executing the `@LRA` annotated method, the implementation must be able to guaran
 participant has been or will eventually be registered with the LRA.
 This implies that the LRA is effectively validated as being active (since
 it is not possible to enlist with an inactive LRA).
+footnoteref:[note1,there is a caveat for nested LRAs which have an extra state model transition which allows an inactive LRA to transition from the `Closed` state to the `Cancelling` state. To facilitate this transition we allow enlistement if the implementation knows that the incoming context is nested and in the state `Closed` or `Active`. And in the case where the LRA is `Closed`, if the method finishes without error, the complete method, if present, would be invoked as normal.]
```
xstefank: or Status if Complete is not idempotent.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
rdebusscher: And it is not the nested LRA itself, but the parent of the nested LRAs that cancels


`if an LRA cancels then all the participants linked to the nested LRA children of this parent LRA need to compensate, even if they are already completed - see 2)`

But probably the word 'linked' is not the best term.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
rdebusscher: Is the term `enlistment` correct in this context `To facilitate this transition we allow enlistement if the implementation knows that the incoming context is nested and in the state Closed`

The participant related to the nested LRA is already enlisted and instructed to complete. Later on, this same participant is called to compensate because of the cancel of the parent LRA.

For me, enlistment is when a participant is 'created' due to a user-initiated call where we need to make sure the compensate/complete is possible in the future.

xstefank:
@mmusgrov can you take a look at this PR again, please?

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
mmusgrov: I have added more clarifying text.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: This footnote is to cover the situation where a participant wishes to join with a closed nested LRA (but the parent LRA is still active) and therefore needs to be notified when the nested LRA closes.

mmusgrov: This caveat allows new participants to register with closed nested LRA's (provided that the parent LRA is still active).

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -408,6 +408,12 @@ executing the `@LRA` annotated method, the implementation must be able to guaran
 participant has been or will eventually be registered with the LRA.
 This implies that the LRA is effectively validated as being active (since
 it is not possible to enlist with an inactive LRA).
+footnoteref:[note1,there is a caveat for nested LRAs which have an extra state model transition
+which allows an inactive LRA to transition from the `Closed` state to the `Cancelling` state.
+To facilitate this transition we allow enlistement if the implementation knows that the incoming
+context is nested and in the state `Closed` or `Active`. And in the case where the LRA is `Closed`,
+if the method finishes without error, the complete method, if present, would be invoked as normal.]
+
```
xstefank: Sorry, this footnote to me still causes more confusion then it explains the intended behavior. For instance, is `return Response.status(4xx).build()` finish with or without error?

In the issue you mentioned:

if the implementation knows that the incoming context is nested and not cancelled then we allow enlistment (even if it is closed), otherwise the method is not allowed to proceed.

Can you incorporate rather this information, please?

xstefank: And if the Complete method can be invoked again should we mark this transition in the state diagram? What happens if the nested Complete method on repeated close of nested LRA transforms nested LRA from Closed to FailedToClose state?

mmusgrov: The text `if the method finishes without error` means that the method does not result in a status code that causes the LRA to cancel: HTTP error statuses in the context of LRA mean to transition to Cancelling (ie the ones in cancelOn, cancelOnFamily attributes). I thought that was implicit but I can change it something more explicit.

To your second comment about `the Complete method can be invoked again ...`: the complete method is not invoked a second time so the transition cannot occur (for nested LRAs, only Closed -> Cancelling is permitted which would result in the Compensate method being called). The text you quote is referring to a new participant joining a closed nested LRA so its complete method would only be called once.

xstefank: > the method does not result in a status code that causes the LRA to cancel

This would be better so please change it to something more explicit. Don't forget also about non-JAX-RS cases.

>  The text you quote is referring to a new participant joining a closed nested LRA so its complete method would only be called once.

Ok, thanks. I understand now. But still, this new participant can return FailedToComplete from its Complete method invocation and thus there is a transition Closed -> FailedToClose or it can return Closing and thus there is the transition Closed -> Closing?

mmusgrov: This is a great observation. I do not think we should start messing up the state model with complex transitions so let's either:

1. Prohibit new participants from joining closed nested LRAs, or
2. Prohibit new participants with an `@Complete` callback from joining closed nested LRAs,

or both.

In the first situation the nested LRA can be cancelled by invoking a method on the original participant which returns a status code that results in cancel.

In the second situation the nested LRA can be cancelled by invoking a method on the original participant which returns a status code that results in cancel.

mmusgrov: FYI I just updated my last comment which had a typo in the last paragraph (I replaced `original` with `new`).

xstefank: I would vote for both. Only already enlisted nested participant LRA method can be invoked again and if this method closes LRA the Complete methods are not replayed again (thus no-op). If the LRA method cancels the nested LRA is moved to Cancelling state.

New participants are rejected with 412. This also confirms with the statement that we don't allow enlistment with already closed LRA.

mmusgrov: Agreed. This approach keeps the model relatively simple (and it means that the state models for LRAs and Participants are kept in harmony). I will update the proposal to include these two options and also include the text you proposed in the last comment.

In addition, your earlier comment about missing transitions prompted me to revisit the participant model. Would you agree that the participant model needs the transition `Completed -> Compensating` to match the LRA state model transition `Closed -> Cancelling` when the LRA is nested. If so then I will also update `Figure 3. Participant State Model` in this PR.

xstefank: > Would you agree that the participant model needs the transition Completed -> Compensating to match the LRA state model transition Closed -> Cancelling when the LRA is nested. If so then I will also update Figure 3. Participant State Model in this PR.

Yes, this model should also be updated.

xstefank: my bad, there is already an issue open to update the model - https://github.com/eclipse/microprofile-lra/issues/277. So this PR is good to go in my opinion. Only waiting for approval from @rdebusscher.

mmusgrov: > Don't forget also about non-JAX-RS cases.

This text refers to methods annotated with @LRA but I will note your comment when updating the rest of this PR.

mmusgrov:
@rdebusscher I made quite a few changes to the text in this PR. You might want to check whether or not you still want to approve it.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: I see that you decided to rather prohibit the Complete invocation in case the nested LRA method closes the nested LRA again. I don't see a problem with this approach, just documenting it here in case someone reads this thread again in the future.

xstefank:
Looks good to me. I will give another few days for @rdebusscher (as we still have another issue open).

mmusgrov:
@rdebusscher I am merging this PR. If you discover any further problems with this aspect of the spec then please reopen. Thanks

rdebusscher:
Sorry I missed this.

I don't think you need to `enlist` in the case of the nested case going from `Closed` to `cancelling`. 

For me enlisting means the participant is 'added' to the LRA and becomes active and part of LRA.

In this case, it is just the @Compensate method which is called on the participant (and thus participant never becomes active in that case)

But otherwise I don't see immediately any major concerns.

xstefank:
@rdebusscher I took a note of your last comment for final sweep / clean up.

Issue #279, opened 2020-02-24
TCK check for an active LRA can have a side effect of cancelling the LRA


mmusgrov:
Some TCK tests need to validate whether or not an LRA is active (see method isLRAFinished https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java#L81). The implementation does this by attempting to register a participant with the LRA using the method joinWithExistingLRA ( https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/ParticipatingTckResource.java#L129). But this method has the wrong LRA annotation since the default is to cancel the LRA on CLIENT (4xx) or SERVER (5xx) status responses. 

In other words the method has the side effect of cancelling the LRA if the method returns any 4xx or 5xx status code (and these codes can be returned for reasons other than failing to join with the LRA).

The fix is to not cancel the LRA on these codes, ie:
```java
@LRA(value = LRA.Type.MANDATORY, end = false, cancelOnFamily = {})
```

PR #280, opened 2020-02-24
issue279 TCK check for active LRAs can have the side effect of closin…

mmusgrov:
#279


xstefank:
@rdebusscher I will merge this as it is only a bug fix.

Issue #193, opened 2019-07-18
Interpretation of 404 status code


xstefank:
The javadoc for participant annotations states [now](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java#L117):

```xml
 * <tr>
 *   <td scope="row">404</td>
 *   <td scope="row">Empty</td>
 *   <td scope="row">The resource does not know about the LRA</td>
 * </tr>
```
However, 404 is the status code returned also when the target (participant) service is down for which probably most implementation would like to retry. Does the information that participant "does not know about this LRA" help the caller in any way? What would be wrong in this case with just simple dummy Compensated response?

mmusgrov:
> However, 404 is the status code returned also when the target (participant) service is down for which probably most implementation would like to retry.

If the service is down the caller will either get a `ConnectionRefused` exception (probably wrapped in `javax.ws.rs.ProcessingException` if called from JAX-RS code) or a 5xx code such as 502 or 504. I think we can rely on 404 meaning that the REST server cannot locate the resource.

404 has the meaning that the LRA spec desires:

>> The 404 (Not Found) status code indicates that the origin server did not find a current representation for the target resource

This means that the request has been received by the target server but it was unable to find the requested resource (ie the server is not down).

> Does the information that participant "does not know about this LRA" help the called in any way?

The fact that the server returns 404 means that the coordinator can infer that the participant "obeyed the LRA protocol". For example if it was asked to compensate and it did so then it can legitimately clean up.
In other words 404 is fundamental to the operation of the protocol.

> What would be wrong in this case with just simple dummy Compensated response?

If would be a lie if the participant has already cleaned up since it has erased all knowledge of the LRA. If the server/participant genuinely does not know about the resource then 404 is the correct response (since returning 200 would imply that the resource exists).

mmusgrov:
> What would be wrong in this case with just simple dummy Compensated response?

There is nothing stopping you from doing this but you are effectively creating resources that will live forever and since they are resources that can be legitimately cached they will occupy space in http caches.

tomjenkinson:
I think the crux of this is "404 is the status code returned also when the target (participant) service is down for which probably most implementation would like to retry. "

If that can be true I think it would be a problem, but have we ever seen that to be the case that a service is down and a 404 is returned?

tomjenkinson:
[...]

PR #281, opened 2020-03-02
issue6 TckTests#timeLimitWithPreConditionFailed expected response sho…

mmusgrov:
…uld be GONE or PRECONDITION_FAILED

This PR fixes a bug in the test added during PR #267 
The test timeLimitWithPreConditionFailed starts an LRA, waits for it to time out and then invokes another method that should cause another enlistment. Since the LRA will have timed out, the second call should fail with either 412 or 410 but the test only checks for 412.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
@@ -353,8 +353,8 @@ public class TckTests extends TckTestBase {
      * Service A calls Service B after it has waited 1 sec.
      * Service A returns http Status from the call to Service B.
      *
-     * test calls A and verifies if return is status 412 (precondition failed) since LRA
-     * is not active when Service B endpoint is called.
+     * test calls A and verifies if return is status 412 (precondition failed)
+     * or 410 (gone) since LRA is not active when Service B endpoint is called.
      */
     @Test
     public void timeLimitWithPreConditionFailed() {
@@ -364,9 +364,10 @@ public class TckTests extends TckTestBase {
                 .get();
 
         // verify that a 412 response code was generated
-        assertEquals("Expected 412 response",
-                Response.Status.PRECONDITION_FAILED.getStatusCode(),
-                response.getStatus());
+        assertTrue("Expected 412 or 410 response",
+                response.getStatus() == Response.Status.PRECONDITION_FAILED.getStatusCode() ||
+                        response.getStatus() == Response.Status.GONE.getStatusCode()
+                );
 
         response.close();
     }


rdebusscher:
For me

412 - used by registration of participant
410 - used for status/complete/compensate methods

And wording is MUST (SHOULD is recommendation and does mean not required and no test can be written in that case)

mmusgrov:
There is no spec requirement to keep state associated with terminated LRAs hanging around. Besides, such a requirement would theoretically require an unlimited amount of storage.

When an LRA has finished and when all callbacks have been successfully processed the system is allowed to perform any necessary cleanup.

xstefank:
@mmusgrov isn't that one more argument why the implementation MUST return 412? Because it cannot tell whether the LRA is invalid or ended?

xstefank:
@mmusgrov what is the state of this PR, please?

mmusgrov:
In this test the implementation knows that the LRA has finished. Since there is no spec requirement to keep state associated with terminated LRA's hanging around it is allowed to clean up and report `410 GONE` for any requests that refer to the terminated LRA.

xstefank:
> In this test the implementation knows that the LRA has finished. Since there is no spec requirement to keep state associated with terminated LRA's hanging around it is allowed to clean up and report `410 GONE` for any requests that refer to the terminated LRA.

Your statement contradicts itself. If the implementation is allowed to clean up finished LRAs it can't know that it ever knew about the received LRA. Thus it is impossible for it to distinguish between invalid and already finished LRA (when the finished LRA is cleaned up) and thus 412 is the only possible response here.

mmusgrov:
> > In this test the implementation knows that the LRA has finished. Since there is no spec requirement to keep state associated with terminated LRA's hanging around it is allowed to clean up and report `410 GONE` for any requests that refer to the terminated LRA.
> 
> Your statement contradicts itself. If the implementation is allowed to clean up finished LRAs it can't know that it ever knew about the received LRA. Thus it is impossible for it to distinguish between invalid and already finished LRA (when the finished LRA is cleaned up) and thus 412 is the only possible response here.

1. The problem is that we replaced `404 NOT FOUND` with `410 GONE` a few months back. So 410 GONE must now include the case where the implementation does not know about the LRA. Unless we want to reintroduce 404?
2. 410 GONE (or 404) provides more information than `412 Precondition Failed`.
3. I guess one also should assume that, in the absence of malicious entities, that the LRA context header should always refer to an LRA that was actually created at some previous time.

mmusgrov:
> When reading the spec and grepping the LRA spec then I think @rdebusscher 's comment : [#281 (comment)](https://github.com/eclipse/microprofile-lra/pull/281#issuecomment-593864362) pinpoints what the spec says.
> The response code `410` is used in this situation for `Complete`/`Compensate`/`Status` methods (https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#jax-rs-methods)

The part of the spec that you refer to discusses status codes that should be returned by participant/listener callbacks.
However, this PR concerns the failure, by the implementation, to enlist a participant/listener in a finished/timed out LRA.

> The response code `412` is used when the LRA enlistement and `LRA` annotation is discussed (
>
> https://github.com/eclipse/microprofile-lra/blob/71c39d38f516d7ca15b80e047c223ab5a8944333/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L215
>

The text accompanying the MANDATORY attribute applies when there is no LRA context header present on the incoming request. It does not cover the current situation (see my last comment below).

> )
> From that: if the LRA is gone then the `@Complete` should return `410` and the `@LRA` should return `412`.
>
> From that perspective here the `412` is the only possibility.
>
> Am I missing something?

At worst the spec is under specified and the reason for that is that we weren't diligent enough when we went through the spec a few months ago replacing `404 NOT FOUND` with `410 GONE`.

ochaloup:
@mmusgrov it seems the issue https://github.com/eclipse/microprofile-lra/issues/193 about replacing `404` to `410` was meant to agree that `404` has got an ambiguous meaning and it was changed to `410`. There was no more discussion about the meaning of the `410`.

If the `410` is a possible error code which may be emitted from the `@LRA` method then the current spec is not descriptive in that sense.

If the `412` would mean that there is no LRA context header present and the `410` when the LRA is not active or is gone it sounds reasonable to me. But it should be followed by an enhancement of the text to be clear.

xstefank:
> 410 GONE (or 404) provides more information than 412 Precondition Failed.

What can this information be used for?

> I guess one also should assume that, in the absence of malicious entities, that the LRA context header should always refer to an LRA that was actually created at some previous time.

This is not totally correct as if the implementation forgets about a particular LRA then it is able only to "guess" that the LRA was probably valid previously but can't guarantee it (at least in Narayana where we generate a random id for each new LRA). So what you are proposing is difference:

* 410 valid URI with the correct/expected LRA structure for the particular implementation (may be different for different impls) which is no longer actively tracked
* 412 invalid URI

if I got your ideas right. If I did, this won't work because we don't specify the structure of the URI that represents the LRA context.

mmusgrov:
> @mmusgrov it seems the issue #193 about replacing `404` to `410` was meant to agree that `404` has got an ambiguous meaning and it was changed to `410`. There was no more discussion about the meaning of the `410`.
> 
> If the `410` is a possible error code which may be emitted from the `@LRA` method then the current spec is not descriptive in that sense.
> 
> If the `412` would mean that there is no LRA context header present and the `410` when the LRA is not active or is gone it sounds reasonable to me. But it should be followed by an enhancement of the text to be clear.

Let's enhance the text then.

mmusgrov:
> > 410 GONE (or 404) provides more information than 412 Precondition Failed.
> 
> What can this information be used for?

The caller would know that the LRA has already finished which is not the case with 412.

> 
> > I guess one also should assume that, in the absence of malicious entities, that the LRA context header should always refer to an LRA that was actually created at some previous time.
> 
> This is not totally correct as if the implementation forgets about a particular LRA then it is able only to "guess" that the LRA was probably valid previously but can't guarantee it (at least in Narayana where we generate a random id for each new LRA). So what you are proposing is difference:

I said "in the absence of malicious entities". Therefore the presence of the header definitively states that the LRA exists or existed.

> 
> * 410 valid URI with the correct/expected LRA structure for the particular implementation (may be different for different impls) which is no longer actively tracked
> * 412 invalid URI
> 

We don't need to parse the header to infer that there is an LRA context (since we are assuming no malicious entities). Of course if we try to enlist with it and it has gone then the caller would get an error.

> if I got your ideas right. If I did, this won't work because we don't specify the structure of the URI that represents the LRA context.

My idea is the same as it was when I/we created the initial proposal: if the implementation has no record of the LRA then it reports that information using a `410 GONE`, previously we used 404 which is REST friendly but, as you said, ambiguous status code.

xstefank:
@mmusgrov ok, I understand now your intention. Can you also update the text as @ochaloup suggested, please?

xstefank:
@rdebusscher is this ok with you?

rdebusscher:
For me it should only test on 412 as it is about enlisting a participant.

Also the spec is mentioning status 410 in relation to complete/compensate/status.

But we also have this line which also checks on 410 and 412

https://github.com/eclipse/microprofile-lra/blob/e1ba77c4ef3573d1ce9633683c1f06b359ac091d/tck/src/main/java/org/eclipse/microprofile/lra/tck/LRAClientOps.java#L82

So there is definitively more clarification needed.

xstefank:
@rdebusscher agreed that more clarification is needed. Can we merge this now and do the unification of 412 and/or 410 cases in a separate PR, please?

rdebusscher:
@xstefank ok, but I do not agree that 412 and 410 should be used everywhere (enlistment and callbacks) but each http status should be used for a specific type of calls.

xstefank:
Merging this as discussed with the discussion about correct handling of 410 and 412 moved to https://github.com/eclipse/microprofile-lra/issues/290.

PR #282, opened 2020-03-12
#275 Fix tests that wrongly assume callbacks are delivered with "exac…

### LINKED ISSUE #275: Some tests wrongly assume callbacks are delivered with "exactly once" semantics
An example is TckContextTests#testForget that asserts that the forget callback is delivered exactly once. For example the notification could reach the participant but the response could be lost which would cause the implementation to reissue the forget call. Our delivery semantics is "at least once".

xstefank:
…tly once" semantics

resolves #275 

Changes:
* Complete/Compensate to "at least once" where there is no Status method in the resource
  * otherwise keep exactly-once semantics, we probably need to cover how the implementation should behave if there is the Status method and Complete/Compensate call fails (this will also require spec text changes so I will open a new issue - https://github.com/eclipse/microprofile-lra/issues/283)
* AfterLRA, Forget, Status to "at least once"

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
@@ -405,12 +405,13 @@ public class TckTests extends TckTestBase {
         int completionCount = lraMetricService.getMetric(LRAMetricType.Completed, lra, LraResource.class.getName());
         int compensationCount = lraMetricService.getMetric(LRAMetricType.Compensated, lra, LraResource.class.getName());
 
-        boolean wasCalled = (close ? completionCount > 0 : compensationCount > 0);
+        // Complete or Compensate methods MUST not be repeated as the resource contains a Status method
+        boolean wasCalled = (close ? completionCount == 1 : compensationCount == 1);
         boolean wasNotCalled = (close ? compensationCount == 0 : completionCount == 0);
         String lraMode = (close ? "close" : "cancel");
         String participantMode = (close ? "complete" : "compensate");
 
-        assertTrue(String.format("acceptTest with %s: participant (%s) was not asked to %s",
+        assertTrue(String.format("acceptTest with %s: participant (%s) was not asked to %s (expecting only one call)",
                 lraMode, resourcePath.getUri(), participantMode),
                 wasCalled);
         assertTrue(String.format("acceptTest with %s: participant (%s) was asked to %s",
@@ -533,16 +534,16 @@ public class TckTests extends TckTestBase {
             lraClient.closeLRA(lra);
             lraTestService.waitForCallbacks(lra);
             
-            assertTrue(methodName + ": resource should have completed once with no compensations",
-                    1 == lraMetricService.getMetric(LRAMetricType.Completed, lra, ParticipatingTckResource.class.getName()) &&
+            assertTrue(methodName + ": resource should have completed with no compensations",
+                    1 <= lraMetricService.getMetric(LRAMetricType.Completed, lra, ParticipatingTckResource.class.getName()) &&
                     0 == lraMetricService.getMetric(LRAMetricType.Compensated, lra, ParticipatingTckResource.class.getName()));
         } else {
             lraClient.cancelLRA(lra);
             lraTestService.waitForCallbacks(lra);
 
-            assertTrue(methodName + ":: resource should have compensated once with no completions",
+            assertTrue(methodName + ":: resource should have compensated with no completions",
                     0 == lraMetricService.getMetric(LRAMetricType.Completed, lra, ParticipatingTckResource.class.getName()) &&
-                    1 == lraMetricService.getMetric(LRAMetricType.Compensated, lra, ParticipatingTckResource.class.getName()));
+                    1 <= lraMetricService.getMetric(LRAMetricType.Compensated, lra, ParticipatingTckResource.class.getName()));
         }
     }
 
@@ -565,14 +566,14 @@ public class TckTests extends TckTestBase {
 
             assertTrue("joinWithTwoResourcesWithClose: both resources should have completed",
                     lraMetricService.getMetric(LRAMetricType.Completed, lra, LraResource.class.getName()) == 1 &&
-                    lraMetricService.getMetric(LRAMetricType.Completed, lra, ParticipatingTckResource.class.getName()) == 1);
+                    lraMetricService.getMetric(LRAMetricType.Completed, lra, ParticipatingTckResource.class.getName()) >= 1);
         } else {
             lraClient.cancelLRA(lra);
             lraTestService.waitForCallbacks(lra);
 
             assertTrue("joinWithTwoResourcesWithCancel: both resources should have compensated",
                     lraMetricService.getMetric(LRAMetricType.Compensated, lra, LraResource.class.getName()) == 1 &&
-                    lraMetricService.getMetric(LRAMetricType.Compensated, lra, ParticipatingTckResource.class.getName()) == 1);
+                    lraMetricService.getMetric(LRAMetricType.Compensated, lra, ParticipatingTckResource.class.getName()) >= 1);
         }
     }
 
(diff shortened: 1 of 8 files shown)


xstefank:
@ochaloup the answer is already in the PR description - Complete/Compensate to "at least once" where there is no Status method in the resource -- all of the tests that you mentioned invoke resources which have a Status method.

ochaloup:
@xstefank oh... I see. Thanks!

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -402,7 +402,7 @@ private void joinAndEnd(boolean close, String path, String path2)
         int completionCount = lraMetricService.getMetric(LRAMetricType.Completed, lra, LraResource.class.getName());
         int compensationCount = lraMetricService.getMetric(LRAMetricType.Compensated, lra, LraResource.class.getName());
 
-        boolean wasCalled = (close ? completionCount > 0 : compensationCount > 0);
+        boolean wasCalled = (close ? completionCount == 1 : compensationCount == 1);
```
mmusgrov: You cannot guarantee that messages will not be replayed. In a distributed system "at least once" semantics is normal (the way the test was originally coded). "Exactly once" (the way you have just coded the test) is not feasible. And "at most once" is not a useful delivery guarantee for our spec so we don't need to discuss it.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -530,16 +530,16 @@ private void joinWithOneResource(String methodName, boolean close, String resour
             lraClient.closeLRA(lra);
             lraTestService.waitForCallbacks(lra);
             
-            assertTrue(methodName + ": resource should have completed once with no compensations",
```
mmusgrov: Can we keep this text since it provides more information that the changed text

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -530,16 +530,16 @@ private void joinWithOneResource(String methodName, boolean close, String resour
...
-                    1 == lraMetricService.getMetric(LRAMetricType.Completed, lra, ParticipatingTckResource.class.getName()) &&
+            assertTrue(methodName + ": resource should have completed with no compensations",
+                    1 <= lraMetricService.getMetric(LRAMetricType.Completed, lra, ParticipatingTckResource.class.getName()) &&
                     0 == lraMetricService.getMetric(LRAMetricType.Compensated, lra, ParticipatingTckResource.class.getName()));
         } else {
             lraClient.cancelLRA(lra);
             lraTestService.waitForCallbacks(lra);
 
-            assertTrue(methodName + ":: resource should have compensated once with no completions",
```
mmusgrov: Same comment

xstefank:
@mmusgrov This PR has two approvals so I will merge it later today. This is just a notification in case you would still like to take a look.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/RecoveryResource.java
```diff
@@ -115,13 +115,13 @@ public Response phase2(@HeaderParam(LRA_HEADER) URI lraId) {
 
         // assert compensate has been called
         int compensations = lraMetricService.getMetric(LRAMetricType.Compensated, lraId, RecoveryResource.class.getName());
-        if (compensations != 1) {
+        if (compensations < 1) {
```
mmusgrov: compensations == 0 is more efficient (and readable)

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/RecoveryResource.java
```diff
@@ -115,13 +115,13 @@ public Response phase2(@HeaderParam(LRA_HEADER) URI lraId) {
...
             return assertFailedResponse("Compensate on restarted service should have been called. Was " + compensations);
         }
 
         // assert after LRA has been called
         int afterLRACalls = lraMetricService.getMetric(LRAMetricType.Cancelled, lraId, RecoveryResource.class.getName());
-        if (afterLRACalls != 1) {
+        if (afterLRACalls < 1) {
```
mmusgrov: afterLRACalls == 0 is more efficient (and readable)

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
xstefank: this method is used only in acceptCloseTest and acceptCancelTest which both invoke LraResource which has a Status method. The spec says "But if the method `compensatePreviousAction` should not be called a
second time (i.e. it is not idempotent) then the participant has the option of
reporting its` progress using the `@Status` annotation" (since there is no better citation to be used, we should add text that explicitly states it).

The Status method is in my understanding (and above citation) a guarantee that the implementation will not invoke Complete or Compensate methods several times since they don't need to be idempotent. Are you saying that LRA does not provide this guarantee?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
xstefank: what do you mean? Now it doesn't have to complete only *once* so this text is no longer valid.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/RecoveryResource.java
xstefank: `LRAMetricService$getMetric` can return -1 if the Compensate metrics is not incremented.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/RecoveryResource.java
xstefank: same comment as above

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/RecoveryResource.java
mmusgrov: I missed that.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: I am saying that it is not obvious why the assert is asserting only 1 call without me having to inspect the resources and consult the specification. The assert message should describe why it only expects 1 callback.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: It does need to complete so the text still applies.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/RecoveryResource.java
mmusgrov: Ah yes, -1 is a valid return code.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: Ah I see you have added the deleted line back in again so I shall resolve this conversation.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: I see that you have added the deleted line back in again so I will mark this conversation as resolved.

xstefank:
@mmusgrov added a comment and a message to the test that requires only a single complete/compensate call.

PR #284, opened 2020-03-16
Clean up unused TCK

xstefank:
Some small clean up of unused resources in the TCK.


tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/Util.java
```diff
@@ -19,150 +19,15 @@
...
-     * @return  adjusted value by factor
-     */
-    public static int adjust(int value, double factor) {
-        return Math.toIntExact(adjust((long) value, factor));
-    }
-
-    /**
-     * see {@link Util#adjust(int, double)}
-
      * @param value  value to be adjusted
```
mmusgrov: The method `adjust` is only used in one place TckTestBase. Can you move the method into that file and then delete Util.java

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricType.java
```diff
@@ -19,26 +19,20 @@
  *******************************************************************************/
 package org.eclipse.microprofile.lra.tck.service;
 
-import org.eclipse.microprofile.lra.annotation.LRAStatus;
-import org.eclipse.microprofile.lra.annotation.ParticipantStatus;
-
 public enum LRAMetricType {
```
mmusgrov: I am pretty sure there is some code somewhere in the TCK that requires the enum values to match LRAStatus enum names. Can you reinstate the enum values please.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/Util.java
ochaloup: @mmusgrov I was checking that when I was reviewing the code and I was not able to find such place. Can you please elaborate where is it?

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricType.java
xstefank: @mmusgrov I can, but since all values were just copying the same names as the enum values why can't just use enum's `name()` method?

ochaloup: @mmusgrov I put my comment to the wrong place. I would like mention that current implementation of `LRAMetricType` uses the constructor in form of
```
LRAMetricType(String label) {
}
```
which means the enum needs to define the `label` as a `String` but the label is lost after construction. There is no reason to have that string representations in the enum class when they are not used and they can't be used either.

xstefank: @mmusgrov did that answer your point? Can we merge this?

xstefank:
@mmusgrov we have 2 approvals on this PR. I am going to merge this later today so this is just a notification if you want to take another look.

xstefank:

Issue #283, opened 2020-03-12
Document the behavior of failed Complete/Compensate calls with Status method present in the resource


xstefank:
If there is a Status method present in the resource the specification assumes that Complete/Compensate methods are not idempotent. This means that the implementation should not invoke the Complete/Compensate repeatedly in case the Status method is available.

However, if the Complete/Compensate is invoked successfully but the response is lost, the implementation may assume that the Complete/Compensate call didn't happen at all and try to reinvoke these methods which would be an erroneous behavior. Also, the user needs to know that this kind of situation can happen and in case the Status is called before Complete/Compensate to invoke the relevant operation themselves.

I believe this needs to be documented scenario in the spec text.

rdebusscher:
I have some remarks on your reasoning.

You mention "in the case the complete/compensate is called successfully, but the response is lost" this means the actual logic within those methods are executed and thus there is no issue in calling the status method.

But clarification could be needed in the case the implementation calls the complete/compensate methods but those methods are actually never been called.
I think we can say that when a HTTP status is returned, the method is executed and it is safe to call Status. Otherwise (like with connection errors) those methods are re-executed by the implementation.
For the non-JAX-RS case, I need to have a look what should be the correct behaviour.

xstefank:
@rdebusscher I am aiming more to not repeat the execution of the Complete/Compensate (eg. Compensate call times out and the implementation retries it instead of calling status).

> You mention "in the case the complete/compensate is called successfully, but the response is lost" this means the actual logic within those methods are executed and thus there is no issue in calling the status method.

There is no issue in calling Status at any point in LRA execution. I am aiming at not repeating Complete/Compensate when the logic in them is executed already once before.

> I think we can say that when a HTTP status is returned, the method is executed and it is safe to call Status. Otherwise (like with connection errors) those methods are re-executed by the implementation.

+1 I like this reasoning.

> For the non-JAX-RS case, I need to have a look what should be the correct behaviour.

Non-JAX-RS case should be handled by the implementation. The in-VM call to non-JAX-RS method can only fail with exception and this behavior is described by non-JAX-RS cases contract. So this shouldn't be a problem. Repeated calls should be detected by the implementation.

mmusgrov:
If the implementation doesn't know the outcome of the complete/compensate call then it checks the status method. If the participant missed the call it reports `ParticipantStatus.Active` otherwise it reports the actual status. I don't see why we need new text. Which bit of the spec is unclear.

[...]

PR #285, opened 2020-03-16
Fix the contract in Status javadoc

xstefank:
No issue needed.

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
@@ -124,7 +124,7 @@ import java.lang.annotation.Target;
  * after it has called the Complete or Compensated method and received a response which indicates
  * that the process is in progress (with a return code 202, for example). The response code 410
  * which is received when calling this Status annotated method, MUST be interpreted by the implementation
- * that the process is successfully completed and the participant already forget about the LRA.
+ * that the process is successfully completed and the participant has already forgotten about the LRA.
  * </p>
  */
 @Retention(RetentionPolicy.RUNTIME)


api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
```diff
@@ -124,7 +124,7 @@
  * after it has called the Complete or Compensated method and received a response which indicates
  * that the process is in progress (with a return code 202, for example). The response code 410
  * which is received when calling this Status annotated method, MUST be interpreted by the implementation
- * that the process is successfully completed and the participant already forget about the LRA.
+ * that the process is successfully completed and the participant has already forgotten about the LRA.
```
mmusgrov: Why a new PR for text that is already grammatically correct.

xstefank: @mmusgrov in https://github.com/eclipse/microprofile-lra/issues/283#issuecomment-599542365 it seems that the last statement is not clear enough. I am not a native speaker but it seems to me that the wrong tense is used. @ochaloup stated there:

> the last sentence of the paragraph on the Status says that This enables the participant to free up resources.. That statement states that returning 410 or null makes participant to free up resources... right (do I misread it)?

xstefank: @mmusgrov - @ochaloup corrected me that I overlooked where exactly the unclear text was so I opened https://github.com/eclipse/microprofile-lra/pull/286. I still think that the tense here is incorrect but it's not something that I need to push. So if you think my interpretation for this fix is incorrect, go ahead and close this PR.

ochaloup: Yes. There was a not well written by me at #283 what I consider as a problem. By me I think this PR may be closed. But I consider the #286 should be merged.

xstefank:
Outvoted, closing.

api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
mmusgrov: > @mmusgrov - @ochaloup corrected me that I overlooked where exactly the unclear text was so I opened #286. I still think that the tense here is incorrect but it's not something that I need to push. So if you think my interpretation for this fix is incorrect, go ahead and close this PR.

@xstefank So the tense used in the sentence is correct from a grammatical stance. But I agree that your text is correct too and probably easier for a non native speakers. If the change had been part of a larger changeset then the PR would have been fine, but I declined the PR because if we allow too many PRs for these kind of stylistic changes we are going to become overloaded.

If you want to go through the whole document and make changes that would make the grammar easier to parse for non native speakers then I would support that. But bear in mind that any changes that you propose still have to make sense for native speakers too.

xstefank: I plan to do such sweep still before 1.0.

PR #286, opened 2020-03-17
Fix unclear indication about participant resource holding

xstefank:
No issue needed.

It was pointed in [this comment](https://github.com/eclipse/microprofile-lra/issues/283#issuecomment-599450451) that the last sentence of the Status paragraph is ambiguous.

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -1152,7 +1152,8 @@ If the participant has already responded successfully to an `@Compensate`
 or `@Complete` method invocation then it MAY report `410 Gone`
 HTTP status code or in the case of non-JAX-RS method returning
 <<source-ParticipantStatus,ParticipantStatus>> `null`.
-This enables the participant to free up resources.
+This enables the participant to free up resources if it
+hasn't done so previously.
 
 [[forgetting-an-lra]]
 ==== Forgetting an LRA


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,7 +1152,8 @@ If the participant has already responded successfully to an `@Compensate`
 or `@Complete` method invocation then it MAY report `410 Gone`
 HTTP status code or in the case of non-JAX-RS method returning
 <<source-ParticipantStatus,ParticipantStatus>> `null`.
-This enables the participant to free up resources.
+This enables the participant to free up resources if it
+hasn't done so previously.
```
mmusgrov: How can one free up resources that have already been freed. This new text is a tautology.
If we keep adding unnecessary text then the spec will become unreadable and unimplementable by anyone other than ourselves.

xstefank: What tautology? I made a change that is saying you can free up if you haven't done so yet. Or is it really meaning something different? In every case, this sentence is misleading and wrong at this place as it suggests that the participant MUST keep resources until it is called again (and can respond with 410) so we should either remove it or make the suggested change.

ochaloup: @xstefank @mmusgrov by me it's not about the length of the text but about preciseness. I consider this as a good addition which provides better understandability to anyone.

xstefank:
@mmusgrov we have 2 approvals on this PR. I am going to merge this later today so this is just a notification if you want to take another look.

xstefank:

PR #288, opened 2020-03-20
Updated to the path for the LRA.java

tomjenkinson:


xstefank:
@tomjenkinson Thanks, nice catch. You need to sign the ECA (link is in the failed check) otherwise we can't merge the PR.

xstefank:
Thanks @tomjenkinson. I agree the line 40 is better. I will leave this open for other reviews for a while. I will merge this later today.

xstefank:
Not sure if we can automate this. We should always make sure that the links in the specification document are changed when the API classes change. I will try to remember this when reviewing PRs.

PR #289, opened 2020-03-23
[#283] elaboration on @Status method when network error happens

### LINKED ISSUE #283: Document the behavior of failed Complete/Compensate calls with Status method present in the resource

ochaloup:
This is my concept of what should be elaborated regarding of the `@Status` method on failure invocation of the `Compensate/(Complete)` method.

I think the text about the freeing up resources should be removed completely as Mike mentioned in the comment
https://github.com/eclipse/microprofile-lra/issues/283#issuecomment-600218482

The text is not straightforwardly brief but I think this is about to provide an intuition on the behaviour and document the spec actions based on it. I took the inspiration from the part about reactive support (https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#3213-reactive-support) and their examples on AWS and similar.

Let me know what you think.

fixes #283

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -1152,8 +1152,34 @@ If the participant has already responded successfully to an `@Compensate`
 or `@Complete` method invocation then it MAY report `410 Gone`
 HTTP status code or in the case of non-JAX-RS method returning
 <<source-ParticipantStatus,ParticipantStatus>> `null`.
-This enables the participant to free up resources if it
-hasn't done so previously.
+
+Let's provide some intuition on this matter with the following example.
+
+Let's have an LRA participant containing methods annotated with `@Compensate` and `@Status`.
+The LRA is cancelled and the LRA implementation tries to invoke the participant's `@Compensate` method.
+At that time a network failure happens. The LRA implementation receives a network error as the result of the call.
+Now it cannot be sure if the `@Compensate` call passed through and participant has already compensated the work
+and only the response was lost or if the `@Compensate` call has not reached the participant at all.
+As the participant defines the `@Status` method the LRA implementation MUST invoke it to find the participant's true status
+(invocation can be processed several times until succeeds).
+The participant's `@Status` method is expected to indicate whether the `@Compensate` was successfully processed.
+
+* If the participant processed the `@Compensate` call it may <<jaxrs-participant-methods, has already forgotten about the LRA identification>>.
+  Then for JAX-RS participant, the response is `410 Gone`. For non-JAX-RS participant, the method returns `null`.
+  The implementation MUST assume that all required actions are performed and the implementation MUST NOT repeat the invocation.
+* If the participant compensated successfully but it has not erased the notion of the LRA identity yet it may return
+  <<source-ParticipantStatus,ParticipantStatus>> of `Compensated`. For JAX-RS participant it's a body of the response of return code `200`.
+  For the non-JAX-RS participant case, consult the section <<non-jaxrs-participant-methods>> (one option is to provide `Compensated` as the method return value).
+  The implementation MUST NOT repeat the invocation of the `@Compensate` method.
+* If the `@Compensate` method was invoked initially but the compensation logic failed the participant status is `FailedToCompensate`.
+  The LRA implementation then has to <<the-model,log sufficient information>> in such situation.
+  For JAX-RS participant, the response is `2OO OK` with participant status sent in the response body.
+  For non-JAX-RS participant case consult section <<non-jaxrs-participant-methods>> about return values.
+  The implementation MUST NOT repeat the invocation of the `@Compensate` method.
+* If the participant has not yet compensated (ie. the initial invocation of `@Compensate` did not go through)
+  then the `ParticipantStatus` will indicate the participant status with `Active`.
+  The implementation MUST repeat the invocation of the `@Compensate` method.
+
 
 [[forgetting-an-lra]]
 ==== Forgetting an LRA
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java
@@ -76,7 +76,6 @@ import java.lang.annotation.Target;
  * of the <code>&#64;Compensate</code> or <code>&#64;Complete</code> method then it may
  * report <code>410 Gone</code> HTTP status code or in case of
  * non-JAX-RS method returning {@link ParticipantStatus} to return <code>null</code>.
- * This enables the participant to free up resources.
  * </p>
  *
  * <p>


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,31 @@ If the participant has already responded successfully to an `@Compensate`
 or `@Complete` method invocation then it MAY report `410 Gone`
 HTTP status code or in the case of non-JAX-RS method returning
 <<source-ParticipantStatus,ParticipantStatus>> `null`.
-This enables the participant to free up resources if it
-hasn't done so previously.
+
+Let's provide some intuition on this matter with the following example.
+
+Let's have an LRA participant annotated with `@Compensate` and `@Status`.
```
```suggestion
Let's have an LRA participant containing methods annotated with `@Compensate` and `@Status`.
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,31 @@ If the participant has already responded successfully to an `@Compensate`
...
+The LRA is cancelled and the LRA implementation tries to invoke the participant's `@Compensate` method.
+At that time a network failure happens. The LRA implementation receives a network error as the result of the call.
+Now it cannot be sure if the `@Compensate` call pass through and participant has already compensated the work
```
```suggestion
Now it cannot be sure if the `@Compensate` call passed through and participant has already compensated the work
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,31 @@ If the participant has already responded successfully to an `@Compensate`
...
+and only the response was lost. Or if the `@Compensate` call has not reached the participant at all.
+As the participant defines the `@Status` method the LRA implementation will invoke it to find the participant's true status
```
```suggestion
As the participant defines the `@Status` method the LRA implementation MUST invoke it to find the participant's true status
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,31 @@ If the participant has already responded successfully to an `@Compensate`
...
+(invocation can be processed several times until succeeds).
+The participant's `@Status` method is expected to indicates whether the `@Compensate` was successfully processed.
+
+* If the participant processed the `@Compensate` call it may <<jaxrs-participant-methods,forget about the LRA identification>> already.
```
```suggestion
* If the participant processed the `@Compensate` call it may <<jaxrs-participant-methods, has already forgotten about the LRA identification>>.
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,31 @@ If the participant has already responded successfully to an `@Compensate`
...
+  Then for JAX-RS participant, the response is `410 Gone`. For non-JAX-RS participant, the method returns `null`.
+  (The implementation MUST assume that all required actions are performed and the compensate invocation **is not** repeated.)
```
xstefank: I would also add MUST NOT

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,31 @@ If the participant has already responded successfully to an `@Compensate`
...
+* If the participant compensated successfully but it has not erased the notion of the LRA identity yet it may return
+  <<source-ParticipantStatus,ParticipantStatus>> of `Compensated`. For JAX-RS participant it's a body of the response of return code `200`
+  for the non-JAX-RS participant, it's the returned value of the method. The compensate invocation **is not** repeated.
```
xstefank: In non-jax-rs cases, there are several ways of how to return Compensated. Maybe point to the relevant sections, please?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,31 @@ If the participant has already responded successfully to an `@Compensate`
...
+* If the participant compensated but failed during the compensation then `FailedToCompensate` status may be returned.
+  The LRA implementation then has to <<the-model,log sufficient information>> in such situation.
+  The compensate invocation will **not be** repeated.
+* If the participant has not yet compensated (ie. the initial invocation of `@Compensate` did not go through)
+  then the `ParticipantStatus` will indicate the participant status with `Active`.
+  The implementation **repeats** the invocation of the `@Compensate` method.
```
xstefank: MUST repeat

ochaloup:
@xstefank I applied your recommendations. Feel free to re-review if there is something else to adjust.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+At that time a network failure happens. The LRA implementation receives a network error as the result of the call.
+Now it cannot be sure if the `@Compensate` call passed through and participant has already compensated the work
+and only the response was lost. Or if the `@Compensate` call has not reached the participant at all.
+As the participant defines the `@Status` method the LRA implementation MUST invoke it to find the participant's true status
+(invocation can be processed several times until succeeds).
+The participant's `@Status` method is expected to indicates whether the `@Compensate` was successfully processed.
+
+* If the participant processed the `@Compensate` call it may <<jaxrs-participant-methods, has already forgotten about the LRA identification>>.
+  Then for JAX-RS participant, the response is `410 Gone`. For non-JAX-RS participant, the method returns `null`.
+  (The implementation MUST assume that all required actions are performed and the compensate invocation MUST NOT be repeated.)
```
mmusgrov: In some network setups requests are automatically replayed (MicroProfile FT for example) so I would not be this definitive.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+* If the participant compensated successfully but it has not erased the notion of the LRA identity yet it may return
+  <<source-ParticipantStatus,ParticipantStatus>> of `Compensated`. For JAX-RS participant it's a body of the response of return code `200`
```
mmusgrov: Remark: HTTP 410 GONE means the same thing.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+  for the non-JAX-RS participant, one option is returning the value from the method. For other options to return `Compensated` consult the section <<non-jaxrs-participant-methods>>.
```
mmusgrov: This sentence is difficult for an English reader to parse. If other people are able to parse it okay then please ignore this comment. But I would say that too much explanatory and or duplicated text can give the reader the impression that he is missing some subtle meaning.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+  The invocation of `@Compensate` method *is not* repeated.
```
mmusgrov: Is the *is not* meant to conform to RFC 2119?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+* If the participant compensated but failed during the compensation then `FailedToCompensate` status may be returned.
```
mmusgrov: Is this not adequately explained elsewhere. If you include this here then you should also include all of the other cases (which I don't think you do) which have already been described in the spec.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+  The LRA implementation then has to <<the-model,log sufficient information>> in such situation.
```
mmusgrov: Same comment applies here too.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+  The compensate invocation will *not be* repeated.
```
mmusgrov: You have just stated this behaviour on line 1169 and also on line 1173.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+* If the participant has not yet compensated (ie. the initial invocation of `@Compensate` did not go through)
+  then the `ParticipantStatus` will indicate the participant status with `Active`.
```
mmusgrov: Aren't we just repeating what we already have in the state model.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+  The implementation MUST repeat the invocation of the `@Compensate` method.
```
mmusgrov: I agree but it does not read very well. Also this information should already be present elsewhere in the specification.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
-This enables the participant to free up resources if it
-hasn't done so previously.
+
+Let's provide some intuition on this matter with the following example.
+
+Let's have an LRA participant containing methods annotated with `@Compensate` and `@Status`.
+The LRA is cancelled and the LRA implementation tries to invoke the participant's `@Compensate` method.
+At that time a network failure happens. The LRA implementation receives a network error as the result of the call.
+Now it cannot be sure if the `@Compensate` call passed through and participant has already compensated the work
+and only the response was lost. Or if the `@Compensate` call has not reached the participant at all.
```
```suggestion
and only the response was lost or if the `@Compensate` call has not reached the participant at all.
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,32 @@ If the participant has already responded successfully to an `@Compensate`
...
+As the participant defines the `@Status` method the LRA implementation MUST invoke it to find the participant's true status
+(invocation can be processed several times until succeeds).
+The participant's `@Status` method is expected to indicates whether the `@Compensate` was successfully processed.
```
```suggestion
The participant's `@Status` method is expected to indicate whether the `@Compensate` was successfully processed.
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: These invocations are performed by the implementation so I wouldn't expect technologies like FT would be in place. This MUST be definitive because we give a guarantee with the Status method that Compensate doesn't need to be idempotent and thus it shouldn't be replayed.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: agreed,  `for the non-JAX-RS participant case, consult the section <<non-jaxrs-participant-methods>>` may be enough?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: +1 MUST NOT be repeated

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: This one should stay. I agree that the 'FailedToCompensate' section can be removed. But for the example purpose, this was exactly what I was looking for when creating the issue.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: You cannot guarantee that they won't be replayed.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: SHOULD NOT might be more implementable

mmusgrov:
So this issue is really just asking for more text to cover a spec item that is already well stated in the specification.

The number of comments generated by the associated PR is an indication that every piece of text in a spec should also be viewed as an opportunity for adding confusion (as well as for clarification). Scattering text that addresses the same topic throughout the specification will not aid it's comprehension.

Rather than adding new sections it would be preferable to clarify the section where the topic was first introduced. Or, if it's examples of usage that we want then we should clarify the examples section.

xstefank:
@mmusgrov A  lot of these comments are styling/grammar so I wouldn't take that as an argument. I don't necessarily disagree but we can restructure the spec on the final take. I find this section as a good example and it covers the scenario which I don't think is covered anywhere else -- network failures.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: Yeap, thanks for the remark.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: I changed the text.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: ok

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: Agree. I elaborated the text.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: What is the missing piece here? (The details on the behaviour can be checked in the referred section.)

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: For the clear information it's good to provide precise declaration of the behaviour. I think the text should state this here again.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: I think the `FailedToCompensate` should NOT be removed. The point is to provide whole spectrum of the behaviours. It's essential idea of this example.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: You are more than welcome to provide me a suggestion for fixing my stylistic incompetence :-)

ochaloup:
I fully agree with @mmusgrov's point that there are multiple sections where the text clarification would be beneficial.
I don't agree that adding this paragraph is opportunity for confusion. I takes the important part of the behaviour, construct an example and provides a guidepost to other part of the documentation.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
ochaloup: I agree with the @mmusgrov  that we can't guarantee the replay does happen. The text should be written from the perspective of the specification. I rephrased it.

As we don't know if the system uses some replay processing framework and as the message order can't be ensured (like compensate http call is invoked, there is some failure in the network and the call is buffered on some router, the initial invocation may timed-out (or coordinator crashes) and the status+compensate invocation is repeated, then suddenly the initial http request could be processed on the router and the duplicity occurs).

I didn't want to make the text complex initially, now I added a note from this perspective at the end.

ochaloup: After discussing and re-thinking the point I'm leaving just the initial text without any additional note.

My perspective remains here - with status method the implementation has to guarantee that the invocation is not  repeated. The text is already rewritten in this regard.
I agree with @xstefank that specification needs to be definitive in this regard.

I'm in agreement with @mmusgrov's comment (if I have the correct understanding) that specification can't guarantee the replay does not happen. But the specification must guarantee that it's not replayed by the implementation. As mentioned, the text was rewritten in this regard.

ochaloup:
@mmusgrov @xstefank @rdebusscher this PR is stalled for a week now. What is a general consideration about the issue? Should it be fixed? In this (are there some changes needed to be done?) or in a different way (to close this PR and let it for somebody else)?
Please, let me know.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1152,8 +1152,34 @@ If the participant has already responded successfully to an `@Compensate`
...
+(invocation can be processed several times until succeeds).
+The participant's `@Status` method is expected to indicate whether the `@Compensate` was successfully processed.
+
+* If the participant processed the `@Compensate` call it may <<jaxrs-participant-methods, has already forgotten about the LRA identification>>.
+  Then for JAX-RS participant, the response is `410 Gone`. For non-JAX-RS participant, the method returns `null`.
+  The implementation MUST assume that all required actions are performed and the implementation MUST NOT repeat the invocation.
+* If the participant compensated successfully but it has not erased the notion of the LRA identity yet it may return
+  <<source-ParticipantStatus,ParticipantStatus>> of `Compensated`. For JAX-RS participant it's a body of the response of return code `200`.
+  For the non-JAX-RS participant case, consult the section <<non-jaxrs-participant-methods>> (one option is to provide `Compensated` as the method return value).
+  The implementation MUST NOT repeat the invocation.
```
xstefank, line 1173: here is not clear whic invocation you mean - status or compensate. Status can be replayed but it doesn't make sense.

ochaloup: I updated the text to clarified that's the `@Compensate` method.

mmusgrov:
> @mmusgrov @xstefank @rdebusscher this PR is stalled for a week now. What is a general consideration about the issue? Should it be fixed? In this (are there some changes needed to be done?) or in a different way (to close this PR and let it for somebody else)?
> Please, let me know.

I have given my views so I shall abstain. Since you have sufficient approval I would just go ahead and merge.

PR #291, opened 2020-03-30
Use ArquillianResource instead of lra.tck.base.url

idlewis:
For app servers which don't deploy the arquillian tests to the same
context root each time, the base url will change for every test.
Hence for these app servers (such as Open Liberty), setting a single
base url property is unreliable.


mmusgrov:
@rdebusscher What do you think of @idlewis 's PR?

rdebusscher:
@mmusgrov Not familiar with Arquillian, so if it works for Narayana I guess it is ok.

mmusgrov:
@idlewis  Thanks for the contribution

PR #292, opened 2020-04-10
Fix TckRecoveryTests deployment URL handling

xstefank:
With deployment URL handling reworked in #291 the deployment URL is taken now as an Arquillian resource not from the MP config. The issue in TckRecoveryTests is that the deployments are not managed by arquillian and thus the deployment URL is not injected. This fix passes the deployment URL as a header to invocations in the individual phases that are run in the manually started containers (deployments not managed by arquillian).

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckRecoveryTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckRecoveryTests.java
@@ -56,6 +56,8 @@ import java.util.logging.Logger;
  */
 @RunWith(Arquillian.class)
 public class TckRecoveryTests {
+    
+    public static final String LRA_TCK_DEPLOYMENT_URL = "LRA-TCK-Deployment-URL";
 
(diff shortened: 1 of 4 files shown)


xstefank:
@ochaloup also review please.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckRecoveryTests.java
```diff
@@ -56,6 +56,8 @@
  */
 @RunWith(Arquillian.class)
 public class TckRecoveryTests {
+    
+    public static final String DEPLOYMENT_URL_HEADER = "Deployment-URL";
```
mmusgrov, line 60: Why not prefix the header name with `LRA` or `LRA_TCK` to avoid conflicting with any existing headers or to easily see that it is one of ours.

xstefank:
fixed only the @mmusgrov suggestion.

xstefank:
@rdebusscher Rudy, I would like to merge this later today. Can you take a look, please?

PR #293, opened 2020-04-15
issue274 Unignore and fix TckTests#mixedMultiLevelNestedActivity

### LINKED ISSUE #274: Conflict in TCK: unable to cancel already closed nested LRA

xstefank:
resolves #274

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
@@ -22,6 +22,7 @@ package org.eclipse.microprofile.lra.tck;
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
 import static org.eclipse.microprofile.lra.tck.participant.api.AfterLRAListener.AFTER_LRA_LISTENER_WORK;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.ACCEPT_WORK;
+import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.CANCEL_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.LRA_RESOURCE_PATH;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.TIME_LIMIT;
 import static org.eclipse.microprofile.lra.tck.participant.api.LraResource.TIME_LIMIT_HALF_SEC;
@@ -64,7 +65,6 @@ import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 
@@ -178,7 +178,6 @@ public class TckTests extends TckTestBase {
     }
 
     @Test
-    @Ignore("https://github.com/eclipse/microprofile-lra/issues/274")
     public void mixedMultiLevelNestedActivity() throws WebApplicationException {
         multiLevelNestedActivity(CompletionType.mixed, 2);
     }
@@ -635,11 +634,11 @@ public class TckTests extends TckTestBase {
             lraClient.closeLRA(lra);
         } else {
             /*
-             * The test is calling for a mixed outcome (a top level LRA L! and nestedCnt nested LRAs (L2, L3, ...)::
-             * L1 the mandatory call (PUT "activities/multiLevelNestedActivity") registers participant C1
-             *   the resource makes nestedCnt calls to "activities/nestedActivity" each of which create nested LRAs
+             * The test is calling for a mixed outcome (a top level LRA L1 and nestedCnt nested LRAs (L2, L3, ...)::
+             * L1 the mandatory call (PUT "lraresource/multiLevelNestedActivity") registers participant C1
+             * the resource makes nestedCnt calls to "lraresource/nestedActivity" each of which create nested LRAs
              * L2, L3, ... each of which enlists a participant (C2, C3, ...) which are completed when the call returns
-             * L2 is cancelled  which causes C2 to compensate
+             * L2 is cancelled which causes C2 to compensate
              * L1 is closed which triggers the completion of C1
              *
              * To summarise:
@@ -648,7 +647,14 @@ public class TckTests extends TckTestBase {
              * - C2 is completed and then compensated
              * - C3, ... are completed
              */
-            lraClient.cancelLRA(uris[1]); // compensate the first nested LRA
+
+            // compensate the first nested LRA in the enlisted resource
+            tckSuiteTarget
+                .path(LRA_RESOURCE_PATH).path(CANCEL_PATH)
+                .request()
+                .header(LRA_HTTP_CONTEXT_HEADER, uris[1])
+                .put(Entity.text(""));
+
             lraClient.closeLRA(lra); // should not complete any nested LRAs (since they have already completed via the interceptor)
         }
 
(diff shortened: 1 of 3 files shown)


xstefank:
@rdebusscher Rudy, I would like to merge this later today. Can you take a look, please?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -648,7 +647,14 @@ private void multiLevelNestedActivity(CompletionType how, int nestedCnt) throws
              * - C2 is completed and then compensated
              * - C3, ... are completed
              */
-            lraClient.cancelLRA(uris[1]); // compensate the first nested LRA
+
+            // compensate the first nested LRA in the enlisted resource
+            tckSuiteTarget
```
rdebusscher, line 652: This is different then what I always had in mind when we discussed nested LRAs.

Because here we explicitly call an endpoint which was used as a nested call.

What I had in mind (and what I though we were discussing) is that @Compensate can be called for a nested LRA although @Complete is already called.

'Parent LRA' 
    -> nested1 -> success -> @Complete for C1.
    -> nested2 -> success -> @Complete for C2.
    -> nested3 -> Failure
  -> @Compensate for C1 and C2 need to be called.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -648,7 +647,14 @@ private void multiLevelNestedActivity(CompletionType how, int nestedCnt) throws
...
+                .path(LRA_RESOURCE_PATH).path(CANCEL_PATH)
+                .request()
+                .header(LRA_HTTP_CONTEXT_HEADER, uris[1])
```
rdebusscher, line 655: We call it without specifying it as a 'nested call' (not using LRA_HTTP_PARENT_CONTEXT_HEADER)

So the first invocation is as nested, the second is a top level. For me this means that this are 2 different situations and thus another participant is created.

rdebusscher:
I fear that some of the nested LRA scenarios aren't understood enough at this time and will cause great confusion

- allowed to call endpoint again with already closed (nested) LRAid
- Is call with and without LRA_HTTP_PARENT_CONTEXT_HEADER the same invocation?
- which situation exactly trigger the invocation of @Complete and @Compensate (for the same participant)
- the (almost required) definition of a @Forget, etc ...)
- ...

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
xstefank: This is correct, but this test tests different scenario. Namely, that the **user** can cancel already closed nested LRA.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
xstefank: That shouldn't be the case, because uris[1] was created as a nested LRA so the implementation must know that it is nested. This is an end-user call so I would like to avoid adding a requirement for including parent context.

xstefank:
@rdebusscher I agree that the nested scenarios are a little overcomplicated, however, they are specified so they should be consumable for users. But I myself had quite a hard time understanding all possible nesting outcomes. I don't expect that this feature will be heavily used in the beginning.

This PR is in accordance with the definition that we all agreed on in https://github.com/eclipse/microprofile-lra/pull/276.

rdebusscher:
@xstefank But as I mentioned in my last comment on #276 

>I don't think you need to enlist in the case of the nested case going from Closed to cancelling.

I did not think about the user explicitly calling an endpoint which is already enlisted as a nested LRA again.
Because in this example it does a cancel, but the endpoint can just be called successfully, and in that case, is the @Complete called a second time? (corresponding to the second enlistment) This is not specified in the spec if you ask me.

The spec says that participant can only be enlisted once, so only 1 set of calls to @Complete (because multiple calls are possible)
This would mean we have a whole different set of rules related to nested LRA.

xstefank:
@rdebusscher actually I mentioned this case in #276 and it is specified -- https://github.com/eclipse/microprofile-lra/pull/276/files#diff-0d4b5f4f2a4183d5dfb813a9cc447a0eR301-R303.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
mmusgrov: You are correct @xstefank in that we added the parent context header so that the implementation can provide it during participant callbacks (Compensate, Complete, Status and Forget).

But @rdebusscher does make a good point since the [text we added to the LRA annotation](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L117) to describe the header could be interpreted in the way Rudy suggests (even though we did not intend for it to be interpreted in that way). So can we update the LRA annotation text to be more specific. A small change to the first 3 lines of that javadoc would cover it, something like the following:
 
``` java
    /**
     * When an implementation of the LRA specification invokes any of the participant
     * callbacks (namely {@link Compensate}, {@link Complete}, {@link Status} and
     * {@link Forget}) in the context of a nested LRA it must ensure that
     * the parent LRA is made available via an HTTP header field with the
     * following name. The value contains the parent LRA id associated with
     * the HTTP request/response and is of type {@link java.net.URI}.
     */
```

xstefank: @mmusgrov @rdebusscher included this suggestion.

mmusgrov: Thanks. @rdebusscher does this update to the @LRA javadoc resolve your issue?

rdebusscher: Updated JavaDoc is about the Participants callback. But my comment was mainly around calling a Rest endpoint with just the LRAid (not the ParentLRAid although it is supposed to be nested)

And still, I have a bigger problem of calling a Nested Participant again which is already Closed.

But it seems I'm the only one.

ochaloup: @rdebusscher I don't like calling the nested participant again either. I think the nested processing could be difficult to grasp from the user perspective. The responsibility of handling the details of the nested processing are more with user now (but he has enough information to do it safely).
The gitter discussion contains the multiple arguments from different angles https://gitter.im/eclipse/microprofile-lra?at=5e43e045f3718e705df97aff .

While the LRA model was designed that way the change would mean breaking it just before 1.0 release.

I would be happy if there is mood to open the discussion about nested processing again and find a way how to simplify it (here especially from the user perspective).
But I ended up in state of mind to open it after 1.0 (like a topic for the LRA 2.0)

rdebusscher: If a lot of people are not confident that we are doing nested LRAs correctly, have a good understanding of all the scenarios possible with all the variations and all the implications. 

We should not leave it as it is and revise it in 2.0, but just remove.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
xstefank: Ah, I see now that I misunderstood your comment. That is indeed the current scenario.

xstefank:
After the number of discussions in the public chat room ([gitter](https://gitter.im/eclipse/microprofile-lra)) we agreed that we will include the current nested model and thus PR can be merged.

PR #295, opened 2020-04-16
Document the possible return HTTP status codes from the LRA method in…

### LINKED ISSUE #290: Clarify usage of 412 and 410 in the participant enlistment cases
We have to distinguish two cases:

* participant returning values from Compensate, Complete, Status
  * 410 means participant has no knowledge of received LRA (basically same as Compensated/Completed) -- [spec#302]( https://github.com/eclipse/microprofile-lra/blame/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#L302)
  * 412 is not allowed

* values returned from user invocations of LRA methods with included LRA context 
  * this needs clarification as per https://github.com/eclipse/microprofile-lra/pull/281

The options discussed for "values returned from user invocations of LRA methods with included LRA context":

1. use both 412 and 410
* 412 LRA context is present when it shouldn't or not present when it should or invalid context (see MANDATORY or NEVER types) and invalid received URI
* 410 expired context (valid URI but not found in the implementation resources)
  * basically closed or canceled LRA already removed from the implementation resources

2. use only 412 as it is a different use case then participant returning values
* 412 means that the context is present when it shouldn't or not present when it should **or** that the received context is expired (so no distinction) or invalid URI

Can we agree on one of the options and unify the text and test expectations across the specification and the Javadoc?

xstefank:
…vocations

resolves #290

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
@@ -61,6 +61,13 @@ import java.time.temporal.ChronoUnit;
  *   <li>to cancel the LRA context when the method returns particular HTTP
  *       status codes</li>
  * </ul>
+ * 
+ * Some of the above defined LRA operations are performed before the execution
+ * proceeds to the LRA annotated business method. If any of these LRA operations
+ * cannot complete the implementation MUST return one of the defined HTTP status
+ * codes which means that the business method is not invoked. For the definition of
+ * the allowed status codes and condition in which they are returned please
+ * see the specification document.
  *
  * <p>
  * Newly created LRAs are uniquely identified and the id is referred to as the
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -387,6 +387,13 @@ will run in the context of an LRA and controls whether or not:
 - to throw an exception if the method returns particular HTTP status codes.
 - to enlist the class as a participant of the LRA.
 
+If the incoming context has already ended or does not exist the method will 
+immediately return 410 Gone HTTP status code.
+
+Similarly, if any of the preconditions listed in the Type element
+(see <<source-LRA,LRA.Type>>) of the LRA annotation are not met then the method
+will immediately return 412 Precondition Failed HTTP status code.
+
 When the @LRA annotation is defined multiple times, this is the order of precedence
 
 - On the JAX-RS method itself.


xstefank:
@ochaloup alse review please.

xstefank:
Let me know whether having the table in the specification text and reference in JavaDoc would be more preferable. I am not sure what would be better.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -61,6 +61,27 @@
  *   <li>to cancel the LRA context when the method returns particular HTTP
  *       status codes</li>
  * </ul>
+ * 
+ * Possible return HTTP status codes are:
```
mmusgrov, line 65: Isn't this a business method. How can we stipulate what responses a business method should return?

xstefank: Through filters? This how we've defined everything so far. See for example https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L206.

mmusgrov: If we indeed do need to add anything extra then something like the following would be less confusing:
 
> When a method is about to execute in the context of an LRA but the LRA has already ended (and we add the appropriate caveat for nested LRAs) then the method is not executed and a 410 response code is returned.

And if we are not already covered for the 412 case then we should fix the relevant part of the javadoc with something like:

> When an method is about to execute in the context of an LRA and XXX/whatever is in an invalid state then the method is not executed and a 412 response code is returned.

But that is confusing too so we should we would need to be explicit.

xstefank: fair enough. I will add something similar to this PR. But I won't change the already specified 412 elsewhere in the text. I will do so when I will be doing the final sweep of the specification.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -387,6 +387,28 @@ will run in the context of an LRA and controls whether or not:
...
+|===
+|Status code |Conditions
+
+| 410
+a| - the passed LRA context has already ended
+   - the passed LRA context is not found
+
+| 412
+a| - the user passed the LRA context to LRA type which prohibits it (see LRA.Type)
+   - the content of the LRA context is invalid
```
mmusgrov, line 405: We do not mandate that implementations parse the context other than possibly validating that it conforms to [RFC 2396: Uniform Resource Identifiers (URI): Generic Syntax](https://www.ietf.org/rfc/rfc2396.txt). For example the context could have been created by another implementation and implementations should not need to understand the format used by other implementations. So if you are only thinking about URIs then its best that the text says so.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -61,6 +61,13 @@
  *   <li>to cancel the LRA context when the method returns particular HTTP
  *       status codes</li>
  * </ul>
+ * 
+ * The LRA operations are performed before the execution proceeds to the LRA
+ * annotated business method. If any of the LRA operations cannot complete the
+ * implementation MUST return one of the defined HTTP status codes which means
+ * that the business method is not invoked. For the definition of the allowed
+ * status codes and condition in which they are returned please see the 
+ * specification document.
```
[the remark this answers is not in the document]
ochaloup: I wonder if cross-referencing from the javadoc makes sense. The section numbers and titles are changed easily. For me the current text is better.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -387,6 +387,28 @@ will run in the context of an LRA and controls whether or not:
...
+
+|===
+|Status code |Conditions
+
+| 410
+a| - the passed LRA context has already ended
+   - the passed LRA context is not found
+
+| 412
+a| - the user passed the LRA context to LRA type which prohibits it (see LRA.Type)
```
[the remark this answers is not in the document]
ochaloup: @xstefank will the text `LRA.Type` be correctly generated by `javadoc`? Should not be used here something like `@link` or similar?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -387,6 +387,28 @@ will run in the context of an LRA and controls whether or not:
...
+   - the content of the LRA context is invalid
+
+| 5xx
+a| - other unspecified LRA processing errors
```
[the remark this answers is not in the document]
ochaloup: Maybe this could be articulated better but the `5xx` means an unspecified reason, doesn't it? It's something the spec does not define. We consider there could be thrown e.g. a `RuntimeException` but it's just one of possible reasons.
I think other spec uses the similar wording to express that some by spec unknown error happens (https://docs.oracle.com/javaee/7/api/javax/transaction/SystemException.html).
The wording is quite fine for me.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -387,6 +387,28 @@ will run in the context of an LRA and controls whether or not:
 - to throw an exception if the method returns particular HTTP status codes.
 - to enlist the class as a participant of the LRA.
 
+These operations are performed before the execution proceeds to the invoked
+`@LRA` annotated business method. If any of the LRA operations prohibits the
+execution of the business method, the LRA implemenation MUST return one of the
+following status codes to user invoking the LRA annotated method. This means
+that in these conditions the business method IS NOT invoked.
```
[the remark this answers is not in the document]
xstefank: I will add this to my final sweep list. Please don't block this PR for this.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -387,6 +387,28 @@ will run in the context of an LRA and controls whether or not:
 - to throw an exception if the method returns particular HTTP status codes.
 - to enlist the class as a participant of the LRA.
 
+These operations are performed before the execution proceeds to the invoked
```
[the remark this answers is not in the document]
xstefank: yes, should that be somehow interlinked?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
xstefank: thanks @ochaloup. @mmusgrov WDYT about Ondras response?

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -61,6 +61,13 @@
  *   <li>to cancel the LRA context when the method returns particular HTTP
  *       status codes</li>
  * </ul>
+ * 
+ * The LRA operations are performed before the execution proceeds to the LRA
```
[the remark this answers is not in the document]
xstefank: the above list basically. I will add "The above LRA operations" would that be better?

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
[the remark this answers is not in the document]
xstefank: Yes, I agree that keeping this information in Javadoc up to date may be problematic. The Javadoc itself cannot be enough to understand the spec. Otherwise, we don't need the specification document at all.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
xstefank: @mmusgrov I want to reference the case when user send LRA context to NEVER method or doesn't send LRA context to the MANDATORY LRA method. Can you suggest better wording for one line bulletpoint?

@ochaloup this is a specification document. Javadoc markings doesn't work here. However, I can add a link to the LRA class if that would help. However, I don't think that is necessary. Let me know.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: -1, if the user passes `http://example.com` that is also a valid URI but I don't think any implementation will use it as an identifier. We explicitly don't define interoperability in 1.0 so I think that this text is correct.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -387,6 +387,28 @@ will run in the context of an LRA and controls whether or not:
...
+
+| 412
+a| - the user passed the LRA context to LRA type which prohibits it (see LRA.Type)
+   - the content of the LRA context is invalid
+
+| 5xx
+a| - other unspecified LRA processing errors
+|===
+
+
```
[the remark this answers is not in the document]
xstefank: will take care of this in next force push.

mmusgrov:
> Let me know whether having the table in the specification text and reference in JavaDoc would be more preferable. I am not sure what would be better.

Table in the javadoc referenced from the spec.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: Well some of the "operations" are performed after the method finishes execution which implies that your statement is unrelated to the items listed above since it does not apply to the 3rd and 5th bullets.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: If we specify that 5xx has an unspecified meaning for the @LRA annotation then it would preclude using any of those codes to mean something specific in a later spec revision.

[Note that we already give the 500 code a specific semantic interpretation for participant responses to the @complete and @compensate calls - it means the response body does not correspond to a valid participant status].

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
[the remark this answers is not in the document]
mmusgrov: Well some of the bulleted items refer to actions the implementation may perform after execution has finished so not really.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: You have said elsewhere that you will be pushing changes to this PR so why wait?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: The text in this table row does not include the case where the `LRA.Type` element of the LRA annotation is MANDATORY and no context is present - ie it is a contradictory statement (that is the danger of making general references).

Slightly better wording could be:

> - the context does not match the requirements stated in the description of the `Type` element (see LRA.Type) of the LRA annotation

but that is still clunky. 

If you really do want to specify the 410 and 412 codes here I would suggest something like:

> If the incoming context has already ended or does not exist the method will immediately return 410 GONE.
>
> Similarly, if any of the preconditions listed in the *Type* element (see *LRA.Type*) of the LRA annotation are not met then the method will immediately return 412 Precondition Failed

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: I do not understand your answer. Are you saying that if we only validate that the context conforms to RFC 2396 then we would end up allowing *http://example.com*?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: Didn't you say elsewhere that you will be doing a force push to this PR. In which case adding *is not* wouldn't be too onerous.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: At most I would say *500 Internal Server Error*. The other 5xx codes mean something specific.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: yes, this is what I understood that your answer is suggesting.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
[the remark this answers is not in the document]
xstefank: ok, I will rephrase.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
xstefank: We can still change the status codes in later revisions, we are allowed to make breaking changes. But I accept your suggestion and I will change it to  500 Internal Server Error.

api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java
```diff
@@ -61,6 +61,13 @@
  *   <li>to cancel the LRA context when the method returns particular HTTP
  *       status codes</li>
  * </ul>
+ * 
+ * Some of the above defined LRA operations are performed before the execution
+ * proceeds to the LRA annotated business method. If any of these LRA operations
+ * cannot complete the implementation MUST return one of the defined HTTP status
+ * codes which means that the business method is not invoked. For the definition of
+ * the allowed  * status codes and condition in which they are returned please
```
ochaloup, line 69: is this a typo or I don't understand the meaning? I mean the `the allowed  * status`

xstefank: typo, thanks @ochaloup

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: The URI http://example.com conforms to RFC 2396 so it is valid. So my question is what does it mean for "the content of the LRA context" to be invalid?

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: We can make breaking changes but we shouldn't unless we have a good reason to.

I still do not see the benefit of specifying it though since leaving it unspecified would give us the maximum latitude to reserve the code for any future need.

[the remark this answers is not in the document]
xstefank: ok, I will remove it. It's not that important for me. I just wanted to show that even LRA code can fail with unexpected errors and in this case the execution also does not proceed to the users business method.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
xstefank: I would think yes, that is an invalid content and I would expect the implementation reply with 412 since it is a user error. Am I missing something?

rdebusscher: How can we know when a valid URI is not a valid LRAid? (from its structure, without checking if it is found internally which would be a 410 status if not found)

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: The implementation should not be throwing unspecified errors. But the infrastructure that the implementation is running on may, however it is not our responsibility to define what those are.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: +1

xstefank: ok, I will change it to 410

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: So we can sidestep all the comments raised by this issue if we go with something close to what I quoted above.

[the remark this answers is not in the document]
xstefank: so you're suggesting to replace table with text? Do I understand correctly? Since there are only two status codes now I don't have an issue with it.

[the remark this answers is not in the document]
mmusgrov: Yes that is what I am suggesting.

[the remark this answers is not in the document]
xstefank: Updated.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -387,6 +387,19 @@ will run in the context of an LRA and controls whether or not:
 - to throw an exception if the method returns particular HTTP status codes.
 - to enlist the class as a participant of the LRA.
 
+Some of the above defined LRA operations are performed before the execution proceeds
+to the invoked `@LRA` annotated business method. If any of these LRA operations
+prohibits the execution of the business method, the LRA implemenation MUST return
+one of the following status codes to user invoking the LRA annotated method. This
+means that in these conditions the business method _is not_ invoked:
```
mmusgrov, line 394: What added value does this paragraph provide? The two bullets below say everything that needs to be said. I had to read this paragraph a couple of times just to make sure I was not missing any subtle points (so adds complexity).

xstefank: I believe that we need to introduce what operations we are talking about and to put the emphasis on the method not being executed. Without this paragraph maybe the user will understand that his business method needs to return these HTTP status codes? I don't think it's adding complexity, rather it's adding context.

ochaloup: I think it's good to define that these error codes relates to case when the business method has not been executed yet.
From that perspective it's valuable to introduce those two points with some text.
I agree in way that the text is a bit hard to read. Maybe just one sentence which emphasizes the key point would be better?

xstefank: I can remove the last sentence but that the emphasis that I wanted to put there for users. Other text is important.

ochaloup: I'm not able to create some nicer rephrasing of the paragraph so that way it would be good from my perspective.

mmusgrov: @ochaloup The two bullets below (turned into two paragraphs) provides all of the information that the user needs about what error codes are returned without the method having to execute. The above paragraph adds nothing to his understanding, on the contrary it can give the reader the impression that he has missed some nuance and therefore obfuscates the meaning of the two bullets.

xstefank: @mmusgrov I see that this argument is a matter of taste but there are two opinions of non-native speakers which think that this paragraph (excluding last sentence) is adding value to the reader. 

I don't see how this reads better:

```
The annotation MUST be applied to JAX-RS annotated methods, classes, superclasses or interface methods otherwise
it has no effect. It determines whether or not the JAX-RS method
will run in the context of an LRA and controls whether or not:

- any incoming context should be suspended and if so if a new one should be started.
- to start a new LRA.
- to end any LRA context when the method ends.
- to throw an exception if there should be an LRA present on method entry.
- to throw an exception if the method returns particular HTTP status codes.
- to enlist the class as a participant of the LRA.

If the incoming context has already ended or does not exist the method will 
immediately return 410 Gone HTTP status code.

Similarly, if any of the preconditions listed in the Type element
(see <<source-LRA,LRA.Type>>) of the LRA annotation are not met then the method
will immediately return 412 Precondition Failed HTTP status code.
```

You yourself argued that I didn't define what operations I was describing. I can see where I would find different meanings if I didn't know what the paragraphs talk about.

ochaloup: @mmusgrov @xstefank For me it seemed that it's good to highlight that the checks are executed before the business method and they cause the method won't bet invoked. It's my preference but I do not have a strong argument if it's necessary.

mmusgrov: @xstefank the text you just quoted looks fine (in fact it is what I have been suggesting). What are you saying here?

@ochaloup The text I suggested does say that the checks are executed before the business method runs. To quote: "... the method will immediately return ...". Are you interpreting those 5 words to mean something other than to not execute any business logic? If you are then would "... the method will not execute and a 410 Gone HTTP status code will be returned back to the caller" be preferable?

mmusgrov: I am not labouring this point to be obstructive. I have said on a number of occasions that adding too much text or saying a similar thing in multiple places or adding superfluous explanations is making our spec quite difficult to parse.

xstefank: ok, I see that this is going nowhere and I really don't have any more time or arguments on this.

mmusgrov: You still have the text that I said was duplicating the bullets in the spec, except now its in the javadoc. But this is fine since the javadoc is allowed to repeat what is said in the main document and it is useful that the duplication is worded differently.

PR #310, opened 2020-06-08
#305 Add a TCK test which ensures compensate may called before the participant's business method finishes. 

### LINKED ISSUE #305: Add a TCK test which ensures the compensate may called before the participant's business method finishes
The goal of this issue is to create a new TCK test case which ensures that the implementation permits the call of the compensate method being processed before the `@LRA` business method ends.
In other words, the test should show that if the business method takes a long time and meanwhile the LRA is cancelled then the `@Compensate` method will be invoked. The business method may still be in progress. The `@Compensate` method could be, for example, used for interruption of the tasks being done in the business method.

This issue was created as a result of discussion at https://github.com/eclipse/microprofile-lra/issues/301. The issue #301 wrongly assumed the behaviour expected by the spec.

The test would consist (approximately) this scenario:

* LRA is started
* Aarticipant's `@LRA` business method is invoked with context of the started LRA
* The invocation of the business `@LRA` method causes the participant is registered with the started LRA
* The participant starts some business processing inside of the invoked method. This is long running action which is ensured not being finished until LRA is cancelled.
* The LRA is meanwhile cancelled. The LRA implementation has to call the compensate callback at the enlisted participant
* The `@Compensate` method has to be invoked and only after that the LRA participant business method may be finished by the test

zezulka:
resolves #305

   - add a new TCK test (as described in the title) `TckParticipantTests#cancelLraDuringBusinessMethod` and an auxiliary participant `LongBusinessMethodParticipant` used in this test.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
@@ -19,6 +19,25 @@
  *******************************************************************************/
 package org.eclipse.microprofile.lra.tck;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.net.URI;
+import java.time.temporal.ChronoUnit;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.logging.Logger;
+
+import javax.inject.Inject;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
+
+import org.eclipse.microprofile.lra.annotation.ws.rs.LRA;
+import org.eclipse.microprofile.lra.tck.participant.nonjaxrs.valid.LongBusinessMethodParticipant;
 import org.eclipse.microprofile.lra.tck.participant.nonjaxrs.valid.ValidLRACSParticipant;
 import org.eclipse.microprofile.lra.tck.participant.nonjaxrs.valid.ValidLRAParticipant;
 import org.eclipse.microprofile.lra.tck.service.LRAMetricService;
@@ -27,26 +46,18 @@ import org.eclipse.microprofile.lra.tck.service.LRATestService;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.junit.Arquillian;
 import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.junit.Assert;
 import org.junit.Test;
 import org.junit.runner.RunWith;
-
-import javax.inject.Inject;
-import javax.ws.rs.WebApplicationException;
-import javax.ws.rs.client.WebTarget;
-import javax.ws.rs.core.Response;
-
-import java.net.URI;
-
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertTrue;
 
 /**
  * TCK to verify that valid non-JAX-RS participant method signatures are respected
  */
 @RunWith(Arquillian.class)
 public class TckParticipantTests extends TckTestBase {
-    
+
     private static final String VALID_DEPLOYMENT = "valid-deploy";
+    private static final Logger LOGGER = Logger.getLogger(TckParticipantTests.class.getName());
 
     @Inject
     private LRAMetricService lraMetricService;
@@ -59,18 +70,18 @@ public class TckParticipantTests extends TckTestBase {
         return TckTestBase.deploy(VALID_DEPLOYMENT)
             .addPackage(ValidLRAParticipant.class.getPackage());
     }
-    
+
     /**
      * Test verifies that non-JAX-RS @Complete method is invoked according to the
      * LRA protocol and that if {@link WebApplicationException} is
-     * thrown inside of a non-JAX-RS participant method than {@link Response} it 
+     * thrown inside of a non-JAX-RS participant method than {@link Response} it
      * is carrying is extracted and acted upon according to LRA response handling
      */
     @Test
     public void validWebApplicationExceptionReturnedTest() {
         WebTarget resourcePath = tckSuiteTarget.path(ValidLRAParticipant.RESOURCE_PATH)
             .path(ValidLRAParticipant.ENLIST_WITH_COMPLETE);
-        
+
         Response response = resourcePath.request().get();
         URI lraId = URI.create(checkStatusReadAndCloseResponse(Response.Status.OK, response, resourcePath));
 
@@ -89,13 +100,13 @@ public class TckParticipantTests extends TckTestBase {
     }
 
     /**
-     * This test verifies chained call of non-JAX-RS participant methods. First 
+     * This test verifies chained call of non-JAX-RS participant methods. First
      * it starts and cancels a new LRA. @Compensate non-JAX-RS method then returns
-     * {@link org.eclipse.microprofile.lra.annotation.ParticipantStatus#Compensating} 
-     * (see {@link ValidLRAParticipant}) indicating that non-JAX-RS @Status method should 
-     * be invoked. The test then waits for recovery and then verifies that @Status method is 
-     * called. This method finishes compensation with return of the {@link Response} object 
-     * indicating failure and so the test then verifies that non-JAX-RS @Forget method has 
+     * {@link org.eclipse.microprofile.lra.annotation.ParticipantStatus#Compensating}
+     * (see {@link ValidLRAParticipant}) indicating that non-JAX-RS @Status method should
+     * be invoked. The test then waits for recovery and then verifies that @Status method is
+     * called. This method finishes compensation with return of the {@link Response} object
+     * indicating failure and so the test then verifies that non-JAX-RS @Forget method has
      * also been called.
      *
      * @throws InterruptedException When Test is interrupted during sleep.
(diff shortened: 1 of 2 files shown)


xstefank:
hi @zezulka, thank you for the PR. Can I help you with the ECA signing (the errored check)?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -169,7 +180,42 @@ public void testNonJaxRsCompletionStageResponseAndParticipantStatus() throws Int
...
+        ExecutorService es = Executors.newSingleThreadExecutor();
+        LRAClientOps lraOps = lraTestService.getLRAClient();
+        URI lraId = lraOps.startLRA(null, lraClientId(), 0L, ChronoUnit.MILLIS);
+        LOGGER.info(String.format("Started LRA with URI %s", lraId));
+        // Start business method asynchronously, i.e. return immediately.
+        Future<Response> lraFuture = es.submit(() -> tckSuiteTarget.path(LongBusinessMethodParticipant.ROOT_PATH)
+                .path(LongBusinessMethodParticipant.BUSINESS_METHOD)
+                .request()
+                .header(LRA.LRA_HTTP_CONTEXT_HEADER, lraId)
+                .put(Entity.text("")));
```
xstefank, line 198: you could use `async()` to handle async REST invocation. - https://javaee.github.io/javaee-spec/javadocs/javax/ws/rs/client/Invocation.Builder.html#async--

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/LongBusinessMethodParticipant.java
```diff
@@ -0,0 +1,89 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2019 Contributors to the Eclipse Foundation
```
xstefank, line 3:
```suggestion
 * Copyright (c) 2020 Contributors to the Eclipse Foundation
```

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -19,6 +19,25 @@
  *******************************************************************************/
 package org.eclipse.microprofile.lra.tck;
 
+import static org.junit.Assert.assertEquals;
```
xstefank, line 22: Copyright (c) 2019-2020

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/LongBusinessMethodParticipant.java
```diff
@@ -0,0 +1,89 @@
...
+            return Response.ok(lraId).build();
+        } catch (InterruptedException ex) {
+            return Response.serverError().build();
+        }
+    }
+
+    @PUT
+    @Path(SYNC_METHOD)
+    public Response sync() {
+        LOGGER.info("call of waitForArleadyIn");
```
xstefank, line 81: I don't understand this.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
mmusgrov: What would the benefit be (this is a TCK test)

mmusgrov: +1

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -169,7 +180,42 @@ public void testNonJaxRsCompletionStageResponseAndParticipantStatus() throws Int
...
-        
+
         lraTestService.waitForRecovery(lraId);
     }
+
+    @Test
+    public void cancelLraDuringBusinessMethod() throws InterruptedException, ExecutionException {
+        ExecutorService es = Executors.newSingleThreadExecutor();
+        LRAClientOps lraOps = lraTestService.getLRAClient();
+        URI lraId = lraOps.startLRA(null, lraClientId(), 0L, ChronoUnit.MILLIS);
```
mmusgrov, line 191: Would it be possible to also have a test where the spec implementation starts the LRA:
``` java
@LRA(value = LRA.Type.REQUIRES_NEW)
```

So it would be a similar test but without the explicit cancel: `lraOps.cancelLRA(lraId);`.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
zezulka: That's a good point, I've incorporated `acync()` into the test.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/LongBusinessMethodParticipant.java
zezulka: I suppose you didn't understand the log message? That was a typo, I've fixed it, thanks!

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -169,7 +178,42 @@ public void testNonJaxRsCompletionStageResponseAndParticipantStatus() throws Int
...
+        // -1 indicates that the LRAMetricType.Compensated key is not yet present in the metrics Map.
+        // This in turn means that @Compensate could not have been called yet.
+        Assert.assertEquals(-1, lraMetricService.getMetric(
+                LRAMetricType.Compensated, lraId, LongBusinessMethodParticipant.class.getName()));
+        LOGGER.info(String.format("Cancelled LRA with URI %s", lraId));
+        lraOps.cancelLRA(lraId);
+        // waiting for the LRA to be finished
+        lraTestService.waitForRecovery(lraId);
+        // participant has to be compensated
+        Assert.assertEquals(1, lraMetricService.getMetric(LRAMetricType.Compensated, lraId, LongBusinessMethodParticipant.class.getName()));
```
ochaloup, line 213: there can't be introduced the assertion _equal to 1_ but  _equal or bigger than 1_. The compensate method call may be called multiple times.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
zezulka: I'm not quite sure whether this test will be still similar to `cancelLraDuringBusinessMethod`. Since we don'ŧ externally cancel the LRA, do we still expect `@Compensate` to be executed? Were you thinking about incorporating timeout in some way to the additional test?

xstefank: I believe that is exactly what @mmusgrov is talking about. However, I don't think an additional test is necessary as it would be just additionally checking that timelimit is fired and we have a test for that elsewhere.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
xstefank: +1

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -169,7 +178,42 @@ public void testNonJaxRsCompletionStageResponseAndParticipantStatus() throws Int
...
+
+        Response response = lraFuture.get();
```
xstefank, line 215: It would be better to use timeout here because this will block forever, so your next assert may never be executed.

ochaloup: +1

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -169,7 +180,42 @@ public void testNonJaxRsCompletionStageResponseAndParticipantStatus() throws Int
...
+        // -1 indicates that the LRAMetricType.Compensated key is not yet present in the metrics Map.
+        // This in turn means that @Compensate could not have been called yet.
+        Assert.assertEquals(-1, lraMetricService.getMetric(
+                LRAMetricType.Compensated, lraId, LongBusinessMethodParticipant.class.getName()));
+        LOGGER.info(String.format("Cancelled LRA with URI %s", lraId));
+        lraOps.cancelLRA(lraId);
+        // waiting for the LRA to be finished
+        lraTestService.waitForRecovery(lraId);
+        // participant has to be compensated
+        Assert.assertTrue("@Compensate method should have been called at least once.",lraMetricService.getMetric(LRAMetricType.Compensated, lraId, LongBusinessMethodParticipant.class.getName()) >= 1);
```
ochaloup, line 215: Should not be this formatted to two lines rather?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -1,6 +1,6 @@
 /*
  *******************************************************************************
- * Copyright (c) 2019 Contributors to the Eclipse Foundation
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
```
ochaloup, line 3: maybe the year could not be changed?

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
zezulka: Done, thanks

PR #311, opened 2020-06-08
[294] refactor TCK tests to define in-container tests and run-as-client ones

### LINKED ISSUE #294: Refactoring the TCK test classes to separate their purposes in a cleaner way
There are currently two TCK test classes `TCKTests` and `TCKRecoverytTests`. Especially with fixes which came with `@ArquillianResource URL url` injection and the later one (https://github.com/eclipse/microprofile-lra/pull/291, https://github.com/eclipse/microprofile-lra/pull/292) it can be seen that they are not sharing the same processing.

The `TCKRecoveryTests` is meant to be run as `@RunAsClient` mode where the URL injection does not work and need to be passed by remote url call as http header attribute. While the `TCKTests` is in-container test running with full fledged injection capabilities.

From that it would be good to split those tests to not share the same parent class as they are managed in a different way and the `TCKRecoveryTest` should be explicitly marked as running outside of the container.

ochaloup:
fixes #294 

One poinit is to emphasize the functionality of the `TckRecoveryTests`.
They were run out of container by Arquillian already. But it was not clear from the comment.
Changing the text and adding the annotation `@RunAsClient` in particular defines it clearer.

The other main point is to to refactor the TckRecoveryTests to split the resource and the test functionality.
The test functionality can be run out of the container and being part of the @Test. It's easier to read and understand the test purpose.

As the last one there are few minor refactoring fixes.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
@@ -126,6 +126,8 @@ public class TckParticipantTests extends TckTestBase {
     /**
      * Test verifies {@link java.util.concurrent.CompletionStage} parametrized with 
      * {@link Void} as valid non-JAX-RS participant method return type
+     *
+     * @throws InterruptedException when waiting for the finishing the completion is interrupted
      */
     @Test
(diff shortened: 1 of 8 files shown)


ochaloup:
@xstefank thank you for your feedback. I tried to address your points. Take a look please.

PR #313, opened 2020-06-16
[lra#312] adjust sleep time by timeout factor in TckRecoveryTests

### LINKED ISSUE #312: Timeout adjustment should be available at all places when timeout is used
This is a follow-up issue to my previous one where I was asking for TCK changing timeout based of the property `lra.tck.timeout.factor` not only in-container test methods but in `TckTests#timeLimit` bean which is `LraResource` one.

But there are few more places where the timeout factor is still not used. One is the `LraResource` bean and the other is recovery tests. The recovery tests is a little bit more complicated as they are run outside of the container.

ochaloup:
fixes #312 

I propose a change on timeout handling in the TCK as I assume the timeout factor should be considered at all places where `Thread.sleep` is used.

@xstefank @mmusgrov @rdebusscher : if you may please take a look if such a change is fine from your perspective

### DIFF
--- a/tck/running_the_tck.asciidoc
+++ b/tck/running_the_tck.asciidoc
@@ -37,6 +37,7 @@ The LRA TCK suite can be parametrized by following properties, handled in the su
   When set-up lower then the timeouts will be shorter.
   Thus on slower machines it's expected longer timeouts will be needed. For example if test expects
   some waiting time to be 10 seconds and this factor is set to `1.5` then the result waiting time is 15 seconds.
+  The timeout factor must be configured as system property because of the test limitations.
 `lra.http.recovery.host`, `lra.http.recovery.port`, `lra.http.recovery.path`::
   Hostname, port and path for the recovery endpoint that will be contacted in tests checking recovery capabilities.
 
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
@@ -35,6 +35,18 @@ public class LraTckConfigBean {
     private static final Long LRA_TIMEOUT_MILLIS = 50000L;
 
     /**
+     * Name of the config property that is used as factor adjusting timeout values
+     * used in the testsuite.
+     */
+    public static final String LRA_TCK_TIMEOUT_FACTOR_PROPETY_NAME = "lra.tck.timeout.factor";
+
+    /**
+     * Name of the config property which is used to configure the TCK base url.
+     * See {@link LraTckConfigBean#tckSuiteBaseUrl}.
+     */
+    public static final String LRA_TCK_BASE_URL_PROPERTY_NAME = "lra.tck.base.url";
+
+    /**
      * <p>
      * Timeout factor which adjusts waiting time and timeouts for the TCK suite.
      * <p>
@@ -45,7 +57,7 @@ public class LraTckConfigBean {
      * If you wish the test waits shorter time than designed
      * or the timeout is elapsed faster then set the value less than <code>1.0</code>
      */
-    @Inject @ConfigProperty(name = "lra.tck.timeout.factor", defaultValue = "1.0")
+    @Inject @ConfigProperty(name = LRA_TCK_TIMEOUT_FACTOR_PROPETY_NAME, defaultValue = "1.0")
     private double timeoutFactor;
 
     /**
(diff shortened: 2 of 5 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
```diff
@@ -34,6 +34,45 @@
      */
     private static final Long LRA_TIMEOUT_MILLIS = 50000L;
 
+    /**
+     * Name of the system property that is used as factor adjusting timeout values
+     * used in the testsuite.
+     */
+    private static final String LRA_TCK_TIMEOUT_FACTOR_PROPETY_NAME = "lra.tck.timeout.factor";
```
xstefank, line 41: I wonder whether we should prefix every (even tck) property with `mp.lra.` to be aligned with general MP config names. I will do that in different PR.

tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
```diff
@@ -34,6 +34,45 @@
...
+
+    /**
+     * Name of the system property which is used to configure the TCK base url.
```
xstefank, line 44: This is actually a config property - https://github.com/eclipse/microprofile-lra/pull/313/files#diff-d7c8d58d8c6caa6639e33500b0985073R96. And we should make all TCK properties a config properties (previous is really a system property only).

tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
```diff
@@ -34,6 +34,45 @@
...
+
+    /**
+     * A helper method which is capable to adjust timeout value
+     * by factor obtained from system property {@code #LRA_TCK_TIMEOUT_FACTOR_PROPETY_NAME}.
+     *
+     * @param timeout  timeout value which will be adjusted
+     * @return  adjusted timeout by te default factor
+     */
+    public static long adjustTimeoutByDefaultFactor(long timeout) {
+        String timeoutFactor = System.getProperty(LraTckConfigBean.LRA_TCK_TIMEOUT_FACTOR_PROPETY_NAME, "1.0");
```
xstefank, line 72: use please the config property from line https://github.com/eclipse/microprofile-lra/pull/313/files#diff-d7c8d58d8c6caa6639e33500b0985073R87 as that can be also taken from env vars, file, etc. not only system properties.

ochaloup: @xstefank that's the trouble :-) I absolutely understand what you talk about and I was a bit hesitating to propose this change. The reason why I provided this method which uses the system property lookup is the `TckRecoveryTests` which run out of the container. Because of that there is no MP config available
See here: https://github.com/eclipse/microprofile-lra/pull/313/files#diff-0148f6b901fbaaf1ace873d4d13ce058R189

As I'm thinking about it once again I think it's a bit misleading and not a good idea to add this method under the config bean at the first place. The config bean should be just operating with MP Config stuff.
I can see two ways how to change this this - either completely omit the tck timeout factor from the `TckRecoveryTests` or (and I would probably prefer this option) adding the `System.getProperty` call only for the `TckRecoveryTests` class and not into the `LraTckConfigBean`. Would that be ok by you?

ochaloup: @xstefank I did the adjustment with system property used only in the `TckRecoveryTests`. Please, take a look.

tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
ochaloup: aghrr... I missed the point of this comment before. Now I got it. I changed the text for 'config property'.

ochaloup:
@xstefank please, take a look I hope I addressed both your points now.

tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
xstefank: I wonder whether it must then be marked as run as client? In every case, document please that this one must be a system property in the `running_the_tck.asciidoc` please.

ochaloup:
@xstefank I added the referred line to the documentation.

tck/running_the_tck.asciidoc
```diff
@@ -37,6 +37,7 @@ The LRA TCK suite can be parametrized by following properties, handled in the su
   When set-up lower then the timeouts will be shorter.
   Thus on slower machines it's expected longer timeouts will be needed. For example if test expects
   some waiting time to be 10 seconds and this factor is set to `1.5` then the result waiting time is 15 seconds.
+  The timeout factor should be configured as system property because of the test limitations.
```
xstefank, line 40: must be

tck/src/main/java/org/eclipse/microprofile/lra/tck/LraTckConfigBean.java
```diff
@@ -70,10 +82,19 @@ public long getDefaultTimeout() {
...
      * @return value of adjusted timeout
      */
     public long adjustTimeout(long timeout) {
+        return adjustTimeout(timeout, timeoutFactor);
+    }
+
+    private long adjustTimeout(long timeout, double timeoutFactor) {
+        if (timeout < 0 || timeoutFactor < 0)  {
+            throw new IllegalArgumentException(String.format(
+                    "Provided arguments (timeout=%d, timeoutFactor=%.2f) has to be positive", timeout, timeoutFactor));
```
xstefank, line 95: have to be

ochaloup:
@xstefank I changed the text you pointed at.

Issue #298, opened 2020-04-27
TCK: Replace LRAClientOps#isLRAFinished with afterLRA calls


xstefank:
As per Gitter discussion, replace the current way of checking of active LRA in the TCK (try enlisting) with direct SPI.

mmusgrov:
+1

xstefank:
Changed the title of issue after the discussion in the PR #320

PR #315, opened 2020-06-17
[314] refactor LRAMetrics to assert easier

### LINKED ISSUE #296: Some TCK test assertions do not report the expected versus the actual value
For example [TckTests#joinWithOneResource](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java#L537). Reporting this information provides the developer with a quicker route to tracking down the root cause.

Remark: this issue overlaps with issue #314 . However the fix for issue #314 (pr #315 ) still does not show the expected versus the actual values.

### LINKED ISSUE #314: Enhance a way how LRAMetric assertions are used
I would like to enhance the way how LRAMetric is used in tests. It's about refactoring by creating a "helper" assertion class that could be injected to tests and used for checking the results.

The idea is that it would be easier to use and understand for developers.

The aim is to

* creating assertion for the LRAMetrics to be more straightforward to check them in tests
* enforce using the participant name to not make a error when creating new test

This issue should subsume the requirements stated in issue #296

ochaloup:
fixes #314 
fixes #296 

Signed-off-by: Ondra Chaloupka <ochaloup@redhat.com>

@xstefank @mmusgrov @rdebusscher if you have a chance please take a look

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricAssertions.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricAssertions.java
@@ -1,0 +1,343 @@
+/*
+ *******************************************************************************
+ * Copyright (c) 2020 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * You may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *******************************************************************************/
+package org.eclipse.microprofile.lra.tck.service;
+
+import org.hamcrest.Matchers;
+
+import javax.enterprise.context.Dependent;
+import javax.inject.Inject;
+import java.net.URI;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Assertion methods usable with metrics.
+ */
+@Dependent
+public final class LRAMetricAssertions {
+
+    @Inject
+    private LRAMetricService lraMetricService;
+
+    // ----------------------------- COMPENSATED -----------------------------------
+    /**
+     * Asserts that <b>compensated</b> was called for given LRA and participant class translated to fully qualified classname as String,
+     * if not the {@link AssertionError} with the given message is thrown.
+     *
+     * @param message assertion message when the check fails
+     * @param lraId LRA id which the assertion check will be taken against
+     * @param participantClazz  the participant class used as resource name in the map
+     */
+    public void assertCompensated(String message, URI lraId, Class<?> participantClazz) {
+        assertYes(message, LRAMetricType.Compensated, lraId, participantClazz);
+    }
+
+    /**
+     * Asserts that <b>compensated</b> was not called for given LRA and participant class translated to fully qualified classname as String,
+     * if not the {@link AssertionError} with the given message is thrown.
+     *
+     * @param message assertion message when the check fails
+     * @param lraId LRA id which the assertion check will be taken against
+     * @param participantClazz  the participant class used as resource name in the map
+     */
+    public void assertNotCompensated(String message, URI lraId, Class<?> participantClazz) {
+        assertNot(message, LRAMetricType.Compensated, lraId, participantClazz);
+    }
+
+    /**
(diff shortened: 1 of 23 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricService.java
```diff
@@ -28,49 +30,107 @@
...
+     * TODO if the current PR is acceptable then delete the old isLRAFinished method
+     * (which tests whether an LRA is active by making an attempt to enlist with it
+     * which some spec implementations report as a stack trace WARNING if the LRA is
+     * no longer active).
+     *
+     * @param lraId the LRA id to test
+     * @param participantClazz class of the participant that the metrics parameter applies to
+     * @return whether or not an LRA has finished
+     */
+    boolean isLRAFinished(URI lraId, Class<?> participantClazz) {
```
xstefank, line 120: I am not sure about this approach. This will require that each test that wants to assert the finish of LRA then it must provide AfterLRA method which must remember to invoke the correct "end" metric. Yes, this can work, but we must remember to keep an eye on new tests (I verified that your current PR passes the TCK). Also, this prohibits the simple tests without any (even AfterLRA) enlistments which may also be required. I believe that https://github.com/eclipse/microprofile-lra/issues/298 should still be implemeted.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricService.java
```diff
@@ -79,21 +139,23 @@ public void clear() {
...
             .collect(Collectors.toMap(Function.identity(), t -> new AtomicInteger(0)));
 
-        void increment(LRAMetricType metric) {
-            if (metrics.containsKey(metric)) {
-                metrics.get(metric).incrementAndGet();
+        void increment(LRAMetricType metricType) {
+            if (metrics.containsKey(metricType)) {
+                metrics.get(metricType).incrementAndGet();
+            } else {
+                throw new IllegalStateException("Cannot increment metric type " + metricType.name());
```
xstefank, line 149:
```suggestion
                throw new IllegalArgumentException("Cannot increment metric type " + metricType.name());
```

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricService.java
ochaloup: ad #298) yes, I think this should still be implemented. Just I haven't touched that as I didn't know if you are working on it. I can take care of it (but I don't remember details troubles here)
ad approach) I think it's fine just force this method to use the `AfterLRAParticipant` as the "`participantClazz`" that will be checked here all the time. My refactoring didn't go anyhow far from what was used until now as the client call (before my refactoring) did
```
assertTrue("testAfterLRAParticipant: LRA did not finish",
lraClient.isLRAFinished(lra, lraMetricService, AfterLRAParticipant.class.getName()));
```
Should I freeze this assertion method to the `AfterLRAParticipant`, correct?

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricAssertions.java
```diff
@@ -0,0 +1,349 @@
...
+
+    @Inject
+    private LRAMetricService lraMetricService;
+
+    /**
+     * Returning the underlying {@link LRAMetricService} used within the assertions methods.
+     *
+     * @return lra metric service for further use
+     */
+    public LRAMetricService getLraMetricService() {
```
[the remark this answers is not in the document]
ochaloup: +1, agree, thanks

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricService.java
xstefank: I will create PR for #298 just facing some issues in our implementation. And no, I don't think that any change to the current code is needed. You don't need to hardcode the AfterLRAParticipant for now.

mmusgrov:
@ochaloup The way you have implemented this enhancement invalidates issue #296

ochaloup:
@mmusgrov hi. please, can you elaborate what this meant by invalidating the issue #296? Should be changed something in this PR?
As far as I understand the #296 assertion should be more informative, have I got it right? Is the approach of this PR good enough?

mmusgrov:
> @mmusgrov hi. please, can you elaborate what this meant by invalidating the issue #296? Should be changed something in this PR?
> As far as I understand the #296 assertion should be more informative, have I got it right? Is the approach of this PR good enough?

Issue #296 was created because there are some test assertions that do not report *both* the expected and the actual values. Reporting both values is fairly standard in tests, they enable the reader to see at a glance what the expected outcome should have been.

Your PR is still not reporting both the expected and the actual values (in fact you have changed some existing assertions so that they no longer report both values). In other words your PR is at odds with (contradicts) #296

ochaloup:
@mmusgrov ok, I see the point now, thanks. I changed the PR to reflect your remark.

ochaloup:
@mmusgrov I see, agree. I updated this PR to address your remark.

xstefank:

xstefank:

Issue #306, opened 2020-06-05
Enhance the text that describes nested LRA behaviour


mmusgrov:
The [NESTED LRA annotation element](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L351) has extra requirements beyond the ones for ['top level' LRAs](https://github.com/eclipse/microprofile-lra/blob/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/ws/rs/LRA.java#L109). To facilitate safe usage of the feature we need to provide clear documentation. This task should at least include the following changes:

1. Update the javadoc for the @AfterLRA annotation to describe the meaning of the LRA_HTTP_PARENT_CONTEXT_HEADER header
2. Include a comprehensive example showing best practice when using nested LRAs.
3. Ensure that the spec examples include nesting.
4. Review the existing text around nesting to verify its accuracy, clarity and sufficiency.

xstefank:
PR merged.

PR #316, opened 2020-07-05
Issue306 Enhance description of nested contexts

mmusgrov:
#306 

This PR follows up on the discussion we had in the gitter room about the clarification of nested contexts and ensuring the parent context header is present in the examples.

I split it across two commits (since each commit arose from a separate discussion).

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -279,6 +279,8 @@ package:
   | Indicates that this class is no longer interested in this LRA.
 | https://github.com/eclipse/microprofile-lra/tree/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java#L30[@Status]
   | When the annotated method is invoked it should report the status.
+| https://github.com/eclipse/microprofile-lra/tree/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java#L30[@AfterLRA]
+  | When an LRA has reached a final state the annotated method is invoked will be invoked.
 |===
 
 Briefly, these annotations are used as follows:
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -280,7 +280,7 @@ package:
 | https://github.com/eclipse/microprofile-lra/tree/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java#L30[@Status]
   | When the annotated method is invoked it should report the status.
 | https://github.com/eclipse/microprofile-lra/tree/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java#L30[@AfterLRA]
-  | When an LRA has reached a final state the annotated method is invoked will be invoked.
+  | When an LRA has reached a final state the annotated method is invoked.
 |===
 
 Briefly, these annotations are used as follows:
@@ -337,14 +337,14 @@ security is specifically not defined in the specification and it is up to implem
 [[lra-context]]
 ==== The LRA Context
 
-When a method is invoked in the context of an LRA its identifier (of type java.net.URI) MUST be made
+When a method (that is annotated with any of the annotations defined in this specification)
+is invoked in the context of an LRA its identifier (of type java.net.URI) MUST be made
 available to the business logic. This LRA is referred to as the `active context`.
 
-For JAX-RS resource methods, the identifier is made available via request and
-response headers which the business method can obtain using standard JAX-RS mechanisms,
-i.e. `@Context` or by injecting a JAX-RS header param with the name specified by
-the Java constant `LRA_HTTP_CONTEXT_HEADER` as defined <<source-LRA, in the LRA annotation class>>.
-This header is referred to as the `context header`.
+For JAX-RS resource methods the LRA identifier is made available to the business method
+(via standard JAX-RS mechanisms, i.e. a JAX-RS `@Context` annotation or by injecting a JAX-RS
+header param with the name specified by the Java constant `LRA_HTTP_CONTEXT_HEADER` as defined
+<<source-LRA, in the LRA annotation class>>. This header is referred to as the `context header`.
 
 When using non-JAX-RS based 'complete' and 'compensate' methods (see <<non-jaxrs-participant-methods>>)
 this identifier is passed as a method parameter.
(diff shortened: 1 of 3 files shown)


spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -335,14 +337,14 @@ security is specifically not defined in the specification and it is up to implem
 [[lra-context]]
 ==== The LRA Context
 
-When a method is invoked in the context of an LRA its identifier (of type java.net.URI) MUST be made
+When a method (that is annotated with any of the annotations defined in this specification)
```
xstefank, line 340: I would suggest rephrasing this to "a method that is invoked by the implementation of this specification" because we also include annotations (LRA, Leave) that are executed by users and thus we can't always guarantee that the LRA id is made available.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -637,20 +662,63 @@ meaning of the term `eventual consistency` as used in this specification).
...
+            // A nested context is being cancelled. Nested contexts are explained in detail in a later section.
+
+            // Code which is sensitive to executing with a nested context goes here. Generally
+            // only the @Complete handler is sensitive to executing with a nested context.
+        }
+
+        // perform business logic which compensates for any work performed in the context of the LRA
+        getActivity(lraId).setCompensated();
+
+        return Response.ok();
```
xstefank, line 687:
```suggestion
        return Response.ok().build();
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -637,20 +662,63 @@ meaning of the term `eventual consistency` as used in this specification).
...
+
+            // If a nested context is closed but the parent context is later cancelled
+            // then any work performed in the nested context must still be capable of
+            // being compensated for. Therefore, even though no more work will be executed
+            // with the nested context, the business logic must remember how to compensate
+            // for any work it did do (until explicitly told to forget about the nested context).
+        }
+
+        // business logic for handling the LRA (provided by the business developer)
+        getActivity(lraId).setCompensated();
```
xstefank, line 706: this is probably copy-paste error, this is complete method.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -637,20 +662,63 @@ meaning of the term `eventual consistency` as used in this specification).
...
 
         return Response.ok();
```
xstefank, line 708:
```suggestion
        return Response.ok().build();
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -637,20 +662,63 @@ meaning of the term `eventual consistency` as used in this specification).
...
     }
+
+    @Forget
+    @Path("/forget")
+    @DELETE
+    public Response forget(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId,) {
+                           @HeaderParam(LRA_HTTP_PARENT_CONTEXT_HEADER) URI parentLRA) {
+        if (parentLRA != null) {
+            // code which is sensitive to executing with a nested context goes here
+            ...
```
xstefank, line 718: also please mention that you should clean up the information required for possible nested compensation here.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -688,8 +761,13 @@ reporting its` progress using the `@Status` annotation:
     @Status
     @Path("/status")
     @GET
-    public Response status(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId) {
-        if (isFinished(lraId))  // Business logic, not provided by spec or Implementation
+    public Response status(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId,
+                           @HeaderParam(LRA_HTTP_PARENT_CONTEXT_HEADER) URI parentLRA) {
+        if (parentLRA != null) {
+            // code which is sensitive to executing with a nested context goes here
```
xstefank, line 767: just suggestion, it seems to me that this if statements may bring an impression that the user should do something with the perentLRA. Maybe we can change the comment to"
```suggestion
            // lraId is nested, the code which is sensitive to executing with a nested context goes here
```
or something similar.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -764,14 +863,19 @@ if the LRA is completed or not: the room will be unavailable for others. If it r
...
 `@Complete`.
 
-If the participant successfully compensates or completes then it may forget about
-the LRA. Otherwise it should remember that it is still associated with the LRA and
-it MUST report the status of the association using the
-<<jaxrs-response-table,appropriate JAX-RS response code>> if it uses JAX-RS for
-participant notifications. If it does not use JAX-RS for participant notifications
-then it MUST report the status of the association using one of the values in the
+In the case of a top level LRA, if the participant successfully compensates or
+completes then it may forget about the LRA. In the case of a nested LRA the
```
xstefank, line 867:
```suggestion
completes then it may forget about the LRA. In the case of a nested LRA, the
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -1304,7 +1444,7 @@ the JAX-RS `@Suspended` annotation:
 
 [source,java]
 ----
-    @LRA(value = LRA.Type.REQUIRED, // the method must run with an LRA
+    @LRA(value = LRA.Type.REQUIRES_NEW, // the method must run with an LRA
```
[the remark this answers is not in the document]
mmusgrov: REQUIRES_NEW means there is no reason to inject 

```java
@HeaderParam(LRA_HTTP_CONTEXT_HEADER)
```

We use REQUIRED in a lot more examples than REQUIRES_NEW so I removed it in order to:

- balance it up (to show that handling nested contexts isn't always required);
- avoid too much repetition of text around handling nested contexts (since the reader will get the idea after the first few examples).

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: We still need to make the context available even when clients invoke the methods. For example it is how we implement LRA element types such as REQUIRED, REQUIRES_NEW, etc. An example implementation technique could be to apply a JAX-RS filter to outgoing requests, this filter would add the context header making it available at the target service. The target service would have another incoming JAX-RS filter which would decide what to do with the context before the business method runs.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: I cleaned up all of this example so if you don't mind can you recheck the commentary I added to the code snippets.

spec/src/main/asciidoc/microprofile-lra-spec.adoc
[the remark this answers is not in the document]
mmusgrov: @ochaloup I have just pushed another commit (and squashed) that explains why there are no checks for a nested context when the method is annotated with REQUIRES_NEW (line 1499 below).

spec/src/main/asciidoc/microprofile-lra-spec.adoc
```diff
@@ -279,6 +279,8 @@ package:
   | Indicates that this class is no longer interested in this LRA.
 | https://github.com/eclipse/microprofile-lra/tree/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/Status.java#L30[@Status]
   | When the annotated method is invoked it should report the status.
+| https://github.com/eclipse/microprofile-lra/tree/master/api/src/main/java/org/eclipse/microprofile/lra/annotation/AfterLRA.java#L30[@AfterLRA]
+  | When an LRA has reached a final state the annotated method is invoked will be invoked.
```
xstefank, line 283:
```suggestion
  | When an LRA has reached a final state the annotated method is invoked.
```

spec/src/main/asciidoc/microprofile-lra-spec.adoc
mmusgrov: I put something similar. In fact I reworked the whole example to make it more comprehensive.

xstefank:

xstefank:

Issue #300, opened 2020-04-28
Remove references to userData in the examples section


mmusgrov:
The [examples section](https://github.com/eclipse/microprofile-lra/blob/master/spec/src/main/asciidoc/microprofile-lra-spec.adoc#324-starting-and-ending-lras) of the spec document still refers to userData being passed to the complete and compensate callbacks, for example:

```java
public Response completeWork(@HeaderParam(LRA_HTTP_CONTEXT_HEADER) URI lraId,
                                 String userData)
```

We removed that option a while ago.

PR #317, opened 2020-07-05
issue300 remove references to userData

mmusgrov:
#300 

There were still a couple of old references to userData in the spec but we dropped that a while ago.


xstefank:
@ochaloup this PR already removes the reference you mention. Can you take a look again, please?

PR #319, opened 2020-07-11
issue318 Fix LRA context propagations in TckTests#nestedActivity()

### LINKED ISSUE #318: TckTests#nestedActivity() puts wrong expectations on the LRA_HTTP_CONTEXT_HEADER propagation
The [TckTest#nestedActivity()](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java#L126) calls the [LraResource#nestedActivity(URI, URI)](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/api/LraResource.java#L349) which returns `Response.ok(nestedLRAId).build()` where the `nestedLRAId` is injected `LRA_HTTP_CONTEXT_HEADER`.

The test tries to also receive the content of the **parent** LRA id at line [141](https://github.com/eclipse/microprofile-lra/blob/master/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java#L141) with the following call:

```
Object parentId = response.getHeaders().getFirst(LRA_HTTP_CONTEXT_HEADER);
```

However, the **response** header called **LRA_HTTP_CONTEXT_HEADER** is not included in the response returned from the LraResouce endpoint.

As far as I am aware and by scanning the spec we are not specifying that the response headers should be implicitly filled for user calls and even if so this propagation doesn't make sense since the user is getting `LRA_HTTP_CONTEXT_HEADER` with the content of **parentId** so at least this should be `LRA_HTTP_PARENT_CONTEXT_HEADER`. I suggest to include the parent header in the response directly as we do in the other tests that need to read the LRA ids.

xstefank:
resolves #318

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
@@ -20,6 +20,7 @@
 package org.eclipse.microprofile.lra.tck;
 
 import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_CONTEXT_HEADER;
+import static org.eclipse.microprofile.lra.annotation.ws.rs.LRA.LRA_HTTP_PARENT_CONTEXT_HEADER;
 import static org.eclipse.microprofile.lra.tck.participant.api.AfterLRAListener.AFTER_LRA_LISTENER_WORK;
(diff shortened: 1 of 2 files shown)


xstefank:
@ochaloup also review, please.

PR #321, opened 2020-07-30
#299 Fix participant HTTP responses to 200, 202, 409, 410

### LINKED ISSUE #299: Fix participant responses HTTP statuses to 200 (Compensated/Completed), 202 (Compensating/Completing), and 409 (FailedToCompensate/FailedToComplete)
... and distinguish the LRA participant outcome based on the HTTP message body:

* 200 + ParticipantStatus.Compensated / ParticipantStatus.Completed
* 200 + ParticipantStatus.Compensating / ParticipantStatus.Completing
* 200 + ParticipantStatus.FailedToCompensate / ParticipantStatus.FailedToComplete

xstefank:
resolves #299
﻿

### DIFF
--- a/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
+++ b/api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
@@ -122,12 +122,7 @@ import java.lang.annotation.Target;
  *   <td scope="row">The resource is still attempting compensation</td>
  * </tr>
  * <tr>
- *   <td scope="row">410</td>
- *   <td scope="row">Empty</td>
- *   <td scope="row">The resource does not know about the LRA</td>
- * </tr>
- * <tr>
- *   <td scope="row">500</td>
+ *   <td scope="row">409</td>
  *   <td scope="row">{@link ParticipantStatus} enum value</td>
  *   <td scope="row"><p>The resource has failed to compensate.
  *   The payload contains the reason for the failure.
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/eclipse/microprofile/lra/annotation/Compensate.java
```diff
@@ -127,7 +127,7 @@
  *   <td scope="row">The resource does not know about the LRA</td>
  * </tr>
  * <tr>
- *   <td scope="row">500</td>
```
ochaloup, line 130: I have two points that are not major ones.
* It could be nicer if the table content would be ordered by error code. That mean the line of table with cod `409` belongs before the line with code `410`.
* As the text below says it's possible that the implementation returns the `500`. Is it obvious by the HTTP behaviour or should that code be part of the table?

xstefank: > As the text below says it's possible that the implementation returns the 500. Is it obvious by the HTTP behaviour or should that code be part of the table?

HTTP behavior, this is not a supported code returned by the implementation of LRA but the general error which can be returned from JAX-RS server (same as connection error for instance) so it shouldn't be mentioned in the table.

xstefank:
@ochaloup can you take a look again, please?

xstefank:

xstefank:

xstefank:
only rebase

Issue #327, opened 2020-08-28
TCK won't pass on apache-cxf


idlewis:
There appears to be a difference in behaviour between the JAX-RS client from the Apache-CXF project.
When setting a request header e.g.
```
builder.header(LRA_TCK_HTTP_CONTEXT_HEADER, lraContext);
```
where builder is an InvocationBuilder, if the header value is a Java null, then Apache-CXF will set a header with a String literal value of 'null'. 
Whilst I can't prove this on my setup, I believe that using resteasy, no header is set when the header value is null.

The result of this is that the TCK fails if you try to run on top of Apache-CXF, as TckContextTests sets a null value for a number of tests. When this request gets to the test resource, the Narayana LRA implementation falls over as it tries to interpret an LRA id of 'null'

I think the Narayana implementation ought to handle the situation better, but I think the TCK should probably not be setting null values either.

idlewis:
I've open pull request #328 with a possible fix for this

xstefank:
PR merged.

PR #328, opened 2020-08-28
Don't set the lra header if it is null

idlewis:
If the JAX-RS client is apache-cxf, this results in header being set
with a value of the string 'null'. When the header is sent as part of
the test, it breaks the Narayana implementation


idlewis:
This is aiming to fix #327

mmusgrov (approved):

xstefank (approved):

rdebusscher (approved):

PR #329, opened 2020-09-04
Fix two test bugs

idlewis:
In TckLRATypeTests, the resourceRequest method is catching 'Throwable'.
This means it will catch any assertions thrown inside the method, which
will hide test failures.

in LRAClientOps, the cleanUp method is vulnerable to ConcurrentModificationException
as the cancelCancelation method may cause entries to be removed from the lraTasks
map, while it is being iterated over.


idlewis:
Oops, just seen that there all ready issues open to fix these two problems

PR #331, opened 2020-09-21
Remove pushChanges configuration from the release plugin

xstefank:


xstefank:
This is only a possible release build fix. No review is needed.

ochaloup (approved):

PR #337, opened 2020-10-30
[#336] make LRA service to wait for nested LRAs in the TCK nestedActivity tests

### LINKED ISSUE #336: TCK nestedActivity test uses the LRATestService to wait for top level LRA which was not finished yet
The `*nestedActivity` tests (tests which use method `multiLevelNestedActivity`) use the `LRATestService` to wait for finishing the LRAs (getting the completion callback being called). But the service is instrumented to wait for the top-level LRA which is not finished at that time. Only nested LRAs should be verified as being finished.

See the call which waits for top-level LRA at https://github.com/eclipse/microprofile-lra/blob/1.0-M1/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java#L620

ochaloup:
fixes #336

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
@@ -604,7 +604,10 @@ public class TckTests extends TckTestBase {
         assertFalse("multiLevelNestedActivity: top level LRA should be active (path called " + resourcePath.getUri() + ")",
                 lraTestService.isLRAFinished(URI.create(lraArray[0])));
 
-        lraTestService.waitForCallbacks(lra);
+        // starting at index 1 as LRAResource#multiLevelNestedActivity returns the top-level LRA as the first argument which was not finished yet
+        IntStream.range(1, uris.length).parallel().forEach(i -> {
+            lraTestService.waitForCallbacks(uris[i]);
+        });
 
         // check that all nested activities were told to complete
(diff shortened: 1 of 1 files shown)


tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
```diff
@@ -604,7 +604,10 @@ private void multiLevelNestedActivity(CompletionType how, int nestedCnt) throws
         assertFalse("multiLevelNestedActivity: top level LRA should be active (path called " + resourcePath.getUri() + ")",
                 lraTestService.isLRAFinished(URI.create(lraArray[0])));
 
-        lraTestService.waitForCallbacks(lra);
+        // starting at index 1 as LRAResource#multiLevelNestedActivity returns the top-level LRA as the first argument which was not finished yet
+        IntStream.range(1, uris.length).forEach(i -> {
```
xstefank, line 608: looks good, but maybe we can `.parallel()` this?

xstefank (approved):

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckTests.java
ochaloup: Ok, I added the `.parallel()` here.

mmusgrov:
@ochaloup I think we can merge test issues without needing to reach out to the broader community. If you agree then this one can be merged.

ochaloup:
@mmusgrov yes, I agree.

xstefank:
agreed, merging. Thanks.

PR #340, opened 2020-12-01
[#339] TckRecoveryTests Arquillian URL injection adjustment

### LINKED ISSUE #339: Use of Arquillian URL injection cannot resolved in TckRecoveryTests
I'm hitting a trouble with Arquillian injection of `URL` in `TckRecoveryTests`.
Currently the tests injects with `@ArquillianResource URL deploymentURL`, see here: https://github.com/eclipse/microprofile-lra/blob/1.0-M1/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckRecoveryTests.java#L68

This is problematic for Arquillian resolution. At least the WildFly/Thorntail arquillian managed container adapter has troubles to resolve it (and I think it has to make troubles for any other adapter).

Tthe container is started by Arquillian as this is an managed container. But the deployment happens only at the test method execution (https://github.com/eclipse/microprofile-lra/blob/1.0-M1/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckRecoveryTests.java#L113).
In the same time the `@ArquillianResource` injection happens at instance initialization which is *before* any test method is started. Ie. at time of arquillian resource injection there is no clear idea about deployment (for some adapters this is important - eg. for WildFly where the URL path depends on the name of the deployed `war` file, for Thorntail the `deploy` method means to start the container which makes injection troublesome as well) and the `URL` is just `null`.

I would like to change this by moving the `URL` injection at time of test execution while the `deploy` moving to `@Before` to happens before the test starts.

ochaloup:
fixes #339 

Adjusting the URL injection by Arquillian to be done after the deployment happens. This fixes a trouble for some adapters to inject `URL` as `null`.

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckRecoveryTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckRecoveryTests.java
@@ -67,17 +67,12 @@ import static org.junit.Assert.assertThat;
 public class TckRecoveryTests {
     private static final Logger LOG = Logger.getLogger(TckRecoveryTests.class.getName());
     
-    public static final String LRA_TCK_DEPLOYMENT_URL = "LRA-TCK-Deployment-URL";
-
     private static final String DEPLOYMENT_NAME = "tck-recovery";
     private static final Logger LOGGER = Logger.getLogger(TckRecoveryTests.class.getName());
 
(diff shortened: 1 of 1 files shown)


xstefank (approved):

ochaloup:
@xstefank @mmusgrov some possibility to merge this?

PR #344, opened 2021-01-29
Fix build warnings

xstefank:

Please note that formating changes are due to the `.editorconfig` file that now setups IDEs to remove trailing spaces. This will eventually be done in all saved files so we can do it in one PR if requested but not necessary.


rdebusscher (approved):
Looks Good!

ochaloup (approved):

PR #345, opened 2021-02-04
Remove status indication for the README file

xstefank:


xstefank:
Non-functional change. It's just something that is no longer true for the 1.0 tag :).

xstefank:
This is really just a README clean up so I will merge it as it is blocking the release.

PR #350, opened 2021-04-12
Set nexus plugin stagingProgressTimeoutMinutes to 20 minutes

xstefank:


xstefank:
This is only CI related fix.

PR #353, opened 2021-09-06
Fix a few mistakes in the specification

### LINKED ISSUE #352: A few mistakes in the spec
In LRA, I spotted the following minor mistakes:

- The following code snippet contains the wrong line:

```
    <artifactId>microprofile-lra-api</artifactId>
    <version>${version.microprofile.lra}</version>
    <version>1.0</version>
```

should be 

```
     <artifactId>microprofile-lra-api</artifactId>
     <version>1.0</version>
```

- The following line is wrong

```
@LRA(LRA.Type.MANDATORY, // requires an active context before method can be executed
       end = true) // end the received LRA when the method finishes
```

should be
```
@LRA(value=LRA.Type.MANDATORY, // requires an active context before method can be executed
       end = true) // end the received LRA when the method finishes
```

- The following line is incorrect

`@LRA(LRA.Type.NESTED, end = true) `should be
`@LRA(value = LRA.Type.NESTED, end = true)`

xstefank:
Fixes #352 

I think this should be fixed before we fork jakarta branch.

### DIFF
--- a/spec/src/main/asciidoc/microprofile-lra-spec.adoc
+++ b/spec/src/main/asciidoc/microprofile-lra-spec.adoc
@@ -565,7 +565,7 @@ the `cancelOn` and `cancelOnFamily` elements:
       ...
   }
 
-  @LRA(LRA.Type.MANDATORY, // requires an active context before method can be executed
+  @LRA(value = LRA.Type.MANDATORY, // requires an active context before method can be executed
        end = true) // end the received LRA when the method finishes
   @Path("/confirm")
   @PUT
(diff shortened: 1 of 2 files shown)


mmusgrov (approved):
Great

mmusgrov:
@xstefank I missed some more examples of this:

`./src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/InvalidAfterLRASignatureListener.java ./src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/InvalidArgumentTypesParticipant.java ./src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/InvalidReturnTypeParticipant.java ./src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/TooManyArgsParticipant.java`

xstefank:
@mmusgrov I think I got everything. If there is only one argument the `value =` can be omitted. Java files wouldn't compile and the one remaining spec reference seems ok to me (line 486).

mmusgrov:
@xstefank you are correct. Thanks for providing a patch.
Since the change is straight forward I would argue that we can merge without a review from @rdebusscher

PR #354, opened 2021-09-08
Clarify use of enum LRAStatus as entity parameter

### LINKED ISSUE #347: Use of enum LRAStatus as entity parameter
The `LRAStatus` enum is used as a JAX-RS entity parameter in the spec and the TCKs. Although some JAX-RS implementations support this usage, it is _not mandated_ by the JAX-RS specification. More precisely, implementations are not required to provide an out-of-the-box message body reader to handle enums in this manner (they are required to handle enums this way associated with `*Param` injections).

I suppose it is possible to mandate LRA implementations to extend a JAX-RS implementation to support this, but this does not seem to be mentioned anywhere so I suspect the assumption made here is that all JAX-RS implementations support this, which is not the case. 

A short clarification on what is expected would be useful.

xstefank:
Fixes #347


ochaloup (approved):

mmusgrov (approved):

mmusgrov:
@xstefank again, since this change is not contentious, I think we can merge with just the one approval.

xstefank:
@mmusgrov sure, I was just waiting for more reviews.

PR #368, opened 2025-02-13
Resolve flaky tests

marcosgopen:
solves https://github.com/microprofile/microprofile-lra/issues/369

### DIFF
--- a/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
+++ b/tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
@@ -142,6 +142,8 @@ public class TckParticipantTests extends TckTestBase {
         URI lraId = URI
                 .create(checkStatusReadAndCloseResponse(Response.Status.INTERNAL_SERVER_ERROR, response, resourcePath));
 
+        lraTestService.waitForCallbacks(lraId);
+
         lraMetric.assertCompensatedEquals(
                 "Non JAX-RS @Compensate method with CompletionStage<Void> should have been called",
                 1, lraId, ValidLRACSParticipant.class);
@@ -167,6 +169,8 @@ public class TckParticipantTests extends TckTestBase {
         Response response = resourcePath.request().get();
 
         URI lraId = URI.create(checkStatusReadAndCloseResponse(Response.Status.OK, response, resourcePath));
+
+        lraTestService.waitForCallbacks(lraId);
 
         lraMetric.assertCompletedEquals(
                 "Non JAX-RS @Complete method with CompletionStage<Response> should have been called",


xstefank (commented):
@marcosgopen thanks for the PR, but what kind of flaky test is this fixing? This is a valid case according to the [spec](https://github.com/microprofile/microprofile-lra/blob/main/spec/src/main/asciidoc/microprofile-lra-spec.asciidoc#non-jax-rs-participant-methods) so you might have problem in your implementation, not in the TCK test.

xstefank:
@marcosgopen is this ready for review? It seems like very in the middle of WIP.

pom.xml
```diff
@@ -20,7 +20,7 @@
     <parent>
         <groupId>org.eclipse.microprofile</groupId>
         <artifactId>microprofile-parent</artifactId>
-        <version>2.4</version>
+        <version>3.4</version>
```
xstefank, line 23: This cannot be done in this way.

First, don't do dependency upgrades in the feature PRs.
Second, this would move minimum JDK from 8 to 11 which can only be done in major upgrade of the specification.

pom.xml
```diff
@@ -40,6 +40,7 @@
 <!--        <bnd-maven-plugin.version>3.4.0</bnd-maven-plugin.version>-->
 <!--        <bnd-baseline-maven-plugin.version>4.2.0</bnd-baseline-maven-plugin.version>-->
 <!--        <build-helper-maven-plugin.version>1.12</build-helper-maven-plugin.version>-->
+        <maven.compiler.release>17</maven.compiler.release>
```
xstefank, line 43: Well, this is totally unacceptable :) MP is only JDK 11 minimum, and we should not go over that.

tck/pom.xml
```diff
@@ -63,21 +63,25 @@
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
+            <version>4.13.2</version>
```
xstefank, line 66: No, we should not override versions from the MP parent. That's what parent is for.

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
```diff
@@ -142,6 +142,9 @@ public void testNonJaxRsCompletionStageVoid() throws InterruptedException {
         URI lraId = URI
                 .create(checkStatusReadAndCloseResponse(Response.Status.INTERNAL_SERVER_ERROR, response, resourcePath));
 
+        // lraTestService.waitForCallbacks(lraId);
+        lraTestService.waitForRecovery(lraId);
```
xstefank, line 146: Seems like forgotten testing? Decide if it should be waitForCallbacks (I think it should) or waitForRecovery.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
```diff
@@ -78,29 +80,47 @@ public Response enlistWithCompensate(@HeaderParam(LRA.LRA_HTTP_CONTEXT_HEADER) U
     public CompletionStage<Void> compensate(URI lraId) {
         assert lraId != null;
 
+        Executor delayed = CompletableFuture.delayedExecutor(1L, TimeUnit.SECONDS);
```
xstefank, line 83: Seems like forgotten testing? Or what is the point in the delaying of the test?

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
```diff
@@ -78,29 +80,47 @@ public Response enlistWithCompensate(@HeaderParam(LRA.LRA_HTTP_CONTEXT_HEADER) U
...
+     */
     @Status
     public CompletionStage<ParticipantStatus> status(URI lraId) {
         assert lraId != null;
 
         return CompletableFuture.supplyAsync(() -> {
             lraMetricService.incrementMetric(LRAMetricType.Status, lraId, ValidLRACSParticipant.class);
 
-            return ParticipantStatus.Completed;
+            int completed = lraMetricService.getMetric(LRAMetricType.Completed, lraId, ValidLRACSParticipant.class);
```
xstefank, line 111: I think the original implementation is correct. When the Status is invoked, we should finish the LRA, and thus the hardcoded Completed status.

According to the specification, the status should not be called before Complete or Compensate so this implementation shouldn't ever return Completing.

I think that you ran into this issue, because you call waitForRecovery too soon in the tests (which I know that in narayana implementation it calls recovery two times).

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricType.java
```diff
@@ -27,8 +27,10 @@ public enum LRAMetricType {
     FailedToCancel, // an LRA that failed to cancel
 
     // Participant statistics
-    Compensated, // a participant that was asked to compensate
-    Completed, // a participant that was asked to complete
+    Compensating, // a participant that was asked to compensate
+    Completing, // a participant that was asked to complete
```
xstefank, line 31: These new enum values don't seem to be used.

xstefank (changes requested):

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
tomjenkinson: @xstefank please can you provide a reference to where status should not be called before complete or compensate (and even then it would be good to have a citation to shwo that it can't be called _while_ the participant is in the processs of either).

On the other hand, an example from the spec that appears to show it can be expected during the compensation process is: https://github.com/microprofile/microprofile-lra/blob/main/spec/src/main/asciidoc/microprofile-lra-spec.asciidoc?plain=1#L817

pom.xml
marcosgopen: +1, I will remove that commit from this PR

tck/pom.xml
marcosgopen: +1, I will remove that commit from this PR
BTW the new parent does not contain junit dependency

pom.xml
marcosgopen: Do you mean I should update this to 11 instead? I can do that

tck/src/main/java/org/eclipse/microprofile/lra/tck/TckParticipantTests.java
marcosgopen: Sure, let me change it to waitForCallbacks()

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
marcosgopen: The point here is to simulate a work load and avoid false positive, but I agree it might be confusing and I can remove it.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricType.java
marcosgopen: The completing one is now used here: https://github.com/microprofile/microprofile-lra/pull/368/files#diff-e547d1e228cde545e98971313c9216d09127f9e26176870910ae796172ca67f0R116
I think those 2 extra statuses are important to have better clarify what the Participant statuses can be

pom.xml
xstefank: No, remove it altogether, this is inherited from the parent.

tck/pom.xml
xstefank: We might need to switch to testng but this is a problem for another time.

pom.xml
```diff
@@ -40,6 +40,7 @@
 <!--        <bnd-maven-plugin.version>3.4.0</bnd-maven-plugin.version>-->
 <!--        <bnd-baseline-maven-plugin.version>4.2.0</bnd-baseline-maven-plugin.version>-->
 <!--        <build-helper-maven-plugin.version>1.12</build-helper-maven-plugin.version>-->
+        <maven.compiler.release>11</maven.compiler.release>
```
xstefank, line 43: remove this please

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
```diff
@@ -93,14 +95,40 @@ public CompletionStage<Response> complete(URI lraId) {
...
 
-            return ParticipantStatus.Completed;
+            int completing = lraMetricService.getMetric(LRAMetricType.Completing, lraId, ValidLRACSParticipant.class);
+
+            int compensating = lraMetricService.getMetric(LRAMetricType.Compensating, lraId, ValidLRACSParticipant.class);
+
+            int completed = lraMetricService.getMetric(LRAMetricType.Completed, lraId, ValidLRACSParticipant.class);
+
+            int compensated = lraMetricService.getMetric(LRAMetricType.Compensated, lraId, ValidLRACSParticipant.class);
+            // setting as default Completing, but it could also be Active or Compensating
```
xstefank, line 116: Now you are setting default to Active.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricType.java
xstefank: I still don't understand what is the value of these intermediate stages.  @marcosgopen please explain when they would be useful as you can see above I'm arguing that the Status method shouldn't be invoked before Complete/Compensate.

xstefank (changes requested):

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
marcosgopen: Thanks for your explanation Martin. 
The 'network failure' scenario is very interesting but in my opinion the tests I am changing here are not testing that case, but they are testing an async complete/compensate process. In this case I am seeing a potential problem when the async workload involved takes some time to finish and I want to make sure the test invokes 'waitForCallback' to make sure the complete/compensate process has finished .
IMO the waitForCallback should rely on the @Status method (when that annotated method is implemented) of the  ValidLRACSParticipant to understand the correct status of the LRA  (otherwise it could invoke again the compensate/complete method and wait for a 200 response).

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
marcosgopen: Do you think the default status should be different?

marcosgopen: Oh, maybe you were referring to the comment. Good point, I have just updated it. Thanks

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
xstefank: My concern is that we should probably then go over all Status methods in the TCK and adjust all tests for this? Tests are testing specific scenarios. So in my opinion, if the implementation invokes Status first before Complete/Compensate, there should be a reason. And I don't see any such reason in the specification. 

If you would make the Status implementation fully compliant with the specification (meaning tracking the started LRAs, which is not now done in this PR), you could also respond with 410 Gone because LRA method was not invoked yet. I think you can see my point. Keep the test simple, so the tested scenario is evident.

tck/src/main/java/org/eclipse/microprofile/lra/tck/service/LRAMetricType.java
xstefank: If this is your concern, even if the implementation of the Complete/Compensate basically return immediately, feel free to change the return values to `CompletableFuture.completedFuture`.

xstefank (changes requested):
Sorry @marcosgopen but I still think that Status method implementation brings unnecessary complexity.

tck/src/main/java/org/eclipse/microprofile/lra/tck/participant/nonjaxrs/valid/ValidLRACSParticipant.java
tomjenkinson: I don't think it's necessarily the case that status() is called _first_, but it is that the compensate()/complete() methods that have been called (first) can still be completing asynchronously by the time that the test tries to execute it's assertions. The ValidLRACSParticipant is the only participant in the TCK that does this and `testNonJaxRsCompletionStageVoid` and `testNonJaxRsCompletionStageResponseAndParticipantStatus` are the only two tests that are therefore exposed to a participants status method being incomplete.

If you recall https://github.com/microprofile/microprofile-lra/pull/368#discussion_r1969393875 then if you set that delay to (e.g. 10 seconds) I would presume any implementation would be likely to fail these tck tests. Basically the problem with the TCK tests `testNonJaxRsCompletionStageVoid` and `testNonJaxRsCompletionStageResponseAndParticipantStatus` is that they are both vulnerable to timing issues because their compensate or complete methods may not have updated the relevant metric by the time checkStatusReadAndCloseResponse returns and test assertions start to be performed. (@marcosgopen it might be worth updating the descriptions of this pull request and the corresponding GitHub issue to be something like "Improve testNonJaxRsCompletionStageVoid and testNonJaxRsCompletionStageResponseAndParticipantStatus to be better able to deal with slow performing complete and compensate methods"?)

xstefank (approved):

